PR #44, opened 2017-06-26
Add recognition of skipped tests to run_tests_parallel

ndevenish:
With the current test runner, tests that are skipped and don't run show as "`[OK]`" - this can be extremely misleading because running the tests without e.g. all the regression data configured will state that tests have all passed, when none of them actually ran. Also confusingly, when `phenix_regression` is configured some of these (but perhaps not all) are classified instead as warnings - which may mask actual important warning information.

This patch alters the test runner to recognise skips the same way as it does for the JUnit XML output.

This means that it works by looking for the case insensitive word 'skip' in the output. Consequently, output that's a multi-test script consisting of passes/skips are classified as skipped - under the reasoning that it's better to know when tests haven't run. Output classified as warnings by `phenix_regression` are left as warnings, under the same reasoning.

The change is in two commits; the first does some tidying and refactoring to make it easier to add the new categorisation, and the second actually adds it.

Examples of new output (run on libtbx,scitbx,cctbx,iotbx,rstbx,dxtbx,dials,dials_regression):
```
libtbx.python ".../cctbx_project/iotbx/dsn6/tst.py" P31 [OK]
libtbx.python ".../cctbx_project/iotbx/regression/ncs/tst_mtrix_biomt_cmdl.py" [SKIP]
libtbx.python ".../cctbx_project/iotbx/regression/ncs/tst_ncs.py" [OK]

.....

Summary:
  Tests run                    : 576
  Failures                     : 0
  Warnings (possible failures) : 0
  Stderr output (discouraged)  : 0
  Skipped                      : 53
```

### DIFF
--- a/libtbx/test_utils/parallel.py
+++ b/libtbx/test_utils/parallel.py
@@ -23,6 +23,7 @@ class Status(IntEnum):
   OK = 0
   WARNING = 1
   FAIL = 2
+  SKIPPED = 3
 
 QUIET = 0
 DEFAULT_VERBOSITY = 1
@@ -173,7 +174,7 @@ def write_JUnit_XML(results, output_filename="output.xml"):
                   stderr='\n'.join(plain_stderr))
     if result.return_code == 0:
       # Identify skipped tests
-      if re.search('skip', output, re.IGNORECASE):
+      if result.alert_status == Status.SKIPPED:
         # find first line including word 'skip' and use it as message
         skipline = re.search('^((.*)skip(.*))$', output, re.IGNORECASE | re.MULTILINE).group(1)
         tc.add_skipped_info(skipline)
@@ -268,9 +269,11 @@ class run_command_list (object) :
     longjobs = [result for result in self.results if result.wall_time > max_time]
     warnings = [result for result in self.results if result.alert_status == Status.WARNING]
     failures = [result for result in self.results if result.alert_status == Status.FAIL]
+    skipped  = [result for result in self.results if result.alert_status == Status.SKIPPED]
     self.finished = len(self.results)
     self.failure = len(failures)
     self.warning = len(warnings)
+    self.skipped = len(skipped)
 
     # Try writing the XML result file
     write_JUnit_XML(self.results, "output.xml")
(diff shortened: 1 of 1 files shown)


nwmoriarty:
Seems like an improvement. 👍

phyy-nx:
Hi folks, we need to revert this pull request.

1) cctbx_regression.test_nightly is failing on our build bot servers.  I've listed the exception below. 
2) The version as committed finds too many "skips" that are false skips.  These come from tests of programs with skip in their parameter names, or from tests that have skip in their output.

After chatting about it, we're too close to a phenix release.  After the release, it would be great to get this back in because it is super useful, although it will need a smarter approach for finding skipped tests.

I'm checking in the revert in a bit.
-Aaron

Exception:
  File "/Users/builder/slave/cctbx-nightly-mac-intel-osx-x86_64/build/../modules/cctbx_project/cctbx/command_line/cctbx_test_nightly.py", line 23, in <module>
    if (run_tests_parallel.run(args) > 0) :
  File "/Users/builder/slave/cctbx-nightly-mac-intel-osx-x86_64/modules/cctbx_project/libtbx/command_line/run_tests_parallel.py", line 121, in run
    max_time=params.max_time)
  File "/Users/builder/slave/cctbx-nightly-mac-intel-osx-x86_64/modules/cctbx_project/libtbx/test_utils/parallel.py", line 279, in __init__
    write_JUnit_XML(self.results, "output.xml")
  File "/Users/builder/slave/cctbx-nightly-mac-intel-osx-x86_64/modules/cctbx_project/libtbx/test_utils/parallel.py", line 179, in write_JUnit_XML
    skipline = re.search('^((.*)skip(.*))$', output, re.IGNORECASE | re.MULTILINE).group(1)
AttributeError: 'NoneType' object has no attribute 'group'

ndevenish:
That's strange. I'd like to ask:
- Which tests it fails like this on
- Examples of tests which are incorrectly classified - I looked at a large chunk of the ones classified as skipped and most were due to missing phenix_regression

nwmoriarty:
If have phenix_regression (which all of our BuildBots do), all will be revealed. The command

cctbx_regression.test_nightly 

fails in the summary step. And there are more than one test involving phil parameters like skip_twin_test=True, etc.

ndevenish:
Hi, I'm trying to replicate the environment and testing that was causing these issues, in order to improve the classification. The crash was an easy enough fix, but I am only finding one test in `run_tests_parallel module=phenix_regression` being misclassified, and 3 out of the 127 in `phenix_regression.test_nightly`. Is this the level of skipped tests that you observed, or am I missing some large part of the testing suite?

Worrying, there appear to be tests in `phenix_regression.test_nightly` that aren't run when running via `libtbx.run_tests_parallel` (e.g. entries missing from `run_tests.py`). Is this deliberate?

PR #103, opened 2017-11-30
Add development version of cxi.merge that uses a reflection table

phyy-nx:
Two changes to cxi.merge:
  Move summation code out of run method into method of the scaling manager
  Move plot single reflection histograms to own function

Concurrent changes to dev.cxi.merge

New program dev.cxi.merge_refltable that uses a dials reflection table as the underlying data structure for holding reflections during merging.

### DIFF
--- a/xfel/command_line/cxi_merge.py
+++ b/xfel/command_line/cxi_merge.py
@@ -1736,6 +1736,17 @@ class scaling_manager (intensity_data) :
     data.show_log_out(sys.stdout)
     return data
 
+  def sum_intensities(self):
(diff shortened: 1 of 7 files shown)


nksauter:
I'd like to reverse the decision to put show_overall_observations() in the scaling_manager class.  The method is concerned with reporting results to the log file, not primarily with producing the data analysis itself.  It makes specific decisions about what columns to display--these types of decisions should be refactored away from the main class, not enshrined within.  In the future, we will want a flexible tool to manage table display that allows us to add new columns and analyze results in different ways.  We want that code as far away as possible.

PR #105, opened 2017-12-04
Updated cxi_merge.py for cell rejection

mlxd:
Changes to cxi_merge.py to selectively filter unit cells.


nksauter:
Everything looks good.  Will rebase and merge now.

PR #113, opened 2017-12-12
Retire stdlib top level module.

Anthchirp:
Before this can be merged the following files need to be refactored:
* [x] phenix/refinement/weight_xray_chem.py
* [x] phenix/ligands/invariant_generator.py

**I'll merge this myself but want to leave it until after the Phenix release is done.**

This was used as a workaround in scitbx because scitbx is shadowing two system library modules (math, random), but relies on being able to import the system libraries as well.
Scitbx could be refactored to remove this stdlib hack entirely, but here I've opted to move the stdlib hack from libtbx/pythonpath into scitbx. This frees up another item from libtbx/pythonpath.

Code was refactored to make explicit whether it refers to python math vs scitbx.math.

### DIFF
--- a/cctbx/eltbx/distance_based_connectivity.py
+++ b/cctbx/eltbx/distance_based_connectivity.py
@@ -1,5 +1,5 @@
-from __future__ import division
-from stdlib import math as smath
+from __future__ import absolute_import, division, print_function
+import math
 
 expected_bond_lengths_by_element_pair = {
   # based on elbow/chemistry/BondLengths.py rev. 42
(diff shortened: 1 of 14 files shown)


Anthchirp:
The required patch for the phenix repository can be found at https://gist.github.com/Anthchirp/9f4aeb6e1fda66681d0cf3a01d8f77f5 and can be applied any time.

Anthchirp:
prerequisite phenix patch committed.

Anthchirp:
more complex than expected, but looks okay now.

Result.

PR #118, opened 2018-01-11
Add PDB-REDO mirror to fetch function

fsimkovic:
Add a function to allow `cif` and `pdb` downloads from the [PDB-REDO server](https://pdb-redo.eu/).

Example to fetch
```python
from iotbx.pdb.fetch import fetch
pdb = "1aba"
fetch(pdb, data_type='pdb', format='pdb', mirror='pdb-redo')
fetch(pdb, data_type='xray', mirror='pdb-redo')
fetch(pdb, format='cif', mirror='pdb-redo')
fetch(pdb, format='xml', mirror='pdb-redo')
```


bkpoon:
Looks good, except for the unverified SSL context (lines 155-156) for downloading from the PDB-REDO https website. Are you getting an error when running the function? The Python that we build should have a relatively up-to-date SSL package and certificates for validating http websites.

We'll also be updating the base URL for the other websites to use https over http.

fsimkovic:
I do get an error when attempting to download via CCP4. Besides the SSL context, I tried replacing `urllib2` in `libtbx.utils.urlopen` with [`requests.get`](http://docs.python-requests.org/en/master/), which also worked fine.

I can remove the created `context` and modify the URLs to use `https` over `http`.

bkpoon:
Some issues that I found from more testing is that the model in CIF format is being downloaded in the section about x-ray data and the MTZ file from PDB-REDO is not downloaded. I made some changes and committed them to my forked repository (https://github.com/bkpoon/cctbx_project/commit/9926961b9f2dd928b8692885a0816185943ebbf4)
I also centralized the changes into one section similar to how the PDBj mirror is handled. Please check that I did not miss any functionality and update the pull request.

I could not find data in CIF format on the PDB-REDO server, so there may be some issues with that, but that's fine.

Also, we do not depend on ```requests```, so we are just using standard Python modules. We'll update when we move to Python 3.

fsimkovic:
I've merged your suggested changes as well as the other updates.

bkpoon:
Great, thanks! I've merged your pull request.

PR #119, opened 2018-01-12
Virtualenv issues

gbunkoczi:
Another solution for b46e27a is to set PYTHONPATH to include python lib for python used to create the virtualenv (__future__.py is not present in virtualenv's lib/python27).

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -1559,8 +1559,8 @@ selfx:
     else:
       # fallback in case of virtualenv
       # https://github.com/pypa/virtualenv/issues/355
-      pythonpath = [self.as_relocatable_path(
-        op.join(sys.prefix,'lib',self.python_exe.basename(),'site-packages'))]
+      from distutils.sysconfig import get_python_lib
+      pythonpath = [self.as_relocatable_path(get_python_lib())]
     pythonpath.append(self.lib_path)
     for module in self.module_list:
       pythonpath.extend(module.assemble_pythonpath())


Anthchirp:
correct solution definitely would indeed have been to remove __future__ import from site.py. If libtbx complains about that then the problem is with libtbx.
site.py removed, see 916f588783765f94464068d6887401bdfbba3a08

not sure about the other virtualenv stuff or whether that is now still required.

gbunkoczi:
Strictly speaking, the "other stuff" (I am assuming you are referring to
commit fd6fc27) is not required, as it does not fix (affect?) the current
issue. However, IMHO it is not right to set the PYTHONPATH incorrectly, and
it fixes that (this got me sidetracked in resolving the current issue, and
I want to spare someone else the same time investment).

I will merge in the current change to resolve the conflict, and leave the
request open for Billy to comment.

Anthchirp:
instead of adding a full merge into the pull request to remove a single, now no longer required commit from the same pull request...
please run the following in your repository with the pull request branch checked out:
```
git reset --hard HEAD^ # remove the added merge again
git rebase -i HEAD~2 # interactively rebase the two remaining commits of the pull request
# an editor opens. delete the line with the 'Remove import' commit. Save and quit
git push -f
```
Then this pull request will consist of a single commit again.
You can rebase onto cctbx master if you wish, but that can also be done when merging.

Anthchirp:
also note that the email address you are using in your git configuration is not registered with Github, hence the question marks next to your name in the commit views, eg. https://github.com/cctbx/cctbx_project/pull/119/commits/fd6fc279ea49bb526fb6b22be4e310d9f13eebb0

gbunkoczi:
I am happy to be educated on best practises, so many thanks for bringing
this up!

bkpoon:
Hi Gabor,

Sorry for the problem! I originally modified the code because some user had their own local site-packages directory that was interfering with ours. Some of our servers in the build system were using virutalenv, so I added that fix for virtualenv, but I only really tested it on our systems.

I like your fix and prefer that PYTHONPATH is correctly set, so feel free to merge the commit.

gbunkoczi:
Hi Billy,

well, the bigger question is whether you think this would cause any
problems on your side. I haven't done exhaustive testing on this (and since
we lack Phenix, I will not be able to run the phenix_regression part). For
now it is probably enough that you are aware of this change, and know where
to look if things go pear-shaped.

I go ahead and merge this into master.

Many thanks Markus and Billy for the helpful comments.

BW, Gabor

PR #120, opened 2018-01-18
Added phil scoping and CS model to mpi merging

mlxd:
Enabling client-server mode in dev_mpi_cluster_two_merge.py is enabled using the PHIL params `mpi.cs=True`

Additionally, logging the MPI timings is enabled with `mpi.logging=True`
The logging statements are of the format '~<Unique string> START/END RANK=<RANK> TIME=<Unix time>;'  where the tilde and semicolon are used to help clean up the messages when post-parsing the output.

### DIFF
--- a/xfel/merging/command_line/dev_mpi_cluster_two_merge.py
+++ b/xfel/merging/command_line/dev_mpi_cluster_two_merge.py
@@ -1,8 +1,10 @@
 # LIBTBX_SET_DISPATCHER_NAME dev.mpi.cluster_two_merge
 from __future__ import division
 import sys,time
+from libtbx.utils import Usage
 
 from xfel.command_line.cxi_merge import scaling_manager as scaling_manager_base
+
 class scaling_manager_mpi(scaling_manager_base):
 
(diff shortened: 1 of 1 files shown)


nksauter:
Review comments from Nick:
1) Please run libtbx.find_clutter
2) In an effort to become future proof, please "from future import print_function"
3) If dev_mpi_cluster_two_merge_cs.py is no longer needed, then perhaps it could be removed as part of the pull request?
4) The initialize() method seems to duplicate code that was in the parent class in dev.cxi.merge.  Is there a better way to handle this without code duplication.  For example, one could inject the new phil string into the parent scope, then call the parent initialize() function.  Ask me if it is unclear how.
5) I'll review the timing statements at a later date, perhaps after the pull request is committed.
6) Please change the structure of the phil scope so that load_balancing algorithm is a choice;
mpi { load_balancing {
   .type = choice *striping cs
}}
7) Once this is done, the if statement should be made to be more self documenting:
if params.mpi.load_balancing=="cs":
   <something>
elif params.mpi.load_balancing=="striping":
   <something else>

PR #121, opened 2018-01-19
Scan varying beam

dagewa:
This branch contains changes to dxtbx to support scan-varying beam vector s0. These changes add rather than alter functionality and are expected to be safe. However, the details of serialisation in particular are tricky, so I request review from @jmp1985 to check I haven't missed anything.

### DIFF
--- a/dxtbx/model/beam.h
+++ b/dxtbx/model/beam.h
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <cmath>
 #include <scitbx/vec3.h>
+#include <scitbx/array_family/shared.h>
 #include <scitbx/array_family/simple_io.h>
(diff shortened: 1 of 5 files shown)


dagewa:
Hi James, thanks for the review. I read that link and I think you're right, if `getstate` and `setstate` are defined then it is _advised_ to manually pickle `__dict__`. It isn't always necessary, but it is safer to do so for the reasons described in the section "Pitfall and Safety Guard". Because of that, the "Practical advice" states:

> * If `__getstate__` is required, include the instance's `__dict__` in the Python object that is returned.

In this case, `getstate` / `setstate` **are** required because the scan-varying s0 has to be set via a method call after construction and cannot be passed into `__init__`. So here I think the beam `__dict__` should be manually pickled.

PR #122, opened 2018-01-23
remove libtbx/pexpect.py

Anthchirp:
As far as I can tell this is not used anywhere. This seems derived from the [pypi module pexpect](https://pypi.python.org/pypi/pexpect), the installation of which should be preferred over copying the code anyway.

Luc, do you use this anywhere or can it be removed safely?


luc-j-bourhis:
This is a leftover from a dead project: you can remove it indeed.

Anthchirp:
Thank you.

PR #123, opened 2018-01-23
Fix xtc_process when reading multiple runs

elliottslaughter:
Fixes a bug where enumerating multiple runs causes events to be processed under the wrong run, because `ds.events()` returns all the events, instead of the events for a specific run.


phyy-nx:
Looks good.  Thanks!

PR #124, opened 2018-01-25
Fix MosaicCrystal(Sauter2014|Kabsch2010) warning messages

ndevenish:
The `dxtbx::model::Crystal` subclasses `MosaicCrystalKabsch2010` and `MosaicCrystalSauter2014` both defined their [own](https://github.com/cctbx/cctbx_project/blob/99ffd3c3d4483e83c6466b1803775d40ce12baec/dxtbx/model/crystal.h#L1012-L1017), [extended](https://github.com/cctbx/cctbx_project/blob/99ffd3c3d4483e83c6466b1803775d40ce12baec/dxtbx/model/crystal.h#L1136-L1142) function signatures for the virtual [is_similar_to](https://github.com/cctbx/cctbx_project/blob/99ffd3c3d4483e83c6466b1803775d40ce12baec/dxtbx/model/crystal.h#L214-L218). This appears to hide the superclass implementation, breaking polymorphism, and clang indeed emits warnings for this (_"hides overloaded virtual function"_).

This moves the default instantiation to an inheritance override function, to resolve the warning, and passes explicit member-function pointers into the boost::python bindings (required to resolve the overload).

The slightly obscure pointer dance seems to be required to get Boost::Python to choose a specific overload - otherwise the `def` line errors. I don't know another way around this.

### DIFF
--- a/dxtbx/model/boost_python/crystal.cc
+++ b/dxtbx/model/boost_python/crystal.cc
@@ -352,6 +352,16 @@ namespace dxtbx { namespace model { namespace boost_python {
             arg("space_group_symbol"))))
       .def_pickle(CrystalPickleSuite());
 
+    // Create member-function pointers to specific is_simiar_to overloads
+    // - each of these crystals has a custom implementation along with the
+    //   inherited interface, and we want to expose these explicitly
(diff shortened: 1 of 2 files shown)


ndevenish:
Thanks!

PR #127, opened 2018-01-26
Additions to dxtbx model code to include a scaling model

jbeilstenedmands:
I have added a scaling_model to the Experiment list and experiment files, alongside a scaling model interface and factory in scaling_model.py
I have mirrored the layout on the profile model example, and the actual code for the scaling models will be kept within the DIALS project, as is the case for the profile model.

One thing that I haven't modified in experiment_list.py that may need to be updated is how imagesets are constructed - I haven't included the scaling models in this (in _extract_experiments).
I have also made as few changes as possible to the .cpp and .h files but there may need to be more places where these are updated to include the scaling_model.


graeme-winter:
@jmp1985 please could you look at this one?

dxtbx/model/scaling_model.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import absolute_import, division
```
Anthchirp: print_function

dxtbx/model/scaling_model.py
```diff
@@ -0,0 +1,72 @@
...
+
+from abc import ABCMeta, abstractmethod
+
+class ScalingModelBaseIface(object):
```
Anthchirp: does this class actually add any value? basically you just say any implementation of this will have to_dict and from_dict but not much else. In Java this interface would make sense, but in python any object that has those methods has them.

dxtbx/model/scaling_model.py
```diff
@@ -0,0 +1,72 @@
...
+  @abstractmethod
+  def from_dict(cls, obj):
+    '''
+    Convert a dictionary to the scaling model
+
+    '''
+    pass
+
+
+class ScalingModelFactory(object):
```
Anthchirp: Is this class cargo-cul... erm, copied from somewhere, or is this from scratch?
In any case would suggest using the python setuptools registry rather than rolling our own.

dxtbx/model/__init__.py
```diff
@@ -9,6 +9,7 @@
 from dxtbx.model.scan import *
 from dxtbx.model.crystal import *
 from dxtbx.model.profile import *
+from dxtbx.model.scaling_model import *
```
Anthchirp: ...could we please not _import *_?

This makes it basically impossible to find out where a name came from and thus incurs a huge technical debt for the unfortunate soul who one day wants to figure out how the stuff in this file works. And that is without any name conflicts.

dxtbx/model/__init__.py
```diff
@@ -396,6 +397,12 @@ def profiles(self):
     ''' Get a list of the unique profile models (includes None). '''
     from libtbx.containers import OrderedSet
     return list(OrderedSet([e.profile for e in self]))
+  
+  def scaling_models(self):
+    ''' Get a list of the unique scaling models (includes None). '''
+    from libtbx.containers import OrderedSet
```
Anthchirp: is the order relevant? why not simply _return list(set(e.profile for e in self))_?
If order is actually relevant please remove all _import OrderedSet_ from file and add a single import to the top.

dxtbx/model/__init__.py
```diff
@@ -396,6 +397,12 @@ def profiles(self):
...
+    return list(OrderedSet([e.scaling_model for e in self]))
```
Anthchirp: Whether set or OrderedSet - in any case please drop the list comprehension brackets

dxtbx/model/scaling_model.py
jbeilstenedmands: As you hint to below, most of this is copied straight from James Parkhurst's profile model interface code. Wasn't sure if it was necessary, so happy to remove it if everything will still work fine.

jbeilstenedmands: Okay, I've checked and I don't actually need to use this in the rest of my code so we can remove it.

dxtbx/model/__init__.py
jmp1985: Order may be relevant, the functions were intended to return the unique models in the order they were encountered in the experiment list so I think this method is fine. Also I prefer having imports inside functions rather than in the file scope so I think this is fine.

dxtbx/model/__init__.py
```diff
@@ -542,6 +552,7 @@ def get_template(imset):
     result['scan']       = [s.to_dict() for s in slist if s is not None]
     result['crystal']    = [c.to_dict() for c in clist if c is not None]
     result['profile']    = [p.to_dict() for p in plist if p is not None]
+    result['scaling_model']    = [s.to_dict() for s in scalelist if s is not None]
```
jmp1985: This indentation is probably my fault! Feel free to unalign the lines.

dxtbx/model/boost_python/experiment_list.cc
```diff
@@ -262,7 +262,8 @@ namespace dxtbx { namespace model { namespace boost_python {
         arg("scan")       = boost::shared_ptr<Scan>(),
         arg("crystal")    = boost::shared_ptr<CrystalBase>(),
         arg("profile")    = boost::python::object(),
-        arg("imageset")   = boost::python::object()))
+        arg("imageset")   = boost::python::object(),
+        arg("scaling_model")   = boost::python::object()))
```
jmp1985: Again this is due to my habit of aligning stuff so feel free to unalign the =

dxtbx/model/boost_python/experiment.cc
```diff
@@ -94,14 +95,16 @@ namespace dxtbx { namespace model { namespace boost_python {
...
           boost::python::object>((
               arg("beam")       = boost::shared_ptr<BeamBase>(),
               arg("detector")   = boost::shared_ptr<Detector>(),
               arg("goniometer") = boost::shared_ptr<Goniometer>(),
               arg("scan")       = boost::shared_ptr<Scan>(),
               arg("crystal")    = boost::shared_ptr<CrystalBase>(),
               arg("profile")    = boost::python::object(),
-              arg("imageset")   = boost::python::object())))
+              arg("imageset")   = boost::python::object(),
+              arg("scaling_model") = boost::python::object())))
```
jmp1985: Again this is due to my habit of aligning stuff so feel free to unalign the =

dxtbx/model/experiment_list.h
```diff
@@ -412,7 +412,8 @@ namespace dxtbx { namespace model {
       }
       scitbx::af::shared<std::size_t> result;
       for (std::size_t i = 0; i < size(); ++i) {
-         if (data_[i].get_profile() == obj || data_[i].get_imageset() == obj) {
+         if (data_[i].get_profile() == obj || data_[i].get_imageset() == obj ||
+           data_[i].get_scaling_model() == obj) {
```
jmp1985: This might look nicer but I'm just being picky and it's only an opinion!
```
         if (data_[i].get_profile() == obj || 
             data_[i].get_imageset() == obj ||
             data_[i].get_scaling_model() == obj) {
```

dxtbx/model/experiment_list.py
```diff
@@ -486,9 +497,11 @@ def as_json(self, filename=None, compact=False, split=False):
           e['crystal'] = clist[e['crystal']][0]
         if 'profile' in e:
           e['profile'] = plist[e['profile']][0]
+        if 'scaling_model' in e:
+          e['scaling_model'] = plist[e['scaling_model']][0]
```
jmp1985: I think this should be
```
        if 'scaling_model' in e:
          e['scaling_model'] = scalelist[e['scaling_model']][0]
```

dxtbx/model/scaling_model.py
jmp1985: This should probably be done for the profile model code as well at some point as well!

dxtbx/model/experiment_list.py
jbeilstenedmands: Yes you are correct, well spotted.

dxtbx/model/__init__.py
jbeilstenedmands: Okay, I will unalign all relevant instances in the modified files.

dxtbx/model/__init__.py
jbeilstenedmands: Okay, I think I can remove this and just use an import in the experiment_list.py instead when I import ScalingModelFactory - and then even this can be removed if we use the setuptools?

dxtbx/model/__init__.py
Anthchirp: imports inside functions not really sensible here though. Imo that should only be done in a case where lazy importing actually makes a difference. I do not think this is the case here?
I would put _import libtbx.containers at the top, then the dependencies are clear to the reader at a glance, as is the origin of the name.

dxtbx/model/__init__.py
Anthchirp: comment was along the lines of 

```python
import dxtbx.model.scaling_model
import otherthing
dxtbx.model.scaling_model.thing() # so that's a scaling_model thing
```
is better than

```python
from dxtbx.model.scaling_model import *
from otherthing import *
thing() # so where does this come from? Even if you know that scaling_model has a thing, does otherthing have a thing that overwrote the thing from scaling_model? Now you have to look into 2 more files to find the answer. Good luck if any of them use import *.
```

If the ```import``` itself is not required then that is obviously even better.

jbeilstenedmands:
I have now made the changes suggested. The scaling_model.py is now not needed, if one registers scaling model extensions in other modules under the 'dxtbx.scaling_model_ext' entry point.

jbeilstenedmands:
I am going to register the entry points and extensions within the dials code rather than here.

Anthchirp:
👍

dxtbx/model/experiment_list.h
Anthchirp:
```python
if obj in (data_[i].get_profile(),
           data_[i].get_imageset(),
           data_[i].get_scaling_model()) {
```

PR #128, opened 2018-01-26
Add an memory allocator model to scitbx.array_family

ndevenish:
This change adds custom memory allocation control, (mostly compatible with the new standard library [polymorphic_resource](http://en.cppreference.com/w/cpp/header/memory_resource)  API) to the shared and versa families of scitbx's array_family.

There are no changes required to any other existing code in order to continue with the current behaviour (which uses `new`/`delete` for allocation). To use a custom allocator, a persistent `memory_resource` instance can be passed as a new, optional final argument to most of the constructors:

```C++
shared<T> some_data(SIZE, some_memory_resource);
```
Additionally, the default behaviour can be changed at run-time so that the allocation model doesn't have to be threaded through lots of other code.

I'm aware that this is a large change to a relatively core part of the packages, so am expecting feedback and comments. I'm happy to discuss or expand on any part of my reasoning behind these changes, and any mistakes you think I might have made.

# Detailed Description of Changes

### Motivation

On profiling of dials spotfinding we were consistently noticing that memory allocation was taking a reasonable proportion of the run-time; with images being allocated, wiped, used and then freed multiple many times for each frame being processed. Moving to a more complex memory model e.g. memory pools requires the capability to intercept or override the `new`/`delete` calls made inside the allocating containers. This requires the choosing or design of an allocator model.

### Allocation Model

C++98/03 contains an allocator model, which is implemented on the standard library containers. It is entirely based around template parameters, complex, and widely considered not-fit-for-purpose. Attempts to retrofit features in C++11 and C++14 only complicated matters. Therefore, a new model with "polymorphic allocators" was [proposed](https://isocpp.org/files/papers/N3916.pdf) and [accepted](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4480.html) into the language for C++17 (§23.12 in the C++17 standard). This model is much simpler, and more intuitively fits to what you actually want out of an allocation model. There is a talk from CPPCON17 by the author of the proposal that summarized the problems and motivations for this reasonably well: https://youtu.be/v3dz-AKOVL8.

Using the newer C++17 API I believe makes the library more future-proof than a custom implementation; Once onto C++11 we could switch to an externally developed and supported third-part implementation, and in the distant future when C++17 is the default we can remove entirely the support code.

This model and API can be 100% implemented in C++11. A large proportion of the most useful parts can be implemented in C++03, and that's what I've implemented here (almost all of it can be ported with heavy template metaprogramming - and indeed Boost has done a lot of this - except the boost implementation requires an extra boost library to be compiled and linked, and I know that we're resistant to those sorts of changes). The parts of the main API that are not implemented are:

- Automatic forwarding of scoped allocators via contruct
- Alignment specification. Some parts of the API allow specification of alignment separately from the C++11 `alignas` type specifiers, but these are not currently implemented. In debug builds, the alignment is asserted to be the default alignment (16 bytes for 64-bit platforms - which I believe is all we support now)
- Specialized allocators (`synchronized_pool_resource`, `unsynchronized_pool_resource` and `monotonic_buffer_resource`) are not implemented. Although I plan to implement custom allocators to solve our problem, these don't need to adhere to the C++17 public API and thus can be less generic.

### Summary of per-file changes

All of these files in `scitbx/array_family/` have been changed:

| File                          | Changes                                     |
| ----                          | -------                                     |
| SConscript                    | Add tests to compilation                    |
| boost_python/shared_wrapper.h | Choose specific overload of `deep_copy` to export |
| memory.h                      | Removed - was old implementation to ensure 16-byte alignment |
| polymorphic_allocator.h       | Implementation of C++17 allocator model     |
| tests/test_pmr.cpp            | Tests of basic allocator model and interface |
| tests/test_pmr_shared_plain.cpp | Tests for using allocators in `shared_plain` |
| tests/testing.h               | Basic C++ testing framework; uses googletest instead if available |
| tests/testing_main.cpp        | Contains the main function for the basic test framework |
| shared.h                      | Implement allocator API                     |
| shared_plain.h                | Implement allocator API                     |
| versa.h                       | Implement allocator API                     |
| versa_plain.h                 | Implement allocator API                     |

In addition, the newly built test is added to `scitbx/run_tests.py`.

### Implementation details/Notes

Some details or notes on caveats/pitfalls/odd code (I can expand with examples if required):

- In order to cope with static globals in a build environment where we don't have .cpp files ignore the capabilities of code-sharing, the shared "global" data is stored inside a static template in order to get around this and the One-definition-rule. It's not ideal, but the only way we have. Luckily, the way python loads modules seems not to cause issues with this across different extension modules.
- There's some slightly sticky behaviour with copy-assignment and swap for the array_family containers when applying to two containers with different allocators. Essentially, because these classes change the semantic meaning of copying to mean sharing common data, potentially assignment could require either a) changing the allocator of the object being assigned to - this is confusing, as if you construct an object with a specific allocator you expect it to use that, or b) an actual data copy operation to move the data to the new allocator. This breaks the array_family semantics that assignment is cheap, and the expectation that you would end up with a shared reference. Therefore, these operations have a hard assert that the allocators are the same. In practice, cases where this might happen can usually be resolved with careful copy-construction instead.

### Testing

I've added tests for the new code and methods of creating and passing allocators through the shared_plain class directly. Additionally, I've run the following test groups, which I believe should cover things tests both here and over there:

  - `libtbx.run_tests_parallel` for modules libtbx, scitbx, cctbx, iotbx,
    rstbx, dxtbx, dials, dials_regression, xia2 and xia2_regression
  - cctbx_regression.test_nightly
  - phenix_regression.test_nightly

I'd love to be advised on any additional test modules or suites that I should be including in this. I'm also aware of our limited variety of platforms and environments we have for testing here, so this needs to be looked at more widely.


nksauter:
From our viewpoint at LBNL, many of us are quite concerned about making
low-level changes in the fundamental array types that CCTBX has used for
the past 15 years.  We need to be convinced that there is a serious
problem, before the code is called into question.

We are skeptical at a basic level.  For instance you state that the
spotfinder allocates, wipes, uses and frees each image many times.  Why is
it not possible to simply rewrite the application?  Also, in our experience
there are a few optimizations that everyone should know about flex types,
but we don't know if you have taken them into account:
1) The default behavior of flex arrays is to initialize all array elements
upon construction, which can be quite inefficient.  Therefore, if we are
constructing a new flex array destined to be filled with data, we skip the
element initialization by providing an init_functor_null<> to the
constructor.
2) If we are iterating though a flex array, the default accessor,
operator[], is usually extremely slow.  We always cast the iterator to a
raw pointer and then access the elements through the pointer:  double *
fastptr = array.begin().

Are we sure this is a defect in the lower level allocation, and not some
side effect of wrapping arrays improperly in Boost.Python?

Also, if memory pools are really the answer for this one application, why
not start off in a more gentle way and implement a memory pool at the
application level--perhaps a collection of constant-sized flex arrays that
would stay in scope for the whole application?

How does the code performance improve with the proposed code?  Do you have
before & after metrics?

Why was memory.h taken out?  Luc Bourhis implemented it in 2012, so did you
check with him?

The LBNL group is still supporting Centos 5.  Will the new code function
properly?

Cheers,
Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

On Fri, Jan 26, 2018 at 7:23 AM, Nicholas Devenish <notifications@github.com

nksauter:
Hi Everyone,

  I greatly appreciate Nicholas’s goal of future proofing and improving functionality. However, I do share Nick and others’ concerns about the impact on current applications. In particular, the need to support older platforms could be an issue for Phenix. I’ll also suggest that there is a much larger set of tests that need to be run to verify Phenix - that test full applications (others can provide details I hope). Additionally, we’d need to be very sure that everything works as expected on all platforms (Mac, Linux, Windows) for all currently supported versions of OSs. I advocate for a lot of testing before consideration of checking in changes to this code - the impact of Phenix being non-functional on one of the platforms would be a major problem for us.

  Cheers,
	Paul

> On Jan 26, 2018, at 5:12 PM, Nicholas Sauter <nksauter@lbl.gov> wrote:
> 
> From our viewpoint at LBNL, many of us are quite concerned about making low-level changes in the fundamental array types that CCTBX has used for the past 15 years.  We need to be convinced that there is a serious problem, before the code is called into question.  
> 
> We are skeptical at a basic level.  For instance you state that the spotfinder allocates, wipes, uses and frees each image many times.  Why is it not possible to simply rewrite the application?  Also, in our experience there are a few optimizations that everyone should know about flex types, but we don't know if you have taken them into account:
> 1) The default behavior of flex arrays is to initialize all array elements upon construction, which can be quite inefficient.  Therefore, if we are constructing a new flex array destined to be filled with data, we skip the element initialization by providing an init_functor_null<> to the constructor.
> 2) If we are iterating though a flex array, the default accessor, operator[], is usually extremely slow.  We always cast the iterator to a raw pointer and then access the elements through the pointer:  double * fastptr = array.begin().  
> 
> Are we sure this is a defect in the lower level allocation, and not some side effect of wrapping arrays improperly in Boost.Python?
> 
> Also, if memory pools are really the answer for this one application, why not start off in a more gentle way and implement a memory pool at the application level--perhaps a collection of constant-sized flex arrays that would stay in scope for the whole application?  
> 
> How does the code performance improve with the proposed code?  Do you have before & after metrics?
> 
> Why was memory.h taken out?  Luc Bourhis implemented it in 2012, so did you check with him?
> 
> The LBNL group is still supporting Centos 5.  Will the new code function properly?
> 
> Cheers,
> Nick
> 
> Nicholas K. Sauter, Ph. D.
> Senior Scientist, Molecular Biophysics & Integrated Bioimaging Division
> Lawrence Berkeley National Laboratory
> 1 Cyclotron Rd., Bldg. 33R0345
> Berkeley, CA 94720
> (510) 486-5713
> 
> On Fri, Jan 26, 2018 at 7:23 AM, Nicholas Devenish <notifications@github.com> wrote:
> This change adds custom memory allocation control, (mostly compatible with the new standard library polymorphic_resource API) to the shared and versa families of scitbx's array_family.
> 
> There are no changes required to any other existing code in order to continue with the current behaviour (which uses new/delete for allocation). To use a custom allocator, a persistent memory_resource instance can be passed as a new, optional final argument to most of the constructors:
> 
> shared<T> some_data(SIZE, some_memory_resource);
> Additionally, the default behaviour can be changed at run-time so that the allocation model doesn't have to be threaded through lots of other code.
> 
> I'm aware that this is a large change to a relatively core part of the packages, so am expecting feedback and comments. I'm happy to discuss or expand on any part of my reasoning behind these changes, and any mistakes you think I might have made.
> 
> Detailed Description of Changes
> 
> Motivation
> 
> On profiling of dials spotfinding we were consistently noticing that memory allocation was taking a reasonable proportion of the run-time; with images being allocated, wiped, used and then freed multiple many times for each frame being processed. Moving to a more complex memory model e.g. memory pools requires the capability to intercept or override the new/delete calls made inside the allocating containers. This requires the choosing or design of an allocator model.
> 
> Allocation Model
> 
> C++98/03 contains an allocator model, which is implemented on the standard library containers. It is entirely based around template parameters, complex, and widely considered not-fit-for-purpose. Attempts to retrofit features in C++11 and C++14 only complicated matters. Therefore, a new model with "polymorphic allocators" was proposed and accepted into the language for C++17 (§23.12 in the C++17 standard). This model is much simpler, and more intuitively fits to what you actually want out of an allocation model. There is a talk from CPPCON17 by the author of the proposal that summarized the problems and motivations for this reasonably well: https://youtu.be/v3dz-AKOVL8.
> 
> Using the newer C++17 API I believe makes the library more future-proof than a custom implementation; Once onto C++11 we could switch to an externally developed and supported third-part implementation, and in the distant future when C++17 is the default we can remove entirely the support code.
> 
> This model and API can be 100% implemented in C++11. A large proportion of the most useful parts can be implemented in C++03, and that's what I've implemented here (almost all of it can be ported with heavy template metaprogramming - and indeed Boost has done a lot of this - except the boost implementation requires an extra boost library to be compiled and linked, and I know that we're resistant to those sorts of changes). The parts of the main API that are not implemented are:
> 
> 	• Automatic forwarding of scoped allocators via contruct
> 	• Alignment specification. Some parts of the API allow specification of alignment separately from the C++11 alignas type specifiers, but these are not currently implemented. In debug builds, the alignment is asserted to be the default alignment (16 bytes for 64-bit platforms - which I believe is all we support now)
> 	• Specialized allocators (synchronized_pool_resource, unsynchronized_pool_resource and monotonic_buffer_resource) are not implemented. Although I plan to implement custom allocators to solve our problem, these don't need to adhere to the C++17 public API and thus can be less generic.
> Summary of per-file changes
> 
> All of these files in scitbx/array_family/ have been changed:
> 
> File	Changes
> SConscript	Add tests to compilation
> boost_python/shared_wrapper.h	Choose specific overload of deep_copy to export
> memory.h	Removed - was old implementation to ensure 16-byte alignment
> polymorphic_allocator.h	Implementation of C++17 allocator model
> tests/test_pmr.cpp	Tests of basic allocator model and interface
> tests/test_pmr_shared_plain.cpp	Tests for using allocators in shared_plain
> tests/testing.h	Basic C++ testing framework; uses googletest instead if available
> tests/testing_main.cpp	Contains the main function for the basic test framework
> shared.h	Implement allocator API
> shared_plain.h	Implement allocator API
> versa.h	Implement allocator API
> versa_plain.h	Implement allocator API
> In addition, the newly built test is added to scitbx/run_tests.py.
> 
> Implementation details/Notes
> 
> Some details or notes on caveats/pitfalls/odd code (I can expand with examples if required):
> 
> 	• In order to cope with static globals in a build environment where we don't have .cpp files ignore the capabilities of code-sharing, the shared "global" data is stored inside a static template in order to get around this and the One-definition-rule. It's not ideal, but the only way we have. Luckily, the way python loads modules seems not to cause issues with this across different extension modules.
> 	• There's some slightly sticky behaviour with copy-assignment and swap for the array_family containers when applying to two containers with different allocators. Essentially, because these classes change the semantic meaning of copying to mean sharing common data, potentially assignment could require either a) changing the allocator of the object being assigned to - this is confusing, as if you construct an object with a specific allocator you expect it to use that, or b) an actual data copy operation to move the data to the new allocator. This breaks the array_family semantics that assignment is cheap, and the expectation that you would end up with a shared reference. Therefore, these operations have a hard assert that the allocators are the same. In practice, cases where this might happen can usually be resolved with careful copy-construction instead.
> Testing
> 
> I've added tests for the new code and methods of creating and passing allocators through the shared_plain class directly. Additionally, I've run the following test groups, which I believe should cover things tests both here and over there:
> 
> 	• libtbx.run_tests_parallel for modules libtbx, scitbx, cctbx, iotbx,
> rstbx, dxtbx, dials, dials_regression, xia2 and xia2_regression
> 	• cctbx_regression.test_nightly
> 	• phenix_regression.test_nightly
> I'd love to be advised on any additional test modules or suites that I should be including in this. I'm also aware of our limited variety of platforms and environments we have for testing here, so this needs to be looked at more widely.
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cctbx/cctbx_project/pull/128
> 
> Commit Summary
> 
> 	• Add an C++17-style allocator model to array_family
> File Changes
> 
> 	• M scitbx/array_family/SConscript (20)
> 	• M scitbx/array_family/boost_python/shared_wrapper.h (3)
> 	• D scitbx/array_family/memory.h (105)
> 	• A scitbx/array_family/polymorphic_allocator.h (338)
> 	• M scitbx/array_family/shared.h (44)
> 	• M scitbx/array_family/shared_plain.h (197)
> 	• A scitbx/array_family/tests/test_pmr.cpp (59)
> 	• A scitbx/array_family/tests/test_pmr_shared_plain.cpp (282)
> 	• A scitbx/array_family/tests/testing.h (153)
> 	• A scitbx/array_family/tests/testing_main.cpp (11)
> 	• M scitbx/array_family/versa.h (52)
> 	• M scitbx/array_family/versa_plain.h (47)
> 	• M scitbx/run_tests.py (1)
> Patch Links:
> 
> 	• https://github.com/cctbx/cctbx_project/pull/128.patch
> 	• https://github.com/cctbx/cctbx_project/pull/128.diff
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 
> 

-- 
Paul Adams
Division Director, Molecular Biophysics & Integrated Bioimaging, Lawrence Berkeley Lab
Division Deputy for Biosciences, Advanced Light Source, Lawrence Berkeley Lab
Adjunct Professor, Department of Bioengineering, U.C. Berkeley
Vice President for Technology, the Joint BioEnergy Institute
Laboratory Research Manager, ENIGMA Science Focus Area

Building 33, Room 347
Building 978, Room 4126
Tel: 1-510-486-4225, Fax: 1-510-486-5909
http://cci.lbl.gov/paul

Lawrence Berkeley Laboratory
1 Cyclotron Road
BLDG 33R0345
Berkeley, CA 94720, USA.

Executive Assistant: Louise Benvenue [ LBenvenue@lbl.gov ][ 1-510-495-2506 ]
--

graeme-winter:
There are several very valid points to consider here, which I will try to comment on. While RedHat 5 is not a target platform for dials I do see the need for everything to be very thoroughly tested on that platform to ensure that no harm comes to other software. The particular risk here is compilation errors since these changes shouldn’t change the default behaviour, however I agree that this should be proved rather than asserted. A comprehensive lists of tests to pass would be appreciated here. 

Regarding whether this is a problem and the fact that the code has been in use for 15 years, once again I see your point however Nick has gone to some effort to perform thorough profiling of the code. The outcome of this should be included to demonstrate the need and the underlying thought process. That this is a problem now is less surprising, since dials uses massively more memory and memory bandwidth than most other cctbx programs. While it’s true that we could probably work around this by not using flex arrays, that would complicate the code and move away from some of the core benefits of using the cctbx platform for dials. 

We also need finer grained control over memory for other optimisation for example knights landing, as we’re facing some big data challenges over the forthcoming year with Eiger detectors and serial crystallography. For us performance has become a very serious concern. 

In summary I think there are some important improvements here, that would greatly help, but we need to show that they do no harm to the wider code base.

Best wishes Geaeme

nksauter:
Another observation is that the pull request in itself does not solve the
problem that was originally observed.  Buried in the comments is the
statement that "I plan to implement custom allocators to solve our
problem"--that means it has not been demonstrated that the present pull
request will lead to the desired performance boost.  So it seems premature
to make these major changes now.  I'd prefer to see before & after
performance metrics to assure us that we are on the right track.
Otherwise, we may be having the same conversation again in six months.
Cheers,
Nick

rwgk:
Nick Sauter asked me to take a look.
I only spent 10 minutes.
The proposed change looks good to me. 15+ years ago I actually removed the allocator from the vector<T> implementation I was starting with; just because it was more than I could handle back then. I think it's really great that someone is brave enough to try bringing back a modern allocator.
Here is what I would try:
- pull & immediately start a full phenix build & test
- if something breaks, immediately back out the change and use the logs for offline troubleshooting
- try again some time later

luc-j-bourhis:
A quick comment only about alignment for now. The reason I wrote `memory.h` was to support SSE2. Indeed many SSE2 instructions come in two variants: a slow one which works with data with arbitrary alignment and a fast one which works only with alignment on 16-bit boundary. The slow one is slow enough as to make investing time in vectorisation useless. The fast one crashes on data not aligned on 16-bit boundaries. It was therefore essential to make sure scitbx.array_family arrays were properly aligned. As precisely explained in the comments, and as implemented by the `#ifdef`'s, on 32-bit Linux and Windows, the standard allocators do not align on 16-bit boundaries. Hence `memory.h`.

As far as I understand this pull request, this support for 32-bit system has not been implemented. No production code depends on the feature I implemented in `memory.h`, as my experimentations staid so. However I have a couple of unreleased tools which do depend on that. In this situation, I would understand an executive decision telling me: "well, touch luck, luc!" and going ahead, but I wanted to make sure everybody understands what's involved.

Now, I hear you say, SSE2 is really old and Intel processors have supported for years SSE4 and more importantly AVE, with its twice-longer vector unit. SSE4 is affected by the same issue as SSE2 afaik. Alignment requisites have been "relaxed" for AVE, in the sense that the slower unaligned instructions are not hopelessly slow anymore but still slower. But this is not uniform across all instructions and afaik most people still prefer to align data even for AVE, so as to avoid crossing 4-byte boundaries for doublewords and 8-byte boundaries for quadwords, which means that 16-byte boundaries do the job across all versions of SSE and AVE. I new that when I wrote `memory.h` and designed it accordingly.

ndevenish:
Hi Luc, (Other comments will be visited this week);

As I understand the situation, and everything I've researched reinforces the idea, 16-byte alignment is the "default" e.g. the minimal alignment you get from normal new/malloc, on 64-bit operating  systems. I've also been told that we don't support 32-bit systems any more; you can see given these two facts how the optional 16-byte alignment in memory.h becomes completely redundant and is actually never doing anything, which then becomes completely replaced by the new allocator design.

As Graeme hinted, amongst other things AVX512 operations on e.g. KNL architectures benefit greatly from (IIRC) 64-byte alignment - although this doesn't pass expose the option directly, you can see how the design simply allows replacement with e.g. a memory resource that always uses higher aligned malloc instead of new, dynamically implementable rather than everything at compile-time. Longer-term, the alignment annotations in C++11 solve the problem much better and more generally, as the required alignment can be encoded into the type passed into the container value-type template parameter (this is what the STL implementations rely on - as I'd noted).

Given that the only time we'd ever used the alignment capabilities was to enforce what is now the default, this part of the API I left incomplete and asserted - it can easily be upgraded transparently, as and when we need it, or replaced by a fully-functional backport with the same API in the future.

If absolutely required, the existing compile definitions could just be checked for and diverted to an aligned malloc if you do have important, unreleased tools that rely on 32-bit platforms.

luc-j-bourhis:
Hi Nicholas, if 32-bit system are not supported anymore they I do agree with you. Being able to choose alignment at run-time is also a good thing indeed, as illustrated by the new AVX512 alignment recommendation, which I did not know about.

graeme-winter:
To clarify: 

For DIALS, for high performance computing purposes, 64 bit is explicitly the target platform. 

Clearly 32 bit e.g. RHEL5 continues to be a PHENIX target platform, so maybe this should be left in tact? However the capability for custom allocators could be limited to 64 bit?

ndevenish:
[scitbx memory allocators](https://trello.com/c/6JjJR2oT/26-scitbx-memory-allocators)

graeme-winter:
In conversation with @bkpoon @nksauter @dagewa provided the changes are made, and the new allocators are thread safe, this should all be fine. Berkeley folks do not want google-test like code, want a simple test binary and run this through run_tests.py. Pytest modules for non-python tests in dials should work fine. 

@ndevenish if you make these changes should be good to go, it is agreed.

ndevenish:
I've updated the code to make these requested changes. Summary:

- I've put a more detailed note as to why the `StaticStore` is necessary. Normally, global variable symbols would be declared in a common shared library that is imported by other shared libraries, modules and executables, in order to ensure there is only one location in the process. We certainly have a lot of shared *modules* in the `build/lib` - 262 on a quick count - but these aren't intended to be linked to, only loaded at runtime (some platforms enforce this difference). We only have a handful of libraries, including boost, but nothing that is used or that seems to be intended as a common base - `libcctbx.so` might be the closest. Using templates in this way somewhat bypasses this problem.
- I've also hidden the constructors of `StaticStore`, and added a static assert so that it can't be re(mis)used in other ways. This is a minor technicality, as it shouldn't be used outside of it's local implementation scope anyway (again, not having a library/translation unit scope to hide it in)
- I've changed the tests to use the same approach as the rest of array_family, and removed even the fallback testing infrastructure
- A minor issue with the new way the tests are written ran into [this](https://bugzilla.redhat.com/show_bug.cgi?id=985895) compiler bug, but this only affects the test because of it's standalone nature.

I've also rebased on top of the current `master` branch and tested a full phenix+dials build on RHEL6 (`phenix_regress.test_nightly`, `cctbx_regression.test_nightly`, and our usual dials+xia2 suites of tests), and a full DIALS build on mac. The mac build passes, and the phenix build only fails on things that fail even on master branch (`tst_pipeline_ligand_ncs` at the moment). Since this failure is present on the master branch without this patch, I assume it's one of the "failed CI as todo list" tests.

graeme-winter:
@bkpoon if you’re happy with the changes now, please could you merge?

bkpoon:
Looks good!

PR #133, opened 2018-02-05
simTBX dxtbx fixes

dagewa:
This branch contains changes to the nanoBragg constructor from dxtbx Detector and Beam models. Previously the code appeared to invert the direction of beam propagation, due to an idiosyncrasy of the Beam class. After fixing that, other definitions had to change (`odet_vector`, `pix0_vector` and point of closest approach). I want to request a review from @jmholton to check I have fixed everything required. 

The direction of the rotation axis is also now set, as `init_defaults` sets this along the beam direction for typical dxtbx models.

The other commit makes a minor change to the brunger_example test script.

### DIFF
--- a/simtbx/nanoBragg/brunger_example/run_example.py
+++ b/simtbx/nanoBragg/brunger_example/run_example.py
@@ -61,9 +61,9 @@ with open(stolfile, "rb") as fp:
 # open the existing diffraction image: we need it for the background profile
 import dxtbx
 img = dxtbx.load(imgfile)
-panel = img.get_detector().to_dict()['panels'][0]
-pixel_size_mm = panel['pixel_size'][0]
-distance_mm = -panel['origin'][2]
+panel = img.get_detector()[0]
+pixel_size_mm = panel.get_pixel_size[0]
+distance_mm = panel.get_distance()
 #beam_center_mm =
 
 # create the simulation
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -96,6 +96,10 @@ nanoBragg::nanoBragg(
     Sclose = Yclose = -dot_product(pix0_vector,sdet_vector);
     close_distance = distance =  dot_product(pix0_vector,odet_vector);
 
+    /* set beam centre */
+    scitbx::vec2<double> dials_bc = detector[0].get_beam_centre(beam.get_s0());
+    Xbeam = dials_bc[0]/1000.0;
+    Ybeam = dials_bc[1]/1000.0;
 
     /* detector sensor layer properties */
     detector_thick   = detector[0].get_thickness();
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -107,6 +107,14 @@ nanoBragg::nanoBragg(
 
     //adc_offset = detector[0].ADC_OFFSET;
 
+    /* SPINDLE properties */
+
+    /* By default align the rotation axis with the detector fast direction */
+    spindle_vector[1] = fdet_vector[1];
+    spindle_vector[2] = fdet_vector[2];
+    spindle_vector[3] = fdet_vector[3];
+    unitize(spindle_vector,spindle_vector);
+
     /* NOT IMPLEMENTED: read in any other stuff?  */
     show_params();
 
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -28,8 +28,7 @@ nanoBragg::nanoBragg(
     /* BEAM properties first */
 
     /* direction in 3-space of beam vector */
-    /* is this positive or negative? */
-    xyz = beam.get_direction();
+    xyz = beam.get_unit_s0();
     beam_vector[1] = xyz[0];
     beam_vector[2] = xyz[1];
     beam_vector[3] = xyz[2];
@@ -84,17 +83,17 @@ nanoBragg::nanoBragg(
     sdet_vector[2] = detector[0].get_slow_axis()[1];
     sdet_vector[3] = detector[0].get_slow_axis()[2];
     unitize(sdet_vector,sdet_vector);
-    /* must form an orthonormal system, but dxtbx distance is negative?  */
-    cross_product(sdet_vector,fdet_vector,odet_vector);
+    /* set orthogonal vector to the detector pixel array */
+    cross_product(fdet_vector,sdet_vector,odet_vector);
     unitize(odet_vector,odet_vector);
 
-    /* dxtbx origin is location of the first pixel? */
-    pix0_vector[1] = -detector[0].get_origin()[0]/1000.0;
-    pix0_vector[2] = -detector[0].get_origin()[1]/1000.0;
-    pix0_vector[3] = -detector[0].get_origin()[2]/1000.0;
+    /* dxtbx origin is location of outer corner of the first pixel */
+    pix0_vector[1] = detector[0].get_origin()[0]/1000.0;
+    pix0_vector[2] = detector[0].get_origin()[1]/1000.0;
+    pix0_vector[3] = detector[0].get_origin()[2]/1000.0;
     /* what is the point of closest approach between sample and detector? */
-    Fclose = Xclose = dot_product(pix0_vector,fdet_vector);
-    Sclose = Yclose = dot_product(pix0_vector,sdet_vector);
+    Fclose = Xclose = -dot_product(pix0_vector,fdet_vector);
+    Sclose = Yclose = -dot_product(pix0_vector,sdet_vector);
     close_distance = distance =  dot_product(pix0_vector,odet_vector);
 
 


rjgildea:
> dxtbx Beam.get_direction() rather perversely returns the sample-to-source direction, not the direction of propagation of photons

Isn't this just a result of the definition of the imgCIF coordinate system?
http://www.iucr.org/__data/iucr/cifdic_html/2/cif_img.dic/Caxis.html

>      Axis 3 (Z): The Z-axis is derived from the source axis which goes from
>      the sample to the source.  The Z-axis is the component of the source axis
>      in the direction of the source orthogonal to the X-axis in the plane
>      defined by the X-axis and the source axis.

dagewa:
Hi @rjgildea, in my opinion the method name `Beam.get_direction()` is deficient because it does not clear up ambiguity between the direction of propagation and the sample-to-source direction. Indeed developers using dxtbx may not be aware of the ambiguity at all. If unaware that 'direction' in this context means sample-to-source they might, very reasonably in my opinion, assume that the direction means the direction that the beam travels.

jmholton:
Ok, I'm afraid I'm at something of a loss as to how to test these 
changes.  It's in another branch?  If I pull from that it erases all my 
stuff, right?

The important thing is to make sure the same origin, distance, etc come 
out of dials.refine as were put into the simulation, and that we don't 
invert the hand of the world.  I put anomalous into the 
tst_nanoBragg_basic.py example so that this could be tested. I'm also a 
little concerned that odet_vector point away from the sample  I suppose 
I should just go and write a test that makes a rotation-camera dataset.

For the record, I do agree that beam.get_direction() points in the 
opposite direction of what I think is sensible.  But what do I know?  
I'm new here.  But while I'm complaining, another problem with the beam 
class is that it only has one parameter for divergence.  This is fine 
for rotating anode sources but pretty much all accelerator-based sources 
have a wildly asymmetric divergence. Also, I'm still not sure what the 
units of this divergence are supposed to be  Degrees? millirad?  Should 
the beam have a size as well as a flux?  The flux alone is not enough 
information for simulating diffraction, since scattering and absorption 
both arise from cross sections.  What you need is flux density (flux/size).

I'm confused about some of the detector properties as well (such as the 
units of thickness and "mu"), but in general I am loathe to change 
anything in the base classes because I'm sure that would break everyone 
else's stuff.

Any advice is appreciated.

-James

dagewa:
Hi @jmholton, if you `git pull` from `git@github.com:cctbx/cctbx_project.git` then you will get all the active branches, but you will still be on master. If you have local uncommitted changes you can `git stash` to keep these, then change to this branch by `git checkout simtbx-dxtbx-fixes`. All your stuff is still kept safe by git (inside the `.git` directory) but the source files will now be those from the `simtbx-dxtbx-fixes` branch. You can then do a build and run tests. When finished, just `git checkout master` to go back to where you were, and `git stash pop` if you had uncommitted changes you want to reapply.

Since yesterday I realised that these changes don't actually complete everything I need. The issue is that no matter what I set the beam vector to, the beam centre gets forced to be in the centre of the detector. What I want is for the beam centre implied by the Beam and Detector models to be preserved for the simulation. I will look at that today and push another commit to this branch when I figure it out. I think the issue is just that in the dxtbx world the beam centre doesn't exist as a property of the model, it is always just derived when required from the **s0** vector and Panel objects. For nanoBragg, I need to explicitly set the beam centre and then the source position will be calculated as expected (10m 'upstream' from the origin along the -**s0** direction).

Once I've got that sorted then my script might essentially be the rotation dataset test that you mention (if I switch off the beam movement). I'll try the tests that you suggest.

dagewa:
@jmholton BTW I don't have immediate answers for your other questions about dxtbx models, but I think these are all good things to discuss with the group by email to dials-support@lists.sourceforge.net

dagewa:
@jmholton Now the beam centre is set explicitly and as a result the detector origin `pix0_vector` is preserved through the call to `reconcile_parameters()`. I think that means that the simulation keeps the geometry implied by the `Beam` and `Detector` objects. I will generate a full dataset and check that the output from `dials.refine` is as expected.

I'm puzzled by the appearance of a simulated image though. This is what I see if I rotate the beam by 15 degrees around the slow axis:

simtbx

The beam centre is where I expect it, and I think the beam and detector have the right geometry:
```
  DIRECTION_OF_DETECTOR_X-AXIS= 1 0 0
  DIRECTION_OF_DETECTOR_Y-AXIS= 0 -1 0
  DIRECTION_OF_DETECTOR_Z-AXIS= 0 0 -1
  INCIDENT_BEAM_DIRECTION= -0.34202 0 -0.939693
```
However, I'm not sure why I see the strong spots appearing at high resolution. Am I missing something? Cheers,D

jmholton:
The strong spots beyond the resolution limit are because default_F is 
not set to zero.  So, any hkls not in your input amplitudes (such as 
F000) will still show up.  You can make this go away by not setting 
default_F.

Yes, the "beam center" convention is why I introduced "Pivot". Detector 
rotations are either about the "beam center", or about the sample.  The 
latter is convenient if you want to specify a "2theta" rotation.  
Different programs have different conventions (which is why I introduced 
the "convention" enum setting), and perhaps I don't understand the CCTBX 
one.  Since s0 is just a 3-vector it can only contain information about 
direction, not the source position.  In nanoBragg I arbitrarily 
construct sources at 10m distance from the sample (at 0,0,0), but this 
is a hold-over from the nearBragg code, where the exact distance to the 
source is important.  But anyway, specifying the origin, and therefore 
the pix0_vector and then letting everything else update is the right way 
to do it.  Sorry if that wasn't working.

gotta run...

-James

dagewa:
Ah, thanks for the tip about `default_F`. That makes sense now. The dxtbx beam centre should be returned as a (fast, slow) pair in either mm or pixels depending on the method called. The origin (0,0) is the outer corner of the first pixel. I think I'm getting somewhere with my simulation now. I'll generate a full rotation dataset and check that the refined models are as expected and that we haven't entered the inverted world for the anomalous differences and will report back here.

dagewa:
The first of the suggested tests is fine: after processing the simulated data `dials.refine` gives essentially the same detector origin, fast, slow axes and beam direction as what went into the simulation.

Testing the sign of anomalous differences has not been so convincing as yet. What I did was take the top 10% magnitude anomalous differences from the processed data, find the common set between these and the calculated set used in the simulation, and then calculate what fraction of the common sets have the same sign.

For processing of "noise free" images (no `add_background()` or `add_noise()` calls) it looks okay and I get 86% of the common sets of top magnitude anomalous differences have the same sign after processing as what went into the simulation (1356 reflections in this case). However, when processing the noisy images this drops off to 49.5% (741 reflections), thus inconclusive. I'm not sure why this is, though I might be using cctbx's `miller_array` incorrectly.

Maybe there is a better way to check? I suppose I could try to solve the structure, but it is from the dummy PDB used in tst_nanoBragg_basic.py (5 waters and a selenium in much-too-large cell).

dagewa:
Maybe I shouldn't be surprised, the anomalous signal from the fcalc set is something like 1.3%, but my strongest integrated reflection has only ~2500 counts. As soon as the noise is added the anomalous signal must get swamped out. So maybe the fact that most of the larger anomalous differences in the integrated _no noise_ set have the right sign is enough.

jmholton:
I think if you change sc.fdp = expected_henke.fdp() to sc.fdp = 100 you 
will get a much bigger anomalous signal.

dagewa:
Thanks, that gives me an anomalous signal of 10%, but I still see essentially no correlation (or anti-correlation) between the anomalous differences of the calculated set that went into the simulation and the processed data. In fact, the integrated intensities only have about 30% correlation with the ideal intensities. Something has gone wrong, but I'm confused as to what.

dagewa:
A little more progress: if I set `sc.fdp=5` for each scatterer instead of `100` then the simulated anomalous signal is really high - as much as 46% (funny that it is 10% for `sc.fdp=100` though). After processing the noise free images I have integrated intensities with a CC of 99% with the intensities of the fcalc set going into simulation. So then I look at the CC of the anomalous differences, which is still 98%. Almost all anomalous differences in the set going into the simulation have the same sign as the anomalous differences after data processing.

So that all looks okay. However, after processing the noisy images I only have an overall CC of 41% between the intensities going into the simulation and the integrated intensities (to 2.0 Å). The noise appears to have degraded the intensities badly. Indeed, the CC of the anomalous differences is essentially zero and therefore only about a random half have the same sign after processing as they did going into the simulation.

In the original simulation (without hacking the `fdp` value) the inner shell Mean((I)/sd(I)) is 70 for the noise free images and still as high as 40 after adding noise. The noise didn't seem too worrisome and I didn't change the settings for adding noise was for the later simulations. So I ended up being surprised that it seems enough to completely wipe out such a strong anomalous signal. For looking at data quality with Aimless it hasn't helped that I only generated 90 degrees worth of images for the P 1 structure.

I suppose we don't need the simulation of noisy images to confirm that the anomalous sign is correct, I've just got caught up in trying to understand how difficult this can be to pin down in the presence of noise.

jmholton:
Anomalous correlation to the right answer of 50% or less is actually a 
pretty good signal.  In my experience.

Something you will find really kills the anomalous signal is the 
detector calibration error.  By default I have this set to 3%, which is 
typical for modern detectors.  Best one I've ever seen was ~1%.  To turn 
off calibration error completely, do this:

SIM.detector_calibration_noise_pct=0

And you will find your I/sigma will go through the roof, and phasing by 
anomalous becomes trivial.

-James

dagewa:
Right, I'll give that a try. This is all interesting experience. I also found that I _was_ using `miller.array` incorrectly in some places during the analysis. One issue is that depending on the history I could end up with an array of intensities for which `is_xray_intensity_array()` returns `True`, or in other circumstances `False`. This affected the result of things like `anomalous_differences()` or `anomalous_signal()`. I cleaned that up, but it didn't change the conclusions.

Another thing that caught me out is that when I forced each scatterer to have `fdp=100` I got a _lower_ anomalous signal than for `fdp=5`. This seemed counter-intuitive. I thought perhaps it was an effect of setting fdp for all the scatterers, not just the Se, and having anomalous contributions averaging out over the atoms. Now, going back and setting fdp to 5 or 100 for only the Se atom I get anomalous signals of 13% and 18% respectively.

dagewa:
I found that the main confusion in my 'inverted world' test was that occasionally I would not get consistent indexing. I've now enforced that in the script and the results make sense. The script is in `dials_regression/refinement_test_data/varying_beam_direction/sim_images.py`. I ran this with the changes here in this `simtbx-dxtbx-fixes` branch. It simulates a 90° rotation sweep, processes with DIALS then compares the anomalous differences in the processed data with the differences in the ideal structure factors that were fed into the simulation.

The anomalous signal in the ideal amplitudes is 1.3%, but we have the advantage of the 'noise free' images images `intimage_*.img`, so even small differences should be preserved in the processed data. I get a CC of 99% between all intensities common in the processed and ideal data sets, so I know processing hasn't gone wrong and that indexing is consistent. The CC of anomalous differences is 90% and 90% of the differences have the same sign going in as they do coming out.

That is enough to conclude that the changes do not invert the world of the simulation. I could leave it there, but I had the noisy images `noiseimage_*.img` anyway so I did the same check with them. The CC between all intensities is still 99% (no indexing issue) but the CC between anomalous differences drops to 37%, with 66% of the differences having the same sign. Not too bad (this is after I boosted signal to noise in the simulation by increasing the crystal size and reducing detector calibration error to 1%)

I was curious to see what happens if I hacked in a greater anomalous signal. By setting the `fdp` of every scatterer to 5 I found that the anomalous signal in the ideal amplitudes jumps to 34%. This should be clearly detected in the noisy images. In that case, the CC between all intensities is 98%, the CC between anomalous differences is 99% and 96% of these anomalous differences have the same sign. So yes, it is very clear.

That was a bit of a saga but I was learning all the while. My conclusion is that the `simtbx-dxtbx-fixes` changes are good and allow simTBX to correctly interpret dxtbx Detector and Beam models. In future, adding another constructor that takes in a full dxtbx Experiment would be useful, as the code still makes assumptions about things like the orientation of the rotation axis. However, that's for another day.

One thing I haven't been able to test is the effect these changes have on `brunger_example/run_example.py`. That also uses the nanoBragg constructor from dxtbx Detector and Beam, so these changes would affect the coordinate system that runs in. Unfortunately when I tried testing this in the `master` branch it ran for more than a day before crashing out with a Bus error.

I've added @nksauter as a proposed reviewer in case he wants to comment.

nksauter:
David,

Please rebase against the master.  As previously discussed, it was necessary for me to restore my files that had been deleted by accident.  I can't evaluate the pull request without having my restored files side by side with your new code. --Nick

Anthchirp:
Rebased as requested
for reference: I ran
```bash
$ git fetch
$ git checkout simtbx-dxtbx-fixes
$ git pull --rebase
$ git rebase origin/master
$ git push --force-with-lease # a more careful version of 'git push -f'
```

dagewa:
Thanks Markus

nksauter:
simtbx/nanoBragg/brunger_example/run_example.py
<https://github.com/cctbx/cctbx_project/pull/133#diff-0276e38826a6c2354e76ea6faf078544>
Please change line 65 to

pixel_size_mm = panel.get_pixel_size()[0]

However, the example still does not run.  James, can you look into this?
The screen output stops moving at

bin= 999 stol= 0.252963 avg= 5.16406 rmsd= 0.647709  n= 1236

bin= 1000 stol= 0.253194 avg= 5.1429 rmsd= 0.651557  n= 1177

bin= 1001 stol= 0.253425 avg= 4.87426 rmsd= 0.713545  n= 10877

WARNING: not enough pixels in bin= 1002 n=0 stol= 1e+88 median= 0 avg_arej=
nan

WARNING: not enough pixels in bin= 1003 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

WARNING: not enough pixels in bin= 1004 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

WARNING: not enough pixels in bin= 1005 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

WARNING: not enough pixels in bin= 1006 n=0 stol= 1e+88 median= 0 avg_arej=
nan

WARNING: not enough pixels in bin= 1007 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

WARNING: not enough pixels in bin= 1008 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

WARNING: not enough pixels in bin= 1009 n=0 stol= 6.90876e-310 median= 0
avg_arej= nan

done with radial median filter

 about to initialize stol,Fbg

 done

amorphous_molecules= 2.10775e+09 in beam: 9e-14 m^3 * 0.7 g/m^3 *
6.02214e+23 /mol / 18 g/mol

initializing Fbg with pythony array 1000

done initializing stol_of and Fbg_of

dispsteps= 6

divsteps= (2, 2)

oversample= 1

1782 1832 : stol = 5.28693e+08 or 0.558788

 F=2.11716e-53    I = 0.0416667

I/steps   0.04166666667

polar      0.9959670032

omega   6.806201671e-08

pixel       2.235883207


solid angle subtended by detector = 0.912723 steradian ( 7.26322% sphere)

max_I= 4.29052 @ ( 0.160557, 0.182297) sum= 3.57836e+07 avg= 2.13287

mid_sample= 2.22602381863

dispsteps= 6

divsteps= (2, 2)

oversample= 1

scaling data by: intfile_scale = 1

writing intimage_001.img as 2-byte integers; sum= 2.87161e+07 max= 4 @ (
2015 2253)

raw CELL 68.780000 169.260000 287.420000 1.570796 1.570796 1.570796   1

CELL 68.780000 169.260000 287.420000 1.570796 1.570796 1.570796   1

MISSET 1.692050 -0.909198 -0.567546

MISSET 1.692050 -0.909198 -0.567546

Unit Cell: 68.78 169.26 287.42 90 90 90

Recp Cell: 0.0145391 0.00590807 0.00347923 90 90 90

volume = 3.34606e+06 A^3

real-space cell vectors (Angstrom):

     a           b           c

X: 35.63203832 -122.78950991 -130.24262664

Y: -22.71585247 54.00022816 -255.32509779

Z: 54.26818858 103.22625254 -21.35905066

reciprocal-space cell vectors (Angstrom^-1):

     a_star      b_star      c_star

X:  0.00753210 -0.00428600 -0.00157659

Y: -0.00480181  0.00188489 -0.00309072

Z:  0.01147152  0.00360314 -0.00025855

TESTING sincg(1,1)= 1.000000

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

jmholton:
Yes, it's still rendering. Should take about 4 hours.  To make it go 
faster put the word "fast" as a command-line argument, then it's only 
about 40 seconds, but less realistic.

The reason why the rendering takes so long is because this is a very 
complex image.  4k x 4k pixels, three consecutive calls to 
add_nanoBragg_spots with slightly different master orientations, each 
with 20 mosaic domains, 6 wavelength steps and 4 divergence steps, which 
is 1440 times slower than doing just one call with one domain, one 
wavelength and one beam direction.

I note just now that my progress meter seems to not be working. Will 
look into it.

-James

Anthchirp:
fixed the merge conflict and merged as requested by @dagewa

PR #142, opened 2018-02-21
Changes to be consistent with the current nexus format.

asmit3:
Implementing these changes and bringing nexus.py in line with the NeXus format. 
1. Drop _F, _S vectors (including the depends_on list and transformations)
2. Use fast/slow_pixel_direction instead (single entry equal to pixel size) and use module_offset to position these vectors
3. Drop fast_pixel_size
4. Flatten module list

Concurrent changes have been made to dials_regression/test.py to skip testing of LCLS_cspad_nexus master file which is currently inconsistent with the NeXus format. Thus you would need to update dials_regression repository to pass all tests there. 
Work done with @phyy-nx

### DIFF
--- a/dxtbx/format/FormatHDFEigerNearlyNexus.py
+++ b/dxtbx/format/FormatHDFEigerNearlyNexus.py
@@ -156,7 +156,7 @@ class EigerNXmxFixer(object):
       handle['/entry/instrument/detector/x_pixel_size'].value)
     group['fast_pixel_direction'].attrs['transformation_type'] = 'translation'
     group['fast_pixel_direction'].attrs['vector'] = fast_axis
-    group['fast_pixel_direction'].attrs['offset'] = 0
+    group['fast_pixel_direction'].attrs['offset'] = (0, 0, 0)
     group['fast_pixel_direction'].attrs['units'] = "m"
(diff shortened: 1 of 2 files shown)


dxtbx/format/FormatHDFEigerNearlyNexus.py
```diff
@@ -156,7 +156,7 @@ def create_scalar(handle, path, dtype, value):
       handle['/entry/instrument/detector/x_pixel_size'].value)
     group['fast_pixel_direction'].attrs['transformation_type'] = 'translation'
     group['fast_pixel_direction'].attrs['vector'] = fast_axis
-    group['fast_pixel_direction'].attrs['offset'] = 0
+    group['fast_pixel_direction'].attrs['offset'] = (0, 0, 0)
```
jmp1985: It's been a while since I looked at the Nexus stuff. Is offset really a vector rather than a scalar? Looking at http://download.nexusformat.org/sphinx/classes/applications/NXmx.html it isn't obvious to me.

phyy-nx:
http://download.nexusformat.org/sphinx/classes/base_classes/NXdetector_module.html#nxdetector-module
has a bit more information than the NXmx link:

*@offset*: (optional) NX_NUMBER
<http://download.nexusformat.org/sphinx/nxdl-types.html#nx-number>

A fixed offset applied before the transformation (three vector components).

PR #147, opened 2018-03-05
phil: add allow_none (default=True) attribute for int/float types

rjgildea:
This commit adds the "allow_none" attribute for int/float types (c.f. allow_none_elements for ints/floats).

This would allow a developer to explicitly prevent a user from setting "None" as an allowable value for a given phil parameter. See e.g.:

https://github.com/dials/dials/issues/516

```
$ git diff
diff --git a/algorithms/spot_finding/factory.py b/algorithms/spot_finding/factory.py
index 9420d6e..930fa38 100644
--- a/algorithms/spot_finding/factory.py
+++ b/algorithms/spot_finding/factory.py
@@ -73,7 +73,7 @@ def generate_phil_scope():
       max_spot_size = 100
         .help = "The maximum number of contiguous pixels for a spot"
                 "to be accepted by the filtering algorithm."
-        .type = int(value_min=1)
+        .type = int(value_min=1, allow_none=False)
 
       max_separation = 2
         .help = "The maximum peak-to-centroid separation (in pixels)"
```

```
$ dials.find_spots $(libtbx.find_in_repositories dials_regression)/centroid_test_data/centroid_0001.cbf max_spot_size=None
...
RuntimeError: Please report this error to dials-support@lists.sourceforge.net: spotfinder.filter.max_spot_size cannot be None
```

### DIFF
--- a/libtbx/phil/__init__.py
+++ b/libtbx/phil/__init__.py
@@ -299,11 +299,13 @@ class number_converters_base(_check_value_base):
 
   def __init__(self,
       value_min=None,
-      value_max=None):
+      value_max=None,
+      allow_none=True):
     if (value_min is not None and value_max is not None):
       assert value_min <= value_max
     self.value_min = value_min
     self.value_max = value_max
+    self.allow_none = allow_none
 
   def __str__(self):
     kwds = []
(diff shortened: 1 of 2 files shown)


libtbx/phil/__init__.py
```diff
@@ -299,33 +299,47 @@ class number_converters_base(_check_value_base):
...
     self.value_max = value_max
+    self.allow_none = allow_none
 
   def __str__(self):
     kwds = []
     if (self.value_min is not None):
       kwds.append("value_min=" + self._value_as_str(value=self.value_min))
     if (self.value_max is not None):
       kwds.append("value_max=" + self._value_as_str(value=self.value_max))
+    if (self.allow_none):
```
Anthchirp: no need for brackets around the entire if conditional

PR #154, opened 2018-03-26
Implementation of TLS-analysis in c++

nicholas-pearce:
Reimplementation of the tis-analysis of mmtbx/tls/analysis.py in c++ for speed. Compares at ~0.6s in python to 0.002s in c++. Tests implemented against the original mmtbx/tls/analysis.py. Also identified and fixed an error in the original mmtbx/tls/analysis.py.

This is my first commit to the cctbx so you may wish to check this code more thoroughly than you might do for others -- feedback is appreciated to make sure I'm doing things properly.

### DIFF
--- a/mmtbx/regression/tls/tst_tls_decompose.py
+++ b/mmtbx/regression/tls/tst_tls_decompose.py
@@ -1,0 +1,244 @@
+from __future__ import division
+import libtbx.utils
+from libtbx.test_utils import approx_equal
+import scitbx.random
+from scitbx import matrix
+from scitbx.math import basic_statistics
+from scitbx.array_family import flex
+import itertools
+import math, time, sys
+
+from libtbx.utils import null_out, multi_out
+
+from mmtbx.tls import analysis as tls_analysis
+from mmtbx.tls.decompose import decompose_tls_matrices
+
+d2r = math.pi/180
+
+def dot(a,b):
+  return a[0]*b[0] + a[1]*b[1] + a[2]*b[2]
+
+def generate_random_tls_motions(g):
+  """Given a random number generator, return tls motions"""
+
+  # Generate random input
+  tx, ty, tz = sorted(abs(g(3))*2.00 + 0.10, reverse=True)      # on order of 1
+  dx, dy, dz = sorted(abs(g(3))*0.01 + 0.01, reverse=True) # on order of
+  sx, sy, sz = g(3)*5.0                             # on order of 5, don't need sorting, can be negative
+  # Generate three rotation axes
+  l_x = flex.vec3_double([tuple(g(3))])
+  l_y = flex.vec3_double([tuple(g(3))])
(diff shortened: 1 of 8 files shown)


Anthchirp:
Hi Nick,

obviously can't comment on the main point of the mmtbx changes but just two comments.

You have two merge commits as part of your pull request. Could you please get rid of them? The easiest and best way to do this would be to simply rebase the pull request branch on top of cctbx:master. I've tested it, and it should works cleanly. (In future you may find it useful to create separate branches for pull requests, so what you do on your master branch doesn't affect the pull request - currently it does)

The other comment is that it may be of interest to look into using pytest for tests, although I think mmtbx does not support those yet. But that would allow you to get rid of helper constructs such as
```python
try:
  (...)
except Exception as e:
  print e
  raise
```
In case the mmtbx folks are interested in enabling pytests - [here is how](https://github.com/dials/dials/wiki/Why-pytest%3F#enabling-pytests-for-a-cctbx-module).

nicholas-pearce:
Hi,

I will try to rebase. Will find instructions on the google.

I am happy (and would prefer!) to write pytests. Your instructions are pretty good for doing so, so I think I will. The try except that you show is actually from trying to debug the test (there was also a bug in the original code I was reimplementing) and I should remove it.

Anthchirp:
Instructions to rebase should be approximately
```bash
git remote -v
# You should have two entries here, 'origin' pointing to your private repository and 'upstream' pointing to the original cctbx repository

git checkout master # your master branch
git rebase upstream/master # do the rebase
git log -6 # check if results are sensible
git push -f # update pull request
```

pafonine:
> The other comment is that it may be of interest to look into using 
> pytest for tests, 

I wrote the original code that Nick modified, I still occasionally work 
on it and I maintain the code and tests. I won't support it if it starts 
using a different testing method.

Pavel

nicholas-pearce:
Ok. To be clear I was not suggesting re-writing your tests, only my new test. 

I can therefore...
a) Leave my new test in the libtbx framework.
b) re-write my new test in pytest and then make them libtbx-compatible (as detailed in Markus' link).

I have a mild preference for b) because I am more familiar with pytest. However, if you would prefer a), I am happy with that and I will leave as is. Which would you like me to do?

pafonine:
Nick,

a very comprehensive set of tests for this code is located here:

mmtbx/regression/tls

So if your C++ implementation mirrors the original Python implementation 
then it must pass these tests.

All the best,
Pavel

pafonine:
P.S.:
If C++ and Python implementations are redundant then one needs to be 
removed.
Pavel

nicholas-pearce:
Yes, I agree.

I will replace the python implementation with the c++ one and ensure all tests run correctly with the c++ implementation. I will then issue another pull request.

Thanks,
Nick

PR #161, opened 2018-04-03
c++ implementation of TLS analysis

nicholas-pearce:
Have replaced the python implementation in mmtbx/tls/analysis.py with c++ implementation. All relevant tests in mmtbx/regression/tls run correctly.

### DIFF
--- a/mmtbx/regression/tls/tst_formula_t_S_10_vs_11_2igd.py
+++ b/mmtbx/regression/tls/tst_formula_t_S_10_vs_11_2igd.py
@@ -405,8 +405,8 @@ if (__name__ == "__main__"):
       r = exercise_00(pdb_str=pdb_str, formula=formula)
       print "  ", i, r
       if(formula=="10"):
-        if(i==0): r<0.04
-        if(i==1): r>0.08
+        if(i==0): assert r<0.04
+        if(i==1): assert r>0.08
       if(formula=="11"):
-        if(i==0): r<0.02
-        if(i==1): r<0.02
+        if(i==0): assert r<0.02
+        if(i==1): assert r<0.02
--- a/mmtbx/regression/tls/tst_formula_t_S_10_vs_11_4muy.py
+++ b/mmtbx/regression/tls/tst_formula_t_S_10_vs_11_4muy.py
@@ -277,5 +277,5 @@ if (__name__ == "__main__"):
     print "formula:", formula
     r = exercise_00(pdb_str=pdb_str, formula=formula)
     print "  ", r
-    if(formula=="10"): r>0.55
-    if(formula=="11"): r<0.06
+    if(formula=="10"): assert r>0.55
+    if(formula=="11"): assert r<0.06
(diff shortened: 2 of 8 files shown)


olegsobolev:
Hi,

I can testify that this pull-request passes t96 tests (bigger set including
Phenix private tests).

Oleg.

nicholas-pearce:
Hi -- can this pull-request be merged or do I need to do something else?

nicholas-pearce:
What is the next stage for this pull request? Any reason it cannot the merged?

graeme-winter:
@olegsobolev @nicholas-pearce any progress made on this other than assigning to @pafonine? Am going to guess that this will wait on the currently-in-progress PHENIX release but is there anything more which should be done before it gets merged?

nicholas-pearce:
I have recently started working on this again, so there may be a few changes/additions in the coming weeks (such as making it clang-compatible on Mac). It therefore probably makes sense to leave the pull-request until I've finished with that. I don't mind if the pull-request stays open until then, or if we close this one and open another when I'm done.

nicholas-pearce:
Hi Pavel,

First of all, I completely agree that I should have contacted you before I started on the reimplementation. I will do so in the future. 

The most helpful question as to determining a path forward might be: what re-factoring were you planning on doing? I don't know whether you looked at all at my reimplementation, but I have translated your code for the decomposition steps to c++ literally and then plugged it back into your original python class. Therefore, if the refactoring you were planning on doing was more on the surrounding class(es) rather than the number-crunching steps, then it might make more sense to do the re-factoring starting from the new implementation rather than duplicate the job by doing it first on the python code and then mirroring it in the c++ code. As a trivial example, my class already captures the error messages rather than raising an exception (I am currently re-raising the exception at the python level for compatibility with the preceding implementation).

However, I appreciate that you have clearly stated that you don't have time to check for things now, so I would like to suggest a middle way that should avoid problems, and allow me to have my code in cctbx, though it does leave a code-duplication-annoyance in the short term...

1) I issue a new pull request for the c++ implementation only (decompose.h, decompose.cpp, etc), leaving your python implementation as it is. The code is fully self-contained so it won't affect anything on either end.
2) We discuss the re-factoring you were planning on doing, so that we can possibly split the workload (I am happy to help).
3) You add tests and you/we refactor (in accordance with decisions from 2) when you/we have time.
4) The implementations are merged as appropriate.

How does that sound?

Nick

PS: There is an error in the python check_33_34_35 function (line 490) for when you write the tests.

Anthchirp:
How about putting both implementations in, making it selectable which one you want? Then can validate them against one another and potentially run the same tests on both.

(Just a suggestion, am not really invested in this)

ndevenish:
> How about putting both implementations in, making it selectable which one you want?

This sounds sensible, e.g. instead of replacing the contents of the class `run`, make a separate `TLSAnalysisRun` class that duplicates and changes what it needs, then a controllable switch somewhere. 

Only real option if the tests can't be trusted.

PR #175, opened 2018-04-17
dxtbx/model/crystal.h: add constructor from A matrix

rjgildea:


nksauter:
Richard,

In my crystal_orientation code in cctbx I was careful to specify in the
constructor whether the A matrix was expressed in direct space or
reciprocal space.  I'd suggest being more clear about it here.

Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

rjgildea:
@nksauter as suggested I've added a parameter `reciprocal` (default = true) to the constructor:
https://github.com/cctbx/cctbx_project/pull/175/commits/b113ae1bbdefbdd4adb9275b51f36d205b342082

dagewa:
@nksauter What is **A** expressed in real space? I note that in our [conventions](https://dials.github.io/documentation/conventions.html) the **B** matrix is defined as a _transpose_ of the inverse of the real space orthogonalisation matrix. If "**A** expressed in real space" is supposed to be the orthogonalisation matrix then should the setter be `set_A(A.inverse())` or `set_A(A.inverse().transpose())`?

I have always thought of the symbol **A** as being reserved for reciprocal space, avoiding confusion.

nksauter:
David,

Please see Sauter (2006) J Appl Cryst 39: 158 for a treatment of the A
matrix viewed in both direct and reciprocal space.  Maybe it's time to
update the dials conventions to allow both.

Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

dagewa:
Hi @nksauter,

I see. Equations (3) and (4) of that paper make the transpose relationship clear, where **A*** has the basis expressed as column vectors, while **A** has the basis as row vectors. I don't know if it is fair to call both of these "orthogonalisation matrices". If we are happy with the Sauter 2006 definitions then we should update the conventions document and consider renaming a few things.

Much of the terminology of that document came from Kevin Cowtan's [slides](https://www.iucr.org/__data/assets/pdf_file/0009/7011/19_06_cowtan_coordinate_frames.pdf) from which I also took the advice "Don't mix conventions between real and reciprocal space – it will only end in tears"

Cheers, David

nksauter:
Yes, equations (3) and (4) are essential for correct understanding of the
problem.

Just a caution, my request wasn't to interfere with understanding the word
"orthogonalization" or to update the web document.

I was only concerned that we should provide an API so Crystal can be
instantiated from either reciprocal or direct-space basis vectors, in the
form of an A-star or A matrix, respectively, whichever the developer
desires.

Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

graeme-winter:
Would make_crystal_with_UB etc. avoid the ambiguity? i.e. following the established conventions in http://journals.iucr.org/d/issues/2014/10/00/kw5106/index.html without worrying about A per se

PR #179, opened 2018-04-20
Xtc speedup

asmit3:
This pull request introduces and modifies the following features in dxtbx to speed up processing of XTC streams at LCLS - 

1. Add FormatMultiImageJIT (JIT = Just-in-Time) and change FormatXTC to inherit from that class. This allows us to process large XTC streams without loading/setting all the models upfront as is the case with FormatMultiImage. The models are set only when requested by a process. Because of this, a new imageset, ImageSetJIT is necessary since some of the ImageSet features need to be overriden (eg. get_detector etc). The __getitem__ function for ImageSetJIT is also defined since a model (detector,beam etc) for an image needs to be set before further action takes place in imageset.h
Most of the features introduced have little to no duplication of code. The class method get_imageset() function in FormatMultiImage has a new flag **just_in_time** that allows us to implement FormatMultiImageJIT without duplicating code. 

Note, to do this change, in ImageSetJIT, get_format_class(). __current_instance__ has been used. 

2. Incorporate functionality for handling multiple LCLS runs with FormatXTC by using a dictionary to map runs to their events (through the run.times() method)

3. idx mode is made default in all XTC readers. This allows better parallelism when using dials.stills_process

4. Caching has been used to improve FormatXTC code speed. The PSANA detector object from within a run is cached, so is the FormatXTCJungfrau detector object (currently no information is available about parallax effects of the Jungfrau). 

5. New test in dxtbx/test/tst_imageset.py to test the imageset obtained from loading a multiimage file. The tests are done both with and without **just_in_time**=True. 

6. A concomitant pull request has been created in the dials source for utilizing these changes.

By 
Asmit Bhowmick & Aaron Brewster

### DIFF
--- a/dxtbx/format/FormatMultiImage.py
+++ b/dxtbx/format/FormatMultiImage.py
@@ -153,7 +153,7 @@ class FormatMultiImage(object):
                    format_kwargs=None,
                    template=None,
                    check_format=True,
-                   just_in_time=False):
+                   lazy=False):
     '''
     Factory method to create an imageset
 
@@ -229,7 +229,7 @@ class FormatMultiImage(object):
       else:
         is_sweep = False
 
-    assert not (as_sweep and just_in_time), 'No Just-in-Time support for sweeps'
+    assert not (as_sweep and lazy), 'No lazy support for sweeps'
 
     if single_file_indices is not None:
       single_file_indices = flex.size_t(single_file_indices)
@@ -237,11 +237,11 @@ class FormatMultiImage(object):
     # Create an imageset or sweep
     if not is_sweep:
 
-      # Use imagesetJIT
-      # Setup ImageSetJIT and just return it. No models are set.
-      if just_in_time:
-        from dxtbx.imageset import ImageSetJIT
-        iset = ImageSetJIT(
+      # Use imagesetlazy
+      # Setup ImageSetLazy and just return it. No models are set.
+      if lazy:
+        from dxtbx.imageset import ImageSetLazy
+        iset = ImageSetLazy(
           ImageSetData(
             reader = reader,
             masker = masker,
--- a/dxtbx/format/FormatMultiImage.py
+++ b/dxtbx/format/FormatMultiImage.py
@@ -175,9 +175,8 @@ class FormatMultiImage(object):
     if format_kwargs is None:
       format_kwargs = {}
 
-    # If we have no specific format class, we need indices for number of images
-    from dxtbx.format.FormatMultiImageJIT import FormatMultiImageJIT
-    if Class in [FormatMultiImage, FormatMultiImageJIT]:
+    # If get_num_images hasn't been implemented, we need indices for number of images
+    if Class.get_num_images == FormatMultiImage.get_num_images:
       assert single_file_indices is not None
       assert min(single_file_indices) >= 0
       num_images = max(single_file_indices) + 1
--- a/dxtbx/format/FormatMultiImage.py
+++ b/dxtbx/format/FormatMultiImage.py
@@ -176,8 +176,8 @@ class FormatMultiImage(object):
       format_kwargs = {}
 
     # If we have no specific format class, we need indices for number of images
-    import inspect
-    if FormatMultiImage in inspect.getmro(Class):
+    from dxtbx.format.FormatMultiImageLazy import FormatMultiImageLazy
+    if Class in [FormatMultiImage, FormatMultiImageLazy]:
       assert single_file_indices is not None
       assert min(single_file_indices) >= 0
       num_images = max(single_file_indices) + 1
--- a/dxtbx/format/FormatMultiImage.py
+++ b/dxtbx/format/FormatMultiImage.py
@@ -153,7 +153,7 @@ class FormatMultiImage(object):
                    format_kwargs=None,
                    template=None,
                    check_format=True,
-                   lazy=False):
+                   just_in_time=False):
     '''
     Factory method to create an imageset
 
@@ -176,8 +176,8 @@ class FormatMultiImage(object):
       format_kwargs = {}
 
     # If we have no specific format class, we need indices for number of images
-    from dxtbx.format.FormatMultiImageLazy import FormatMultiImageLazy
-    if Class in [FormatMultiImage, FormatMultiImageLazy]:
+    from dxtbx.format.FormatMultiImageJIT import FormatMultiImageJIT
+    if Class in [FormatMultiImage, FormatMultiImageJIT]:
       assert single_file_indices is not None
       assert min(single_file_indices) >= 0
       num_images = max(single_file_indices) + 1
@@ -230,7 +230,7 @@ class FormatMultiImage(object):
       else:
         is_sweep = False
 
-    assert not (as_sweep and lazy), 'No lazy support for sweeps'
+    assert not (as_sweep and just_in_time), 'No Just-in-Time support for sweeps'
 
     if single_file_indices is not None:
       single_file_indices = flex.size_t(single_file_indices)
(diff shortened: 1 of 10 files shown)


graeme-winter:
I've asked Nick and Markus to also look at this with a technical hat on, as this looks like rather a fundamental change (in particular, we use the multi reader for Eiger stuff, too)

I understand that the TL;DR version of the description above is essentially "lazy loading" - is this correct? Seems like a useful thing to have...

dxtbx/format/FormatMultiImageJIT.py
```diff
@@ -0,0 +1,38 @@
...
+                                                           detector,
+                                                           goniometer,
+                                                           scan,
+                                                           as_sweep,
+                                                           as_imageset,
+                                                           single_file_indices,
+                                                           format_kwargs,
+                                                           template,
+                                                           check_format,
+                                                           just_in_time)
```
Anthchirp: Must use keyword arguments here

dxtbx/format/FormatXTC.py
```diff
@@ -96,14 +99,27 @@ def params_from_phil(master_phil, user_phil):
...
-    for nevent,evt in enumerate(self._ds.events()):
-      wavelength = cspad_tbx.evt_wavelength(evt)
-      if wavelength is None: continue
-      self.events_list.append(evt)
+    self.times = []
+    self.run_mapping = {}
+    for run in self._ds.runs():
+      times = run.times()
+      self.run_mapping[run.run()] = len(self.times), len(self.times)+len(times),run
+      self.times.extend(times)
```
Anthchirp: Can this function please have a docstring?

dxtbx/tests/tst_imageset.py
```diff
@@ -659,6 +659,43 @@ def run(self):
...
+
+    for i in range(len(iset)):
+      data = iset.get_raw_data(i)
+#      mask = iset.get_mask(i)
+      b = iset.get_beam(i)
+      d = iset.get_detector(i)
+      g = iset.get_goniometer(i)
+      s = iset.get_scan(i)
+
+    print 'OK'
```
Anthchirp: Never ```print "OK"``` in tests. This is a libtbx antipattern that must be eradicated. It is directly implicated in tests being fundamentally broken.

dxtbx/tests/tst_imageset.py
```diff
@@ -659,6 +659,43 @@ def run(self):
...
+    assert len(iset) == 4
+    for i in range(len(iset)):
+      data = iset.get_raw_data(i)
+#      mask = iset.get_mask(i)
+      b = iset.get_beam(i)
+      d = iset.get_detector(i)
+      g = iset.get_goniometer(i)
+      s = iset.get_scan(i)
+
+    print 'OK'
```
Anthchirp: as below

dxtbx/tests/tst_imageset.py
```diff
@@ -659,6 +659,43 @@ def run(self):
 
     print 'OK'
 
+class Test_SACLA_MPCCD_Cheetah_File(object):
```
Anthchirp: Replace class with test function.

dxtbx/format/FormatMultiImageJIT.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import absolute_import, division
```
Anthchirp: *, print_function*
(you don't have to add this, it is enforced by a test anyway. But good habit.)

dxtbx/format/FormatMultiImage.py
```diff
@@ -230,13 +230,29 @@ def get_imageset(Class,
...
     # Create an imageset or sweep
     if not is_sweep:
 
+      # Use imagesetJIT
+      # Setup ImageSetJIT and just return it. No models are set.
+      if just_in_time:
+        from dxtbx.imageset import ImageSetJIT
+        iset = ImageSetJIT(
+          ImageSetData(
+            reader = reader,
```
Anthchirp: no spaces around '=' in keyword arguments. Helps distinguish from assignments.

dxtbx/format/FormatMultiImage.py
```diff
@@ -177,7 +176,8 @@ def get_imageset(Class,
       format_kwargs = {}
 
     # If we have no specific format class, we need indices for number of images
-    if Class == FormatMultiImage:
+    from dxtbx.format.FormatMultiImageJIT import FormatMultiImageJIT
+    if Class in [FormatMultiImage, FormatMultiImageJIT]:
```
Anthchirp: I think this case should be identified by a property of the class, not by yet another import and comparing to a list.

dxtbx/tests/tst_imageset.py
```diff
@@ -659,6 +659,43 @@ def run(self):
...
+    from dxtbx.format.Registry import Registry
+    format_class = Registry.find(self.filename)
+
+    iset = format_class.get_imageset([self.filename], just_in_time=just_in_time)
+
+    assert len(iset) == 4
+    for i in range(len(iset)):
+      data = iset.get_raw_data(i)
+#      mask = iset.get_mask(i)
+      b = iset.get_beam(i)
```
Anthchirp: if there is nothing else to be done with ```b``` then change to
```assert iset.get_beam(i)```
to guarantee a truthy return value. Same for the following 7 assignments

**Comment by graeme-winter** [comment]
In conversation with @ndevenish @jmp1985 @Anthchirp @dagewa I understand that this is intended to be merged _today_ and is much larger than @jmp1985 and @ndevenish had previously understood. 

For a pull request of this size more than a couple of hours will be needed... these seem deeper changes than just extending FormatXTC

**Comment by phyy-nx** [comment]
Understandable. If it helps, the only format class to use these changes is FormatXTC. None of the other format classes are affected. 

We need these changes for an experiment in a week, but we can use the xtc_speedup branch for that experiment if necessary. 

Thanks for looking at this folks,

-Aaron

**Comment by phyy-nx** [comment]
Comments have been addressed, thanks.

**Comment by phyy-nx** [comment]
@graeme-winter the thought for params.mode in FormatXTC was that since the folks working on psana create new data access modes on occasion, we don't know the full list.  That's why we did it as a str instead of a choice.  Regardless, there's more to do here in the future, as the data reading psana interfaces change depending on the access mode, but we can adapt the code as needed.

File: dxtbx/tests/tst_imageset.py [hunk 10]
```diff
@@ -614,7 +614,39 @@ def run(self):
       g = iset.get_goniometer(i)
       s = iset.get_scan(i)
 
+class Test_SACLA_MPCCD_Cheetah_File(object):
```
Anthchirp: This test class should still be a function instead of a class. Test classes that only have ```__init__``` and a single function and no real state add unnecessary complexity

dxtbx/imageset.py
```diff
@@ -148,6 +148,61 @@ def paths(self):
     return [self.get_path(i) for i in range(len(self))]
 
 
+class ImageSetLazy(ImageSet):
```
jmp1985: A while back, I did some work to try and put the ImageSet/ImageSweep classes into C++ so that the image data could be read both in C++ and Python. It looks like ImageSetLazy just has a python interface at the moment. As it inherits from the ImageSet class, passing an ImageSetLazy object into a C++ function expecting an ImageSet object should work in that it will treat it as an ImageSet object. 

However, I'm just wondering what will happen if the C++ function calls get_detector on this object? I don't think these overloaded methods won't be called unless we do something in boost python (https://www.boost.org/doc/libs/1_61_0/libs/python/doc/html/tutorial/tutorial/exposing.html#tutorial.exposing.class_virtual_functions). I suppose one question would be whether we expect this to be needed; however, I think it would be desirable to ensure that the ImageSetLazy class also works with C++ functions.

phyy-nx: You are correct about get_detector not working with ImageSetLazy through C++ unless the python version has been called first.  However, we grepped the sources for this use case and didn't find it.

PR #186, opened 2018-04-27
convert remaining dxtbx tests to pytest

Anthchirp:
To avoid any conflicts this patch will only land after the XTC speedup pull request has landed.


Anthchirp:
Squashed commits as far as possible and changed the useless commit message to something more useful explaining what is going on and why.

PR #198, opened 2018-05-24
boost.python changes for python 3

Anthchirp:
The first commit 82633df is simply syntax rewriting.

The second commit 10f4e8ad  adds a new ```boost.python``` injection function. 
The ```boost.python.injector``` currently depends on the behaviour of ```__metaclass__```. This has changed in python 3, so boost injections using the class constructor will no longer happen. There are no error messages or anything - the code is simply never executed.

Injections in the form of
```python
class _(boost.python.injector, extension_class):
  def method(...):
     ...
```
will not work.

I have added a ```boost.python.inject``` function to achieve the same goal which works on python 2 and 3. The example above then changes to
```python
class _():
  def method(...):
    ...
boost.python.inject(extension_class, _)
```

The commit contains an example in ```scitbx/array_family/flex.py``` (which is required for the python 3 bootstrap build step).

@luc-j-bourhis since you did the most recent work on boost.python could you please take a look?

### DIFF
--- a/boost/python.py
+++ b/boost/python.py
@@ -222,6 +222,35 @@ class injector(object):
           setattr_from_dict(b.__dict__)
       return type.__init__(self, name, (), {})
 
+def inject(target_class, *mixin_classes):
+  '''Add entries from python class dictionaries to a boost extension class.
+
+     It is used as follows:
+
+           class _():
+             def method(...):
+               ...
+           boost.python.inject(extension_class, _)
+
+     instead of the previous mechanism of
+
+           class _(boost.python.injector, extension_class):
+             def method(...):
+               ...
+
+     which does not work in python 3.
+  '''
+  def setattr_from_dict(d):
+    for key, value in d.items():
+      if key not in ("__init__",
+                     "__del__",
+                     "__module__",
+                     "__file__",
+                     "__dict__") and (key != '__doc__' or value):
+        setattr(target_class, key, value)
+  for m in mixin_classes:
+    setattr_from_dict(m.__dict__)
+
 def process_docstring_options(env_var="BOOST_ADAPTBX_DOCSTRING_OPTIONS"):
   from_env = os.environ.get(env_var)
   if from_env is None: return None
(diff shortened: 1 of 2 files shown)


Anthchirp:
Just had another look at this. Alternatively we could use decorators:
```python
@boost.python.inject_into(extension_class)
class _():
  def method(...):
    ...
```
which are probably more readable because the injection declaration would be at the top of the class, right next to the class declaration.
Is this useful / more intuitive?

We can make the decorator distinguish between classes and functions, so instead of single method classes we could use
```python
@boost.python.inject_into(mersenne_twister)                                                                             
def random_selection(self, population_size, sample_size):                                                               
    (...)
```
which has the benefit of simpler declaration but the downside of affecting the local name space.

Code to define the decorator:
```python
import inspect                                                                                                          
def inject_into(target_class, *mixin_classes):                                                                          
  def _inject(c):                                                                                                       
    if inspect.isclass(c):                                                                                              
      inject(target_class, c, *mixin_classes)                                                                           
    else:                                                                                                               
      setattr(target_class, c.__name__, c)                                                                              
      class empty_class:                                                                                                
        pass                                                                                                            
      inject(target_class, empty_class, *mixin_classes)                                                                 
  return _inject                                                                                                        
```

Anthchirp:
@ndevenish pointed me to the boost::python documentation that contains the original injector pattern (https://www.boost.org/doc/libs/1_67_0/libs/python/doc/html/tutorial/tutorial/techniques.html about 2/3rds down). This advice hasn't changed in more than [14 years](https://github.com/boostorg/python/blame/d4d7f018f996b9c6f6056f1b94c41155a3214506/doc/tutorial/doc/tutorial.qbk#L1873), so it looks like there is currently no recommended Python 3 compatible injector pattern. I've opened a ticket upstream, https://github.com/boostorg/python/issues/210

luc-j-bourhis:
I would definitively favour the decorator approach.

In #10f4e8a, the function `setattr_from_dict` is not useful anymore: I recommend to put its body in the `for m in mixin_classes:` loop. It was necessary for the metaclass implementation only because of the quirks of the metaclass mechanism.

Anthchirp:
feel free to pick up the idea if desired

PR #199, opened 2018-05-24
Syntax changes for Python 3 bootstrap build step

Anthchirp:
This pull requests consists of 4 commits. The first one (f070afa) is only trivial changes, print rewrites etc, which hopefully are generally uninteresting.

The second commit 8256c9c rewrites a bit in ```scitbx/random/__int__.py```. @rjgildea you wrote most of that file, could you please have a look at that?

The third commit f5d6afe replaces ```reduce(operator.add, <iterator>)``` with ```sum(<iterator>)``` in ```scitbx/math/__init__.py``` which should of course produce identical results, but does not rely on ```reduce```, which was removed from the python 3 global namespace. The file was last touched by @pafonine, @gbunkoczi, @luc-j-bourhis - could one of you please have a look at that?

The fourth commit f3633cbee does minor fixes in ```scitbx/matrix/__init__.py```, but I noticed that file contains a very large block that is only test code. I therefore moved the test into a separate file. You can still run the test by calling either file directly. This file was last touched by @pafonine, @olegsobolev, @phyy-nx - could one of you please have a look?

(This pull request is required but not sufficient for Python 3 building, #197 and #198 are also required)

### DIFF
--- a/cctbx/array_family/flex.py
+++ b/cctbx/array_family/flex.py
@@ -1,4 +1,4 @@
-from __future__ import division
+from __future__ import absolute_import, division, print_function
 import scitbx.array_family.flex
 
 import boost.python
--- a/iotbx/phil.py
+++ b/iotbx/phil.py
@@ -1,11 +1,14 @@
-from __future__ import division
+from __future__ import absolute_import, division, print_function
+
+import os
+import sys
+
 from cctbx import sgtbx
 from cctbx import uctbx
 import libtbx.phil.command_line
 from libtbx.utils import Sorry, Usage, import_python_object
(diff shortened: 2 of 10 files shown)


graeme-winter:
Once @gbunkoczi changes have been looked at, @bkpoon please could you take a look as cctbx changes for Python 3 should go through Berkeley?

Anthchirp:
@gbunkoczi I found that sum() doesn't actually work in this context. In a3b2702 I have reverted sum() to reduce() and the generator to a list. The reduce function is imported from the python-internal functools library, which means it will work on both python 2 and 3.

PR #210, opened 2018-06-16
smtbx Python 3 changes

Anthchirp:
I prepared a few changes to smtbx. Feel free to pick none, some, or all of those. 
Together they give you a green tick mark on the [Python 3 syntax compatibility status](https://github.com/cctbx/python3/blob/master/README.md), which is the first step towards Python 3 compatibility.


luc-j-bourhis:
Why did you replace type comparisons with isinstance? Does Python 3 not support that anymore?

Anthchirp:
```type``` comparison is still supported as before, however ```isinstance``` appears to be the preferred method as it doesn't preclude subclassing, which ```type``` does. It also allows comparing against more than one type simultaneously.

In this case I did not make the change for any particular reason, so if you want to specifically use ```type``` I can revert those bits.

Anthchirp:
Reverted ```type```->```isinstance```. This conversion is not one of the default ones. I ran it accidentally as I wanted to split out ```print```->```print()``` from other changes, and missed that this one is not run by default.

PR #222, opened 2018-07-13
Reading data with psana2

monarin:
With these changes, the indexing and integration modules can now use psana2 interface to read out detector data (xtc2 format) and calibration files. The selection between the old psana and psana2 is done automatically through version checking.


asmit3:
Hi Mona,
This looks good. Internal tests I have related to xtc_process pass with psana-1. Merging now.

PR #241, opened 2018-10-18
Upgrade bootstrap/installer minimum python requirement to 2.6

Anthchirp:
A [recent change](https://github.com/cctbx/cctbx_project/commit/bb466f729ad1fb5bee9b89d5227b6d68421f51a5#r30946843) by @bkpoon added this requirement implicitly. This pull request will make that requirement explicit.

Any objections should be filed here.


phyy-nx:
I built on viper no problem.  Thanks.  Merging.

PR #242, opened 2018-10-18
Drop CentOS 5 support

Anthchirp:
Following on from #241: If CentOS 5 support really is dropped then we should consider removing the CentOS 5 specific code. These are the bits I could find.
(Patch is untested)


Anthchirp:
will leave at your discretion.

PR #244, opened 2018-10-21
Minor bits after Python2/3 dict and range changes

Anthchirp:
Hi @bkpoon, following your recent commits I found a few instances in the diff where I thought code could be optimized. These are my suggestions:

* No need to explicitly create lists when they are used
  in an iterator context, for example:
    ```for x in list(...)```
  or
    ```"".join([... for ...])```

* Remove a few unnecessary ```lange(len())``` instances

* Add list and dictionary comprehensions


Anthchirp:
feel free to cherry pick if so desired

PR #245, opened 2018-10-23
libtbx: Python 2/3 changes for python_code_parsing

bkpoon:
Hi Luc,

Here are some changes to libtbx/python_code_parsing.py for Python 3 compatibility. Basically, ```__nonzero__``` is replaced by ```__bool__``` in Python 3 and the test has a bunch of ```print``` statements.

The fixes are,
- use ```object``` from ```future.builtins``` that automatically maps ```__nonzero__``` to ```__bool___``` along with other Python 2/3 object mappings. Alternatively, we can just manually map the two if you do not plan on using any of the other standard object attributes.
- replace the ```print``` statements with ```type``` in the test

You can test this by configuring the ```libtbx``` module with Python 3 and running the test file. The test passes in both Python 2 and 3, so the fixes can be merged, but I wanted to double-check with you. Thanks!

### DIFF
--- a/libtbx/python_code_parsing.py
+++ b/libtbx/python_code_parsing.py
@@ -1,4 +1,7 @@
 from __future__ import division
+
+from builtins import object
+
 import sys
 assert sys.version_info[0:2] >= (2, 6)
 
@@ -29,7 +32,7 @@ class imported_name(object):
     return '.'.join(self.name)
 
 
-class unused_imports(ast.NodeVisitor):
+class unused_imports(ast.NodeVisitor, object):
   """ Unused import's in a module.
   This finds more of them than the algorithm in
   libtbx.find_unused_imports_crude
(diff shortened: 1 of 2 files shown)


Anthchirp:
There is no harm simply supporting both, eg.:
https://github.com/DiamondLightSource/ispyb-api/blob/6ba12723e54c1fd01f1c962d924631f540c6fea4/ispyb/model/__init__.py#L90-L96

luc-j-bourhis:
if `object` from `future.builtins` automatically maps `__nonzero__` to `__bool___`, then why did you need to explicitly rename __nonzero__ into __bool__? In any case, I don't have a Python 3 build of cctbx, so I can't check whether the test pass but I can trust you on that. I am actually amazed it passes as this code is AST wizardry, which I would have thought to break badly in the move from python 2 to 3!

bkpoon:
The renaming is so that the syntax is in Python 3. That way, when we eventually drop support for Python 2 in the future, we can just remove the import statement. Also, ```ast``` is in Python 3, so it seems to behave similarly to the Python 2 version.

PR #248, opened 2018-10-26
libtbx: remove sorting from libtbx.import_all_ext

bkpoon:
- libtbx.import_all_ext seems to run fine without sorting the list of extension modules
- Remove cmp_so function since cmp is not supported by Python 3

Are there cases where the sort order is still important?

The c8ca6424c2081d73a7433c32533352e19f126dbc commit from @phyy-nx seems to have fixed any dependency issues.

I do have a replacement key function for sorting ready, but if sorting is not necessary, it's probably better to just not sort.

### DIFF
--- a/libtbx/command_line/import_all_ext.py
+++ b/libtbx/command_line/import_all_ext.py
@@ -1,22 +1,4 @@
 from __future__ import absolute_import, division, print_function
-prios = ["boost", "scitbx", "cctbx", "rstbx", "labelit"]
-prios = dict(zip(prios, range(len(prios))))
-priority_modules = ['dials_algorithms_profile_model_modeller_ext']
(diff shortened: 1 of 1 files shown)


Anthchirp:
I am not certain what the point of the entire file exactly is. I assume it is for testing purposes only. If it is, then anything it loads should also be covered by more specific tests, which will produce a more useful output than indiscriminately importing everything one can get their hand on.
Additionally, as is said in c8ca642 we don't even import files that way in our anyway, so I'm even less sure what having this test adds to anything.

I would just remove the file.

phyy-nx:
According to @Oeffner, this test has caught problems in a few cases, for example on windows, so it's useful.

PR #249, opened 2018-10-31
str change refactor

Anthchirp:
It's the same test but without overwriting the Python 2 builtin.
This avoids some confusion and makes  clearer what is actually going on.


bkpoon:
I was waffling on the types and this removes a check, so it's fine.

PR #252, opened 2018-11-01
Refactor python3 syntax check to match libtbx programming patterns

phyy-nx:
Thanks @Anthchirp for the python3 syntax check.  I've refactored it to match the libtbx programming model, dropping references to pytest, naming the test like the other tests in the folder, and using libtbx.easy_run.  The last bit is because libtbx shouldn't have a procrunner dependency.

I've tested this to work on:
- Python 2 (conda), no python3.  Test is skipped.
- Python 3 (conda), no python 2.  Test passes and can be caused to fail if a new bad file is introduced.
- Python 2 (conda), with python 3 available.  Test passes and can be caused to fail if a new bad file is introduced.

### DIFF
--- a/libtbx/run_tests.py
+++ b/libtbx/run_tests.py
@@ -38,7 +38,7 @@ tst_list = (
   "$D/tst_representation.py",
   "$D/tst_runtime_utils.py",
 # "$D/tst_xmlrpc_utils.py", # This test is failing
-  "$D/test_python3_regression.py",
+  "$D/tst_python3_regression.py",
   )
 
 def run():
--- a/libtbx/test_utils/python3_regression.py
+++ b/libtbx/test_utils/python3_regression.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import os
-import procrunner
+from libtbx import easy_run
 
 def find_new_python3_incompatible_code(module_under_test):
   '''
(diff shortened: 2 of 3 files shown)


Anthchirp:
This code in libtbx is now used by dxtbx, dials, xia2, dials_research, dlstbx and all other repositories that we have that depend on this check, so I definitely do not want to change it without good reason.
I will put in a check that the test is skipped if procrunner is not present, which is fine because it really only needs to fail on a single system we do regular builds on. By default I assume there will not be a python 3 interpreter on developer systems so nothing is lost by this. The easy_run calls are not equivalent to the procrunner calls.
There is no standard as to how test files are named within libtbx, so unless you plan to enforce this across the board I don't see the point of that change.
The test does not use pytest and is in my opinion quite readable so again I don't see the need to refactor it. I can make it a conditional import though if you think that is useful.

graeme-winter:
@phyy-nx probably worth noting here that there was no dependency on pytest, just a facade to allow (one day, and perhaps elsewhere) pytest to be used to exercise this without substantial code changes. As @Anthchirp mentions above we also use the same code elsewhere (where it _is_ used with PyTest, i.e. dials, xia2 etc.) so it would make sense to keep it as is, so if we find a bug in one or make an improvement we can easily copy it everywhere else.

graeme-winter:
Re: procrunner vs. easy_run, I _can_ see that there is an argument here i.e. if the Python3 monster wanted to be unleashed somewhere else where procrunner is not used (buildbot, say) then it would be good if it worked. So making that possible would be good.

PR #256, opened 2018-11-09
Register flex arrays as Sequences and Iterables

Anthchirp:
The Python collections module has abstract base types for things that can be iterated and things that can be iterated and have a defined length (respectively called Iterables and Sequences).
Here we register the flex array base classes as Iterables and Sequences where appropriate so that flex arrays can now be used in places where the iterability is determined by the Python abstract base class register.

This has no effect on the actual behaviour of flex arrays.

### DIFF
--- a/scitbx/array_family/flex.py
+++ b/scitbx/array_family/flex.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import hashlib
+import six
 import sys
 
 import boost.optional # import dependency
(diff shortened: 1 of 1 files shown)


phyy-nx:
Cool!

graeme-winter:
All, any objections to this?

graeme-winter:
In call right now, @bkpoon says this sounds useful 👍

graeme-winter:
Given there are no objections and positive feedback, will merge this (is easily reverted if any problems arise)

PR #258, opened 2018-11-14
dxtbx.scan: add a valid_image_ranges map to the scan object

jbeilstenedmands:
The purpose of this addition is to allow different images ranges to be indicated for a sweep containing multiple lattices. This allows separated treatment of experiments which share a scan. It is intended that the map key can be the experiment identifier and the values are scitbx-int-arrays of
image ranges valid for each experiment (in the form [start1, end1, start2, end2, ...] for nonconsecutive ranges). The map is initialised as empty for all constructors.

For example, for a multi-lattice sweep with two lattices (with image range (1,100)), this map could be initially set to {"0", [1, 100], "1", [1, 100] }, then to indicate an excluded range (61 - 80) for experiment "0", it would be updated to {"0" : [1, 60, 81, 100], "1" : [1, 100] }. This can be set/managed by a few helper functions that will be added to DIALS.

Although this doesn't change a lot of code, I would welcome a quick review of my code to see if there are any obvious issues as I have less experience with c++.

### DIFF
--- a/dxtbx/model/boost_python/scan.cc
+++ b/dxtbx/model/boost_python/scan.cc
@@ -65,6 +65,8 @@ namespace dxtbx { namespace model { namespace boost_python {
     result["oscillation"] = rad_as_deg(obj.get_oscillation());
     result["exposure_time"] = boost::python::list(obj.get_exposure_times());
     result["epochs"] = boost::python::list(obj.get_epochs());
+    boost::python::dict valid_image_ranges = MaptoPythonDict(obj.get_valid_image_ranges());
+    result["valid_image_ranges"] = valid_image_ranges;
     return result;
   }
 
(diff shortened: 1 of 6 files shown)


Anthchirp:
The interface seems very C'y
```python
  scan = Scan((1, 100), (0.0, 1.0))
  scan.set_valid_image_ranges("0", (1,80))
  scan.set_valid_image_ranges("1", (1,50))
```

I suppose implementing the Python alternative is difficult?
```python
  scan = Scan((1, 100), (0.0, 1.0))
  scan.valid_image_ranges["0"] = (1,80)
  scan.valid_image_ranges["1"] = (1,50)
```

jbeilstenedmands:
Yes I don't know how to implement the python-style interface for the scan. However, I'm writing a few python functions that 'hide' this to some extent, such that you can do what you need to with just with an experiment list and the phil params.

PR #269, opened 2018-12-14
Implement Schrodinger force field 

latrocinia:
This commit should allow the use of the Schrodinger force field in the following applications.    
* geometry_minimization
* refine
* real_space_refinement
* ensemble_refinement

As with other external applications, it requires the setting of an environment variable, SCHRODINGER. This determines whether the Schrodinger software is installed, and options should pop up. 

The implementation allows to treat the whole structure with the force field, or partially, e.g. the ligand and its environment. In addition, PHENIX supported restraints can be 'mixed' in, such as NCS restraints (by default). 

As with other force field implementations there are some limitations. 

* The input structure needs to be prepped with the Schrodinger prepper, so that it's consistent with the OPLS3e force field implementation. 
* Adding water and ions during refinement is not (yet) supported.

The options required for running with force field are two-fold:
use_schrodinger=True maestro_file=<prepped structure in maestro format>


olegsobolev:
Fixed ensemble refine, to eliminate the need to change it in this work. Will look more.

latrocinia:
> I assume one needs Schrodinger to run it effectively but does fail nicely without?

The test won't run unless Schrodinger is installed, at least if run from the command line. The tests should probably be decorated using a skip function from the unittest module to be more robust. 

> The added white space (line 372) in mmtbx/refinement/real_space/individual_sites.py is noise and best avoided.

That one slipped in. 

>Fixed ensemble refine, to eliminate the need to change it in this work. Will look more.

Great, thanks!

>A major item. It would be better for a number of reasons to have sch..ger.py and tst_sch...ger.py in their own GitHub repo. The work flow for the user would clone the repo into modules and configure the new repo into Phenix. It's been done in other situations so we have a model for it. We can even have a script in Phenix that installs the pieces for a Sch...ger user.

How are changes in e.g. `model.py` handled in that case? Is the file patched, or the original file is replaced by a full copy plus additions? 

>BTW, there are a couple of minor changes to the cctbx part but lets get the major changes first.

Sure thing.

olegsobolev:
> How are changes in e.g. model.py handled in that case? Is the file patched, or the original file is replaced by a full copy plus additions?

I guess this will look exactly like it is, but will import your class from another place. Essentially, this and mmtbx/geometry_restraints/external.py should be the only places that need to be changed.

BTW, do you have any idea how to avoid modification of individual_sites.py ? It is another big concern.

latrocinia:
> I guess this will look exactly like it is, but will import your class from another place. Essentially, this and mmtbx/geometry_restraints/external.py should be the only places that need to be changed.

Ok great.

> BTW, do you have any idea how to avoid modification of individual_sites.py ? It is another big concern.

The issue here lies with extract_box_around_model_and_map class. It selects part of the structure and boxes out a region in the map around the selected part. However, it moves the map such that the origin of the map is at (0, 0, 0). In turn the selected structure part is moved. This is also the reason why the reference_coordinate_restraints need to be temporarily removed. For the cctbx geometry restraint manager this is fine, since restraints that involve any of the non-selected atoms are removed, and it considers the selected part in pure isolation, without regards to any bonds it has with the initial structure.
For the force field I'm not able to do that, as the selected atoms are still aware of their environment and bonds with non-selected atoms, their gradients and energies are simple not calculated. Handling the selected substructure in isolation would result in an imbalanced force field, and our software would complain about valency etc. So  the non-selected atoms need to have the same translation as the selected part to keep the integrity of the structure.

To avoid modification and keep the algorithm, the boxing of the map could be made more flexible, i.e. the boxed map should be able to have a float origin, such as is allowed in the .mrc format for EM maps, so that the selected part does not need to be translated. This would also alleviate the requirement of temporarily removing the reference_coordinate_restraints.

latrocinia:
I'll close this pull request and submit a new one including only changes in the existing cctbx_project files.

PR #270, opened 2018-12-18
Implement Schrodinger hooks for refinement programs.

latrocinia:
I left out the `schrodinger.py` and `tst_schrodinger.py` files. 

I updated the ensemble refinement parameter reading, cause, to my understanding, only the default values would be created, and by passing the er_params the schrodinger commandline options would be transferred to the pdb_ip instance.

Also 2 extra lines were added to the individual_sites.py file to shift the structure back to its original position.

### DIFF
--- a/mmtbx/geometry_restraints/external.py
+++ b/mmtbx/geometry_restraints/external.py
@@ -69,17 +69,6 @@ if schrodinger_installed:
       .help = Parameters for using Schrodinger's force fields.
       .expert_level = 3
     {
-      use_schrodinger = False
-        .help = Use Schrodinger supported force field.
-        .type = bool
-      ligand_selection = None
-        .help = Use force field parameters only for selected ligands.
-        .type = atom_selection
-      forcefield = OPLS3e* None
-        .help = Force field to use during refinement.
-        .type = choice
-      maestro_file = None
-        .help = Schrodinger Maestro file containing structure.
-        .type = path
+      include scope mmtbx.geometry_restraints.schrodinger.master_phil_str
     }
 """
--- a/mmtbx/geometry_restraints/external.py
+++ b/mmtbx/geometry_restraints/external.py
@@ -59,16 +59,26 @@ if (afitt_installed) :
 
 # Schrodinger
 schrodinger_installed = False
-if os.environ.get("SCHRODINGER", None):
+if os.environ.get("SCHRODINGER", False):
   if os.path.exists(os.environ["SCHRODINGER"]):
     schrodinger_installed = True
 
 if schrodinger_installed:
-  external_energy_params_str += """
+  from glob import glob
+  paths = glob(os.path.join(
+      os.environ["SCHRODINGER"], 'psp-v*', 'data', 'phenix'))
+  if len(paths) == 1:
+    import sys
+    sys.path.append(paths[0])
+  try:
+    import phenix_schrodinger
+    external_energy_params_str += """
     schrodinger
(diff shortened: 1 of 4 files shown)


olegsobolev:
> I updated the ensemble refinement parameter reading, cause, to my understanding, only the default values would be created, and by passing the er_params the schrodinger commandline options would be transferred to the pdb_ip instance.

No. get_default_pdb_interpretation_params() does not takes any arguments so it would probably just fail. Moreover, it uses external_energy_params_str from mmtbx.geometry_restraints.external which should contain all available external parameters. So ensemble refinement changes do not make sense for me.

I'll look at individual_sites later while you coordinating with @nwmoriarty  particular places/ways of installing Sch...er implementation which is needed for correct import statements.

latrocinia:
> No. get_default_pdb_interpretation_params() does not takes any arguments so it would probably just fail. Moreover, it uses external_energy_params_str from mmtbx.geometry_restraints.external which should contain all available external parameters. So ensemble refinement changes do not make sense for me.

Sorry, was confused with `set_pdb_interpretation_params`. Thanks!

latrocinia:
We were planning on putting the module and test files directly into the Schrodinger software package, this way we can hook up our automated testing machinery to the Phenix/Schrodinger tests. This has the added benefit that the user does not have to install anything extra. Essentially, the PYTHONPATH is adjusted dynamically if Schrodinger is installed and the module and objects are imported. If you guys are onboard with this, I'll update the code once more, and do another pull request. Let me know.

olegsobolev:
I'd recommend committing changes to this pull-request instead of creating another one. This keeps all comments.
> Add more commits by pushing to the schrodinger branch on latrocinia/cctbx_project

latrocinia:
Fair enough, will do.

nwmoriarty:
Are you aware that the Phenix dispatchers generally clean the PYTHONPATH for running Phenix programs? Have you confirmed that it works from both Phenix and Sch..ger, if desired?

latrocinia:
I've updated the branch to show how it would look like when the `schrodinger.py` module is stored in the SCHRODINGER distribution.

> Are you aware that the Phenix dispatchers generally clean the PYTHONPATH for running Phenix programs? Have you confirmed that it works from both Phenix and Sch..ger, if desired?

This should work as long as the SCHRODINGER environment variable is properly set. If not, it would seem like SCHRODINGER is not installed and vanilla phenix just runs its course.

nwmoriarty:
OK, as long as we are aware.

Now we can turn to testing. I have a number of BuildBots that test the external packages. It's a simple matter to add this package but I'll need an working installation or a dummy package. Any thoughts? I'm happy to leave all testing to you if you desire.

nwmoriarty:
You can use the set_source and get_source methods of base_geometry.py to propagate the source of the geometry restraints manager so you don't have to do such things as hasattr in individual_site.py.

latrocinia:
I think it's best if the testing is initially done on our side. When we are out of the alpha / beta stage, it's probably worthwhile to also put some tests for the BuildBots. 

> You can use the set_source and get_source methods of base_geometry.py to propagate the source of the geometry restraints manager so you don't have to do such things as hasattr in individual_site.py.

This has been updated.

nwmoriarty:
All tests pass in t96. Let's see about the overnights.

latrocinia:
Great! 

On a slightly different notice, I'll update you coming month with scientific results on the impact of the force field on reciprocal refinement and hopefully some real_space_refinement.

PR #283, opened 2019-01-18
Changes to support DLS Eiger2 nexus format

rjgildea:
This pull requests includes some changes necessary to support the Eiger2 detector now installed on DLS beamline I04.
- added ability to construct multi-axis goniometer from a nexus file
- bug fix for conversion of units for offset
- add format class for DLS I04 Eiger2 including a smargon shadowing model
- add tests for DLS I04 Eiger2, VMXi Eiger2, Soleil Eiger, SPring8 Eiger
- add format class for SPring8 Eiger handling inverted rotation axis


graeme-winter:
* McStas to CIF frame conversion should be made to be atomic somewhere i.e. function rather than coding everywhere
* Does Eiger NeXus format class work for master and nxs files?
* XXX mask really shouldn’t be none - would be good to have a ticket reference here
* pep8 says “"Don't understand this goniometer: %s" %list(goniometer.get_names()))” -> space either side of %
* assert panel.get_trusted_range() == (-1,4096) # XXX really?! yes really - was at the time of collections
* multiple tests for different facilities good

Issue #151, opened 2018-03-13
Move libtbx.pkg_utils.require out of libtbx.configure/refresh


phyy-nx:
# Summary

libtbx.pkg_utils.require is currently being used to install via pip the packages mock, pytest, orderedset, Jinja2 and procrunner. This is done by the libtbx_refresh.py scripts in cctbx_project, dials and xia2.  In bootstrap, this is triggered in the build step as part of configuration before compilation.  It needs to instead be done in install_base_packages as part of the base step.

# Background
This issue comes from supercomputing environments where cctbx based software must be compiled on a node without internet access or without write privileges to the base folder.  

### Example use case
The user logs into a public node on a supercomputing environment and uses bootstrap.py to run hot and update.  The system provides a custom python with a cluster-dependent MPI package, optimized for their environment.  From the public node, the user logs into a compute node that is internet firewalled and has a different set of compilers as the public node.  The user runs 

```python bootstap.py build --with-python=`which python` ```

but it hangs because a) the user can't download mock, pytest etc. and b), even if they could, they couldn't install them because they lack write access to the system python.

### Conda
This issue is further complicated by conda installations of cctbx-based packages.  XFEL builds at LCLS have moved to conda exclusively.  Quoting the [conda docs](https://conda.io/docs/user-guide/tasks/manage-pkgs.html):

> Pip packages do not have all the features of conda packages, and we recommend first trying to install any package with conda. If the package is unavailable through conda, try installing it with pip. The differences between pip and conda packages cause certain unavoidable limits in compatibility, but conda works hard to be as compatible with pip as possible.

# Proposed change

New set of changes based on discussion 03/16/18

Specifically:

- [ ] Add two checks in libtbx.pkg_utils.require before running pip.  If the install is 'baseless', meaning there is no base folder and a different python is being used, do not try to install the specified package.  Likewise if the environment variable LIBTBX_NO_NETWORK is set, skip installation.  In both cases, print a warning that the package is missing and any code depending on it will not run. Owner: @Anthchirp
- [ ] Change scipy builder in install_base_packages.py to use pip instead of ./configure and make.  This works around fortran compiler issues.  Owner: @phyy-nx
- [ ] Add mock, pytest, orderedset, Jinja2, procrunner and scipy to the appropriate builders in install_base_packages. Owner: @phyy-nx

Feedback on this issue is appreciated :)

graeme-winter:
OK, some thoughts - 
[...]

PR #290, opened 2019-02-15
Define python dependencies in libtbx_config

ndevenish:
At the moment python dependency resolution is done and installed at the `libtbx.refresh` stage, via each packages `libtbx_refresh.py`. This can cause problems if packages cannot be installed e.g. on an offline node, or if the global site-packages folder is read-only. 

Having the configuration inside the `libtbx_refresh.py` file means that the only way to tell what's required is by manual inspection or parsing the output of refresh. And, since each package only knows about it's own requirements, leaves the potential of conflicting requirements going unnoticed.

This patch moves the listing of requirements to the libtbx_config configuration file, where the other module dependencies are specified, which means that it can be easily read before running libtbx.refresh - dependencies can be worked out before you have a fully running and configured distribution, and merged before installation.

This would help some of the issues discussed in #151.

Also, added a utility, `libtbx.show_python_dependencies`, to list/inspect python dependencies in an existing environment. Here's what the standard output looks like where I've converted the `dxtbx`, `dials` (sample commit https://github.com/ndevenish/dials-fork/commit/6e9f736c0d90bcd2f637f28f97398e87b8f4df21) and `xia2` refresh scripts:

image

By passing the `--all` flag you can get a list of all requirements suitable for e.g. building an offline list:
```
$ libtbx.show_python_dependencies --all
blosc Jinja2 'mock>=2.0' msgpack orderedset 'procrunner>=0.6' 'pytest>=3.1,>=3.6' scikit_learn[alldeps] scipy tabulate 'tqdm==4.23.4'
```

### DIFF
--- a/dxtbx/libtbx_config
+++ b/dxtbx/libtbx_config
@@ -1,3 +1,7 @@
 {
   "modules_required_for_use" : [ "iotbx", "cbflib_adaptbx" ],
+  "python_required": [
+    "mock>=2.0",
+    "pytest>=3.1"
+  ],
 }
--- a/dxtbx/libtbx_refresh.py
+++ b/dxtbx/libtbx_refresh.py
@@ -1,6 +1,1 @@
 from __future__ import absolute_import, division, print_function
-
-import libtbx.pkg_utils
-
-libtbx.pkg_utils.require("mock", ">=2.0")
-libtbx.pkg_utils.require("pytest", ">=3.1")
--- a/libtbx/command_line/show_python_dependencies.py
+++ b/libtbx/command_line/show_python_dependencies.py
@@ -214,7 +214,7 @@ if __name__ == "__main__":
             to_action.append(requirement)
 
     # Decide how to do output
-    if normal_output:
+    if normal_output and rows:
         _print_table(rows)
 
         if to_action:
@@ -227,6 +227,8 @@ if __name__ == "__main__":
                     + [_shell_quote_requirement(req) for req in to_action]
                 )
             )
+    elif normal_output and not rows:
+        print("No dependencies found.")
     elif args.all:
         print(" ".join(_shell_quote_requirement(x) for x in requirements))
     elif args.actions:
--- a/libtbx/command_line/show_python_dependencies.py
+++ b/libtbx/command_line/show_python_dependencies.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python
+
+"""
+Lists the module python dependencies and the state of install.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import argparse
+import copy
+import re
+import sys
+
+import libtbx.load_env
+import libtbx.pkg_utils
+from libtbx.pkg_utils import pkg_resources
+from libtbx.pkg_utils import packaging
+
+
+BOLD = "\033[1m"
+RED = "\033[1;31m"
+MAGENTA = "\033[1;35m"
+GREEN = "\033[32m"
+GRAY = "\033[37m"
+NC = "\033[0m"
+
+# A regex to check for requirement characters that are not shell-safe
+reSafe = re.compile(r"[^a-zA-Z0-9,._+:@%/\[\]-]")
+
+
+def _get_requirement_status(requirement):
+    # type: (packaging.requirements.Requirement) -> Tuple[Any, str]
+    """Get the requirement status of a package.
+
+    Args:
+        requirement (packaging.requirements.Requirement):
+            The requirement object to check
+
+    Returns:
+        Tuple[str, str]:
+            The status value, and - if it exists - the current version.
+            The status value can currently be one of:
+                - "SKIPPED": The package is excluded by env markers
+                - "VALID":   Already installed with correct version
+                - "UNKNOWNEXTRA": Installed, but missing extras
+                - "MISSING": The package is not installed
+                - "MISMATCH": An incompatible version is installed
+    """
+    requirement = copy.deepcopy(requirement)
+    if requirement.marker and not requirement.marker.evaluate():
+        # Try to get a version without the marker
+        requirement.marker = None
+        # Look to see if we have an installed version anyway
+        try:
+            version = pkg_resources.require(str(requirement))[0].version
+            return ("SKIPPED", version)
+        except pkg_resources.ResolutionError:
+            return ("SKIPPED", None)
+
+    # Try to find this package
+    try:
+        pkg = pkg_resources.require(str(requirement))[0]
+        return ("VALID", pkg.version)
+    except pkg_resources.UnknownExtra:
+        # Package exists, but extra is missing. Get the base description.
+        requirement.extras = set()
+        version = pkg_resources.require(str(requirement))[0].version
+        return ("UNKNOWNEXTRA", version)
+    except pkg_resources.VersionConflict:
+        requirement.specifier = packaging.specifiers.SpecifierSet()
+        version = pkg_resources.require(str(requirement))[0].version
+        return ("MISMATCH", version)
+    except pkg_resources.DistributionNotFound:
+        # Package not installed
+        return ("MISSING", None)
+
+
+def _shell_quote_requirement(req):
+    # type: (Union[packaging.requirements.Requirement, str]) -> str
+    """Returns a shell-quoted requirement string.
+
+    Since we know the form of the requirement, it is safe to single-quote
+    anything and convert internal single-quotes to double (we know no
+    recursive quoting). But if it's not necessary to do this, it won't.
+
+    Args:
+        req (Requirement or str):
+            The requirement, as an object or PEP508 string.
+
+    Returns:
+        str:
+            A shell-safe, potentially escaped version of the string.
+    """
+    out = str(req)
+    if reSafe.search(out):
+        # Something needs to be quoted. Quote it, but we know the
+        # general form and know we don't have recursively nested quotes
+        out = "'" + out.replace("'", '"') + "'"
+    return out
+
+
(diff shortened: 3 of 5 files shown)


graeme-winter:
This looks like a great idea - I can also see it being very helpful for @bkpoon making conda package updates - having all the dependencies explicit would really help.

graeme-winter:
Will need to merge across the dependency requirement specification for other cctbx packages thoughj e.g. 

```
{
  "modules_required_for_use" : [ "iotbx", "cbflib_adaptbx" ],
  "python_required": [
    "mock>=2.0",
    "pytest>=3.1"
  ],
}
```

for dxtbx - looking at it though these are pretty compact - in fact do other packages even have these?

graeme-winter:
... that would be "no"

```
Grey-Area cctbx_project :( [master] $ grep 'libtbx.pkg_utils.require' */libtbx_refresh.py
dxtbx/libtbx_refresh.py:libtbx.pkg_utils.require('mock', '>=2.0')
dxtbx/libtbx_refresh.py:libtbx.pkg_utils.require('pytest', '>=3.1')
```

bkpoon:
This would be a good opportunity to discuss the design and goals of this functionality. It looks like this was originally added with ea7c2c0513f0787f6362dd45a182d908a27e5610 in September 2017 and there was not much discussion about its purpose. Since this affects everyone and has caused issues in the past as stated by @ndevenish (#151) and also on the [cctbxbb](http://phenix-online.org/pipermail/cctbxbb/2018-April/001869.html), I think we should try to prevent future problems by laying out the problem to be solved and the overall plan.

Here is a list of concerns that I have with this feature.
- My main concern is that this seems to only work for Python dependencies and possibly only pure Python dependencies. We have both Python and non-Python dependencies and some are related (e.g. `libhdf5` and `h5py`, or `mpi4py` and MPI libraries). I think all dependencies should all be handled in a consistent manner. We currently have all the dependencies defined in [`package_defs.py`](../blob/master/libtbx/auto_build/package_defs.py). The file is hand-curated and built regularly to ensure that all the dependencies are compatible. For `conda`, I have combined all the dependencies into a [single file](
https://github.com/cctbx/conda_build/blob/master/cctbx_dependencies/meta.yaml) as well, with `conda` handling the dependency resolution.
- Does this cause each directory to have it's own list of dependencies? And only for packages installable by `pip`? Where would non-Python dependencies go?
- What about duplicates? In `dxtbx` and `dials`, there are duplicated entries for `mock` and `pytest`. Since `dials` depends on `dxtbx`, should those dependencies be defined just for `dxtbx`?
- There seems to be [`libtbx.install`](../blob/master/libtbx/command_line/install.py) command that looks like a wrapper for `pip` but is also used to handle dependencies? I see the header files for `msgpack` but isn't that a duplicated entry in bootstrap.py?
- Will `pip` be used for dependency resolution? It looks like `pip install` would need to be run with the `-r` flag as noted in the [documentation](https://pip.pypa.io/en/stable/user_guide/#requirements-files).

graeme-winter:
@bkpoon the idea of this is to declare explicitly requirements, which seems like _a good idea_ 

One challenge we face is adding modules where we have local dependencies not satisfied by https://github.com/cctbx/cctbx_project/blob/master/libtbx/auto_build/package_defs.py - i.e. dependencies in local modules which are not distributed. This was the basis of the 2017 work. 

Each directory _may_ have dependencies (i.e. not must) but if not, they should probably be declared in https://github.com/cctbx/cctbx_project/blob/master/libtbx/auto_build/package_defs.py 

Non Python dependencies are a different problem but presumably would have to go through the existing mechanisms - we are not trying to solve all problems here just make one thing better. Duplicates are fine, since multiple modules may share dependencies and they will all be satisfied anyway. Pip can be used for dependencies, but if they are all already there then nothing will happen, so no harm will be done.

Other than "this does not solve all the problems" I do not seem what _harm_ is done by this? Does it make anything actually worse.

graeme-winter:
Concern reported by @phyy-nx / @bkpoon that this is turning into a package manager - with good reason? Discuss.

graeme-winter:
TBH I think having explicit per-module dependencies is a good thing, since it avoids one static dependencies-for-all-things file somewhere. But that's my perspective. Internal Diamond examples helpful here.

graeme-winter:
Example from diamond...
```
from __future__ import absolute_import, division, print_function

import ast
import imp
import pkgutil

import dlstbx.services
import libtbx.pkg_utils

try:
  from dlstbx.util.version import dlstbx_version
  print(dlstbx_version())
except Exception:
  pass

libtbx.pkg_utils.require('mock', '>=2.0')
libtbx.pkg_utils.require('pytest', '>=3.1')
libtbx.pkg_utils.require('ispyb', '>=4.13,<4.14')
libtbx.pkg_utils.require('workflows', '>=1.1')
libtbx.pkg_utils.require('drmaa')
libtbx.pkg_utils.require('junit_xml')
libtbx.pkg_utils.require('colorama') # is still used in one place
libtbx.pkg_utils.require('procrunner', '>=0.8.0')
libtbx.pkg_utils.require('zocalo', '>=0.5.1')

# Eiger stream work, SCI-7786
libtbx.pkg_utils.require('confluent-kafka')
libtbx.pkg_utils.require('msgpack')
libtbx.pkg_utils.require('pyzmq')
```

graeme-winter:
Many things in here which are not in standard bootstrap build, and also don’t belong in there either...

ndevenish:
libtbx is already a package manager, for tbx modules. It allows adding arbitrary modules and manages dependencies between them, except that it's not a very good or complete package manager - some class of dependencies (binary, python) are impossible to determine before running the code, and up until the changes in ea7c2c0513f0787f6362dd45a182d908a27e5610, impossible to specify except centrally in the hard-to-edit `install_base_packages.py`. It has lots of other issues that we jump through hoops to work around but those are a separate topic.

Centrally specifying packages doesn't work for us for e.g. diamond-internal packages, with dependencies that don't belong in the central installation. Plus, central lists make it hard to connect requirements with where they came from and what they are for (the dials dependencies, for a start, are for a completely different repository).

The previous `pkg_utils.require` changes let us manage our installations and add/dependencies without having to edit the central dependency lists, but still the problem that it's impossible to tell what the dependencies are without actually configuring and running the packages, and It also made it slightly harder for you to tell what the requirements were (by definition, allowing change at all makes this harder) which you ran into with your offline source-based installations.

All this does is make dependencies easier to write, and easier to introspect. It's the opposite of making libtbx into a package manager - because it deliberately makes it easier to rely on external tools for determining environment.

As pointed out above, this doesn't handle other binary dependencies. Adding binary dependencies is a much bigger deal with more consequences and more difficult cost/benefit decisions.

bkpoon:
I'm getting this traceback when trying to configure `libtbx` by itself.
```
  File "../modules/cctbx_project/libtbx/configure.py", line 34, in <module>
    if not run():
  File "../modules/cctbx_project/libtbx/configure.py", line 28, in run
    import libtbx.env_config
  File "/dev/shm/bkpoon/software/a/modules/cctbx_project/libtbx/env_config.py", line 21, in <module>
    import libtbx.pkg_utils
  File "/dev/shm/bkpoon/software/a/modules/cctbx_project/libtbx/pkg_utils.py", line 12, in <module>
    import libtbx.load_env
  File "/dev/shm/bkpoon/software/a/modules/cctbx_project/libtbx/load_env.py", line 5, in <module>
    libtbx.env = libtbx.env_config.unpickle()
AttributeError: 'module' object has no attribute 'env_config'
```
I have a `modules` directory with just `cctbx_project`. Then I run
1) `git checkout -b ndevenish-python_config master` (GitHub CLI instructions)
2) `git pull https://github.com/ndevenish/cctbx_project-fork.git python_config` (GitHub CLI instructions)
3) Make a `build` directory and copy an existing `base` directory to the same level as the `modules` directory.
4) In the `build` directory, run `../base/bin/python ../modules/cctbx_project/libtbx/configure.py libtbx`

I am not fully aware of the history of CCTBX, but the configuration step in `libtbx` is not designed to be a general package manager. It is based on SCons (https://scons.org), which is like `make`. And like `make`, usually only the dependencies among the source code, in our case, the `*tbx` modules, are described. External dependencies were not included, presumably since there were not many when CCTBX first started.

Looking at the implementation of `libtbx.pkg_utils.require` and the `resolve_python_module_dependencies` function in this pull request, are you checking each line separately? For example, in
```
libtbx.pkg_utils.require('mock', '>=2.0')
libtbx.pkg_utils.require('pytest', '>=3.1')
```
is `mock` checked first, and then `pytest` is checked separately? This pull request mentions "python dependency resolution," but what is the intended behavior? Checking dependencies one at at time with `pip` is not the same as checking them all at once as described in the `pip` documentation [link](https://pip.pypa.io/en/stable/user_guide/#requirements-files)

Would it make sense to construct a `requirements.txt` file in your Diamond-specific packages and run `pip install -r requirements.txt`?

ndevenish:
> I'm getting this traceback when trying to configure `libtbx` by itself.

Thanks, there was a simple import-loading-order problem that I've fixed.

Regarding this commit:

- This doesn't add any stages of install that aren't already done
- This actually does do some dependency merging whereas conflicts just competed before
- This is literally just making the metadata easier to manage
- We discussed this nine months ago and response was positive https://docs.google.com/document/d/1oRXECbX5znnsoL3nLaDr0jqXP-Yh6mFCAx04ftvYr6c/edit#bookmark=id.w47xxbi2zo47

graeme-winter:
From where I am sat, it looks like the issues here are resolved and - from what I can tell - the changes should be invisible for the vast majority of cctbx developers.

Issue #288, opened 2019-02-12
Using a stable sort as default in flex.sort_permutation, flex.sort?


jbeilstenedmands:
Recently I have been using some flex sorting code in DIALS, and after discovering some cross-platform variations in results I was surprised to find that the default is to use an unstable sorting algorithm. In the context of sorting, unstable means that any equal values in a list will not in general be in the same order as in the input list, and the order may be different across platforms based on the c++ compiler (however the sorting is still deterministic and will give the same value when repeated for given compiled code).

An unstable sort is the default for std::sort in c++: for scitbx it looks like the stable option was added in commit 7971e48 and then used for miller sets and arrays in commit d729566 , however the default was kept as False.

The benefit of an unstable sort is a potentially faster runtime, but given the nature of scitbx being a library, I was wondering if the default should be set as a stable sort, and in specific instances where it is safe and useful to do so then the stable=False option can be used? The flex.sort and flex.sort_permutation methods are used extensively throughout the codebase, so it is not clear where this could cause an issue.

In pure python, the standard sorting methods are stable, however numpy uses an unstable quicksort by default. If users are unaware that the cctbx default is unstable, this could lead to subtle and nasty bugs.

phyy-nx:
Total agreement here.  Any identification of platform-specific degeneracy
that can be eliminated is worth eliminating.

graeme-winter:
Practical suggestion:
 - run all tests, record timing (serial not parallel)
 - make change
 - run all tests again, record timing
 - if any times changed significantly, look at setting these to explicitly use unstable version
Opinions?

PR #293, opened 2019-02-28
Use stable sort by default

Anthchirp:
following #288.

Performance impact appears to be, if there is any at all, on the positive side.

### DIFF
--- a/scitbx/array_family/flex.py
+++ b/scitbx/array_family/flex.py
@@ -49,7 +49,7 @@ class _(boost.python.injector, grid):
     print("all:", self.all(), file=f)
     return self
 
-def sorted(data, reverse=False, stable=False):
+def sorted(data, reverse=False, stable=True):
   return data.select(
     sort_permutation(data=data, reverse=reverse, stable=stable))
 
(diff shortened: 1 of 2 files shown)


Anthchirp:
image

phyy-nx:
Are these numbers test timings?

Anthchirp:
Ah yes, units are seconds as reported by bash _time_, running basically
```bash
for m in libtbx cctbx iotbx scitbx rstbx dxtbx; do
  for round in 1 2 3 4 5; do
    mkt # run in a new temporary directory
    time libtbx.run_tests_parallel module=${m} nproc=1
  done
done
```

phyy-nx:
Hi all, I ran the phenix tests and one of them doesn't pass on the stablesort branch, but does pass on master.  Specifically phenix_regression/wizards/command_line_misc_tests/test_hyss/tst_hyss.py.

I'll be in touch with our hyss experts about what is going on, but I don't think we can merge just yet.  Thanks,
-Aaron

graeme-winter:
@phyy-nx interesting - the stable sort main difference is predictable behaviour - unstable sorting obviously does not have this - so very surprising that hyss somehow depends on that 🤔

I suspect that this is a case where the change unmasks another issue, rather than the change itself causing the problem (or the test is not optimally written?)

phyy-nx:
@terwill has fixed the test.  I think we are ok to merge.

rjgildea:
The following commit should also have been included in this pull request:
https://github.com/cctbx/cctbx_project/commit/4354713150bd0c8aa9e1a375bc79e0fa7c7d75c1

PR #295, opened 2019-03-05
Olex2

pcxod:
This is combined pull request for changes previously requested to be merged by Pascal and the other bits implemented by Luc and myself… It is very large but should not affect much outside of smtbx, hopefully!

The main changes are:
-	SVD re-implemented on the C++ side to match the one on Python side. This will be required for some TLS procedure in the future development.
-	RIGU/CHIV restraints as from previous requests by Pascal – the previous one should be ignored. RIGU has been reworked and basic tests added. The debug output stripped from CHIV.
-	FAST_LINALG module which supplies C++ side with some of the BLAS routines. The modules is optional but also is dynamic – it does not require dependencies at the compile time and does not interfere unless is explicitly imported or used. When imported it will try to load functions from openblas.dll on windows or libopenblas.so on other platforms that it finds in the system PATH. If the one is not found on Windows – it will rummage through the SciPy libraries trying to find openblas. It also can be initialised with other libraries that have the same function signatures. The module can be extended with any other BLAS functions in copy/paste mode…
-	The normal equation and design matrix building now can exploit multiple threads

This is an example output for test of LS procedure on cctbx configured without and with fast_linalg (when fast_linals successfully initialises, it prints the BLAS library configuration string):

libtbx.python ..\modules\cctbx_project\smtbx\refinement\tests\tst_least_squares.py
Serial computation of all Fc(h) and their derivatives
        Normal matrix accumulation with BLAS level 2
Parallel computation of all Fc(h) and their derivatives
        Normal matrix accumulation with BLAS level 2
usr+sys time: 12.05 seconds
wall clock time: 8.92 seconds

libtbx.python ..\modules\cctbx_project\smtbx\refinement\tests\tst_least_squares.py
Successfully initialised SciPy OpenBlas:
DYNAMIC_ARCH NO_AFFINITY Sandybridge
Serial computation of all Fc(h) and their derivatives
        Normal matrix accumulation with BLAS level 2
        Normal matrix accumulation with BLAS level 3
Parallel computation of all Fc(h) and their derivatives
        Normal matrix accumulation with BLAS level 2
        Normal matrix accumulation with BLAS level 3
usr+sys time: 26.33 seconds
wall clock time: 19.72 seconds


olegsobolev:
I'm fine with cctbx changes (style-wise). I don't think they will immediately affect any cctbx/Phenix performance.
New test in 
cctbx_project/iotbx/cif/tests/tst_restraints.py is failing for me currently:
-  O1  C1  0.004000  -0.002796  -0.000917   0.000599
-  O1  C2  0.004000   0.000894  -0.001962  -0.005274
-  O1  C3  0.004000  -0.002962   0.000664   0.001305
+  O1  C1  0.004000  -0.002618  -0.001550   0.000599
+  O1  C2  0.004000  -0.000752   0.002098  -0.005274
+  O1  C3  0.004000  -0.002608  -0.001448   0.001305
After this is fixed and @bkpoon verifies that this code can be built on all supported platforms this can be merged.

pcxod:
@olegsobolev Sorry missed that test - in the updated code the Z axis direction has changed that is why only ZZ component wast matching, also there has been a bug when testing the vector length using abs functions in the original code that generated that output.

olegsobolev:
Here's a list of failing tests after the last commit:

cctbx_project/dxtbx/tests/model/test_crystal_model.py::test_change_basis_mosaic_crystal
cctbx_project/cctbx/adp_restraints/tst_ext.py
cctbx_project/iotbx/cif/tests/tst_restraints.py
libtbx.python -m pytest -rsxX --basetemp=pytest/t019 "cctbx_project/smtbx/refinement/restraints/tests/test_manager.py::test_manager" [FAIL] 5.3s

Apparently, it didn't work as expected...

pcxod:
OK, I have committed fixed tests adp_restraints (last 3 test in the list) but looking at the first one - is it supposed to fail?
This  is the comment from the test file at line 409:
    # Currently change_basis downgrades the MosaicCrystalSauter2014 to
    # a Crystal object so this will fail

phyy-nx:
That test is marked as expected to fail using pytest:

@pytest.mark.xfail
def test_change_basis_mosaic_crystal():

olegsobolev:
I'm still getting this test failing:
iotbx/cif/tests/tst_restraints.py
The diff is:
```
-  O1  C1  0.004000  0.002618  -0.001550   0.000599
-  O1  C2  0.004000  0.000752   0.002098  -0.005274
-  O1  C3  0.004000  0.002608  -0.001448   0.001305
+  O1  C1  0.004000  -0.002618  -0.001550   0.000599
+  O1  C2  0.004000  -0.000752   0.002098  -0.005274
+  O1  C3  0.004000  -0.002608  -0.001448   0.001305
```
Are the numbers in fourth column are supposed to be positive or negative? We are getting negative on CentOS 6 with gcc 4.4.7.

pcxod:
Sorry, definitely a typo, thought I did check it yesterday..., Added the negative signs.

ndevenish:
It doesn't rebase cleanly according to github - was the intention to squash down to one commit or merge in-place? It'd be nice to actually have a followable history. Especially because there is stuff like 
https://github.com/cctbx/cctbx_project/pull/295/files#diff-a82cdfd335599b06c3018b99e5cef5d4R22
introducing unrelated changes from a reverted branch, which makes me wonder how much else of the changes might be inadvertent.

Most the comments/requests I'd make don't apply in context of cctbx_project, but two things I did notice that I think do apply:
- Looks like a new unused import in SConscript, but find_clutter seems to fail to recognise it: https://github.com/cctbx/cctbx_project/pull/295/files#diff-526be5fad8b1db0b877e31f683c5e65aR4
- In https://github.com/cctbx/cctbx_project/pull/295/files#diff-ae86a69c0cb9a5d26f10c4af38baf041R67 , Was this #include supposed to go at the top? Looks a bit weird especially with dedent/indent and I'm not sure I've seen the python 'inline imports' 'style' leaking into C++ before.

pcxod:
@ndevenish  OK, that import is obsolete indeed as it was needed before I stripped some code out... and regarding include - yeah, looks odd, will move it up. But in general you can use it anywhere and at any level of course - like to inject a common 'typedef' blocks into a class declaration... And sorry - have to sneak in one more commit that removes some redundant code (the functions are implemented in the base class but still was duplicated in the derived ones) in normal_eqns_solving.py...

pcxod:
@ndevenish Not sure about the rebase issues, any comments?

bkpoon:
Hi @pcxod,

You can try rebasing the `olex2` branch onto the current `master` branch. There seems to be many conflicts. Once those changes are pushed to your branch, this pull request will update and we should be able to "Rebase and merge."

To rebase the `olex2` branch to `master`,
1. Checkout Pascal's fork (`git clone git@github.com:pascal22p/cctbx_project.git`)
2. Add the original repository as `upstream` (`cd cctbx_project; git remote add upstream https://github.com/cctbx/cctbx_project.git`)
3. Fetch everything from `upstream` (`git fetch upstream`)
4. Checkout the `olex2` branch (`git checkout olex2`)
5. Rebase `olex2` onto the `master` branch of the original repository (`git rebase upstream/master`)
6. You'll see each of you commits be applied on top of the current `master` branch. The process will stop when a merge conflict is encountered. You'll see a line like `CONFLICT (content): Merge conflict in smtbx/refinement/least_squares.h`. Fix merge conflicts by picking the changes. Once that file is fixed, run `git add <fixed file>; git rebase --continue`. The next conflict will appear.
7. Once all the files are fixed, you should re-run any checks to make sure your functionality is the same.
8. Then push the changes to the `olex2` branch (`git push`). You may need to force push by adding the `-f` flag.

I can rebuild once the branch is updated. There may be plugins for editors that make this easier. For example, in Visual Studio Code, I get the option to pick the current or incoming changes.
screen shot 2019-03-07 at 4 26 00 pm

pcxod:
Thanks @bkpoon - I will do that tomorrow.

pcxod:
@ndevenish Did I not ask to merge from a particular branch?

luc-j-bourhis:
@bkpoon Yes, Oleg and I also get conflicts where we try to rebase. Hmmm… This looks a lot of work to fix them actually, especially considering that this pull request merges fine into the official repo. By that I mean that the following does succeed:

```
git clone https://github.com/cctbx/cctbx_project.git official_cctbx_project
cd official_cctbx_project/
git remote add pascal git@github.com:pascal22p/cctbx_project.git
git fetch pascal
git merge pascal/olex2
```

That should be exactly equivalent to the first option of the "Merge pull request" button on this very page, shouldn't it? Do you guys mind if we go that route?

pcxod:
Following Luc, and as I said - the merge request is on the Olex2 branch, nothing else. We will investigate the conflicts internally - thanks to @ndevenish .

phyy-nx:
The biggest reason not to merge here is that since the commits from this pull request go back to summer 2018, if the branch were simply merged into master, it would then be impossible to check out the repository by timestamp and reproduce the code base from last year, since the two histories would be intermingled.  While merging would better represent the development pathway the two branches took and as you say it would be easy, rebase preserves what we have actually been distributing to users.

However, fixing all these conflicts could put us in an unstable state since it's not always obvious what code to keep.  We discussed this here at LBL and we think that in this case, it would be ok to do a merge commit instead of a rebase.

Generally speaking, our policy remains rebasing is better.  Here, the scale of the problem makes it seem that an exception is warranted.

bkpoon:
We looked into seeing if we could rebase your branch yesterday and the number of conflicts is large and properly resolving them would be prone to error, so the green "Merge pull request" button sounds like a good idea.

pcxod:
Thanks guys for looking into this!

ndevenish:
Did you find any code that was accidentally leaking into this branch/where the leaks came from?

If the code history is so complicated as to not be able to figure out what is and isn't intended, that seems... suboptimal.

ndevenish:
> However, fixing all these conflicts could put us in an unstable state since it's not always obvious what code to keep

Also, for future reference, next time you need to do a big rebase it's easy to tell if you've destabilised/badly merged, because you can compare the trees before and after and they'll be identical to the cross-merge.

pcxod:
@ndevenish I did merge cctbx trunk into the branch to make sure there is nothing wrong as well as did pull the branch into cctbx trunk to cross check any conflicts... Thanks anyway!

ndevenish:
Slowly digesting some of this.

Does anyone understand the logic in https://github.com/cctbx/cctbx_project/blob/5e9c814d54a49eda0ecce074e7e783d37a9f0ce1/fast_linalg/SConscript e.g. what it's doing? - it appears to be retroactively monkey-patching itself onto... some other environments? Adding itself to every single linked library dependency? But then what's it doing with includes, injecting the python ones onto `env_no_includes_boost_python_ext`? Even by the standards of `SConscripts` in this project this is hard to read, I'm having real difficulty working out the intention. 

Also, it was my understanding that `dl` didn't usually need to be explicitly linked but it's doing so for non `win32` environments (FYI `win32` is the *only* value [this can be](https://docs.python.org/3/library/sys.html#sys.platform)). Why was this necessary?

Also, it adds 'boost_filesystem' explicitly but the only usage I can find is literally just an `#include` - was this left in accidentally after some previous coding? https://github.com/cctbx/cctbx_project/blob/5e9c814d54a49eda0ecce074e7e783d37a9f0ce1/fast_linalg/ext.cpp#L5

luc-j-bourhis:
@ndevenish As for the Sconscript, libtbx/SConscript defines a do-nothing enable_fast_linalg. If the fast_linalg module has been configured, then a non-trivial enable_fast_linalg is defined. Then a couple of Sconscript's call enable_fast_linalg (scitbx/array_family, scitbx/linalg/boost_python, scitbx/lstbx, scitbx/lstbx/benchmarks, scitbx/matrix/tests, and smtbx). Thus I did use an OO pattern here, instead of  peppering the SConscript's with `if not libtbx.env.has_module('fast_linalg'):`.

pcxod:
@ndevenish - dl is needed as the code does dynamic library loading and linking fails if the library not specified explicitly.
Including the boost/filesystems.hpp is indeed not needed although fast_linalg itself does depend on it through boost/dll/shared_library.hpp and the library is needed at link time only...
Thanks for looking into this!

scitbx/lstbx/normal_equations.h
```diff
@@ -102,7 +103,7 @@ namespace scitbx { namespace lstbx { namespace normal_equations {
                     && b.size()   == w.size())(a.n_rows())(b.size())(w.size());
       SCITBX_ASSERT(a.n_cols() == n_parameters());
       sparse::matrix<scalar_t>
-      at_w_a = a.this_transpose_times_diagonal_times_this(w);
+      at_w_a = a.transpose().this_times_diagonal_times_this_transpose(w);
```
jbeilstenedmands: After this pull request I've noticed that scaling in DIALS, when using Levenberg-Marquardt minimisation (which inherits from the normal_equations code), is around 15x slower. Reverting this line only seems to fix the issue. It looks like this line still does the same computation as before, just a low slower, and with reverting the lstbx tests and smtbx.refinement test still pass. Can this line be reverted or am I missing something regarding the behaviour here? Thanks, James

pcxod: Hi @luc-j-bourhis - the commit message says it should perform better :). Has anything else interfered with it?

jbeilstenedmands: Thanks for pointing to the commit in question, hadn't checked this. If it's relevant, in dials we're typically only refining a few hundred parameters at most, maybe you are refining a lot more and the performance trade-offs change?

ndevenish: I'm a little concerned the fact that commits from over three years ago has apparently been accidentally merged in means that we don't understand the changes that have been pulled in here?

luc-j-bourhis: The change was definitively beneficial in my tests, which were all smtbx-centered. Could you point me toward the place where you use Levenberg-Marquardt in DIALS?

ndevenish: @rjgildea came up with this to demonstrate:
```python
import time
from scitbx.random import variate, uniform_distribution
from scitbx import sparse

m = 10000
n = 100
random_matrices = variate(
    sparse.matrix_distribution(
        m,
        n,
        density=0.6,
        elements=uniform_distribution(min=-3, max=10),
    )
)
w = flex.double_range(0, m)
ww = flex.double(m * m)
ww.resize(flex.grid(m, m))
ww.matrix_diagonal_set_in_place(diagonal=w)

a = random_matrices.next()
t0 = time.time()
for i in range(10):
    b = a.self_transpose_times_diagonal_times_self(w)
t1 = time.time()
print(t1 - t0)
for i in range(10):
    b = a.transpose().self_times_diagonal_times_self_transpose(
        w
    )
t2 = time.time()
print(t2 - t1)
```
though now I notice that the functions aren't necessarily expecting sparse matrices at all.

pcxod: The 'wider' the matrix becomes - the current implementation is faster. So for
`m = 100
n = 1000
`
the timing is approximately the same, for 
`m = 10
n = 1000
`
the timing figure get reversed to the original code snippet. I guess there should be a comparison of the matrix dimension to choose the implementation...

pcxod: Also, as @luc-j-bourhis mentioned - it does perform better for smtbx this way - the restraints matrix is much sparser that 0.6, try density=0.05 and a matrix with smtbx-suitable dimensions like 500x300 (500 parameters for 300 restraints) and also note that the current implementation scales much better then more restraints added. The easiest solution would be to add a bool option to control which sub-procedure is to use.

jbeilstenedmands: Thanks for the insights, my cases are not usually very sparse and the matrix dimensions are quite different also. I agree we could add a bool option but this seems like a bit of a hack - I'd like to review my own code further and look further into what else is available in cctbx to see if I should be taking a different route. I'll make a separate issue to discuss this is necessary. Thanks, James

luc-j-bourhis: If the sparsity of 0.6 is typical of your application in DIALS then I am pretty sure the overhead of my sparse matrix code means blunt dense code would do better. That member equation `add_equations` was designed for sparse problems only. Have you tried to use the `add_equation` member function? It lets you add one equation at a time, or said differently one row of the matrix you are currently passing to `add_equations` at a time.

luc-j-bourhis: If that does not work for you, the clean way would be to add an overload to `add_equations` that takes a dense matrix. That would be really easy.

pcxod: Another way would be to add a function that takes a ready-to-use normal matrix of equations (e.g. computed with SciPy). This will then provide more flexibility regarding what functionality to use to get one...

jbeilstenedmands: Thanks for the suggestions. To add further context to my recent pull request, I had since tried the converting to a dense matrix first as there is already an add_equations that takes a dense matrix. This was slightly quicker than using the new code, but still significantly slower than the old code as there is a loop over rows. I think it just happens that the way I have set up my problem is efficiently matched to the previous implementation, so having the option to switch to the other method of calculation is for now the easiest, although not the most elegant, solution.

PR #303, opened 2019-03-20
Lstbx add equations

jbeilstenedmands:
These changes add an option to choose the method for calculating the A^T W A matrix during add_equation, to enable switching depending on the shape and sparsity of the problem. The default is set to the new behaviour after pull request #295. Setting optimise_for_sparse=False enables one to recover the previous behaviour.

I have verified that I can switch between the old and new behaviour for my use case.
@luc-j-bourhis I would appreciate it if you are able to have a quick check over this and maybe test that everything still works the same for smtbx refinement.

Thanks,
James

### DIFF
--- a/scitbx/lstbx/boost_python/normal_equations.cpp
+++ b/scitbx/lstbx/boost_python/normal_equations.cpp
@@ -35,7 +35,8 @@ namespace boost_python {
         .def("add_equations",
              &wt::add_equations,
              (arg("right_hand_side"), arg("design_matrix"), arg("weights"),
-              arg("negate_right_hand_side")=false))
+              arg("negate_right_hand_side")=false,
+              arg("optimise_for_sparse")=true))
         .def("reset", &wt::reset)
         .def("solve", &wt::solve)
         .add_property("solved", &wt::solved)
@@ -66,11 +67,12 @@ namespace boost_python {
       void (wt::*add_dense_eqns)(af::const_ref<scalar_t> const &,
                                  af::const_ref<scalar_t, af::mat_grid> const &,
                                  af::const_ref<scalar_t> const &)
-        = &wt::add_equations;
+        = &wt::add_equations_dense;
       void (wt::*add_sparse_eqns)(af::const_ref<scalar_t> const &,
                                   sparse::matrix<scalar_t> const &,
-                                  af::const_ref<scalar_t> const &)
-        = &wt::add_equations;
+                                  af::const_ref<scalar_t> const &,
+                                  bool, bool)
+        = &wt::add_equations_sparse;
 
       class_<wt>(name, no_init)
         .def(init<int>(arg("n_parameters")))
@@ -99,7 +101,8 @@ namespace boost_python {
              (arg("residuals"), arg("jacobian"), arg("weights")))
         .def("add_equations",
              add_sparse_eqns,
-             (arg("residuals"), arg("jacobian"), arg("weights")))
+             (arg("residuals"), arg("jacobian"), arg("weights"),
+             arg("negate_right_hand_side")=true, arg("optimise_for_sparse")=true))
         .def("reset", &wt::reset)
         /* We use 'def' instead of add_property for those to stay consistent
            with the other wrappers in this module which can't use properties
--- a/scitbx/lstbx/boost_python/normal_equations.cpp
+++ b/scitbx/lstbx/boost_python/normal_equations.cpp
@@ -36,7 +36,7 @@ namespace boost_python {
              &wt::add_equations,
              (arg("right_hand_side"), arg("design_matrix"), arg("weights"),
               arg("negate_right_hand_side")=false,
-              arg("optimise_for_sparse")=true))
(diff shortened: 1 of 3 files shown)


pcxod:
I would still insert the 'middle' function by splitting lstbx_add_equations that creates the normal matrix for the given equations and then adds them to the current matrix into two - this way the one can pre-build it as is more suitable for their problem and then add directly...

luc-j-bourhis:
@jbeilstenedmands Is there a reason you want to keep passing a sparse matrix? What you have renamed as `non_linear_ls::add_equations_dense`,  does it not work for you? I mean this sparse code module is quite naive an implementation: I only intended it for really sparse problems. A point I did not document, fair enough. Hence my question.

jbeilstenedmands:
For scaling, some of my model components are sparse, so I keep things in sparse matrices during the Jacobian calculation and I think it helps a lot. I have tried converting this directly to a dense matrix to pass it through the overloaded add_equations (I changed the name back to the original by the way), but as the matrix typically has lots of rows ~O(10K) the loop in add_equations slows things down (on a few tests datasets the slow-down is around a factor of 4). Perhaps it is possible to do this another way in a vectorised manner, and I could refactor my code to remove the sparseness earlier and vectorise things, but this would be a significant undertaking. Additionally, we're preparing our 1.14 release at the moment in DIALS, so I desire a quick fix at the moment, which could be superseded in future.
@pcxod I'll add in the middle function to give the possibility for going direct.

Anthchirp:
@jbeilstenedmands #295 did not make it into 1.14 and I have no intention of merging it into a release branch.

luc-j-bourhis:
Ok, if it works better for you, then go along. One details then. I would prefer a naming that reads better in English: `add_dense_equations` instead of `add_equations_dense` etc. As a side note, I don't understand the reason of this huge performance difference between `this_transpose_times_diagonal_times_this` and `this_times_diagonal_times_this_transpose` depending on sparsity. I had the idea that it would be easier if the former could switch to the latter auto-magically depending on sparsity, so that we don't need to touch any higher-level API but I am a bit reluctant to pursue this course without a proper understanding of what is going on. Your solution is an acceptable stopgap for now and for everybody, so +1.

jbeilstenedmands:
Looking over the code, it looks like the number of operations in each case is about the same (as it should be), so maybe the time is in the overheads somewhere.

Here is some benchmarking run using @rjgildea's code (posted to the other pull request),
the performance seems heavily dependent on the shape of the matrix for a given density:
old is  ```a.self_transpose_times_diagonal_times_self(w)```
new is ```a.transpose().self_times_diagonal_times_self_transpose(w)```

For a matrix of 1000 x 10 (rows x columns):
Density: 0.01
Old: 0.00010
New: 0.00044

Density: 0.1
Old: 0.00012
New: 0.00171

Density: 0.5
Old: 0.00130
New: 0.01244

For a matrix of 10 x 1000
Density: 0.01
Old: 10.42219
New: 0.00279

Density: 0.1
Old: 10.21473
New: 0.02709

Density: 0.5
Old: 9.77336
New: 0.78524

For a matrix of 100 x 100
Density: 0.01
Old: 0.00784
New: 0.00028

Density: 0.1
Old: 0.00818
New: 0.00384

Density: 0.5
Old: 0.02199
New: 0.09372

So when the Jacobian has n_rows >> n_cols, the new code is slower even for low density (and the difference gets worse as density increases), however for n_rows ~ n_cols, the new code is much more efficient for low density, but worse when density is 0.5 (which is admittedly not what it was designed for).

luc-j-bourhis:
Thanks for doing that benchmarking. In smtbx, `add_equations` is used only to add the restraints to the normal matrix. So the sparse matrix which is passed is the design matrix with one row per restraint and one column per refined parameter. Thus n_rows < n_cols: we are in the second case you tested where "New" trounces "Old". So I gather you are in the case n_rows >> n_cols as, I guess, you deal with the design matrix of the model vs data L.S. fit, right?

jbeilstenedmands:
That makes sense, yes you are correct I am adding the design matrix of the model vs data, so ~100 columns and >1000 rows. I now see that optimise_for_sparse isn't a very good name for controlling switching between the two methods, would a boolean option called optimise_for_long_matrix be a better name?

luc-j-bourhis:
A tall matrix like yours is more typical of L.S. optimisation, so I think I would use a flag optimise_for_tall_matrix with a default of True. Then in smtbx we can set that to False.

pcxod:
Just a comment...
> English: `add_dense_equations` instead of `add_equations_dense`
We do programming and surely it sounds better in English but is much harder to find! I think we should try to keep the context in the name...

luc-j-bourhis:
With regular expressions, both are easy to search. But this is cosmetic, I am fine if you keep `add_equations_dense`.

jbeilstenedmands:
Okay, I've updated the name of the flag to optimise_for_tall_matrix, and set the default to False in smtbx.refinement.least_squares.py. The name of the method in lstbx.normal_equations.h remains add_equations for both methods as previously. Thanks, James.

PR #311, opened 2019-03-28
Reverse the order NeXus reads in data_origin and data_size

### LINKED ISSUE #305: NeXus: get_detector_module_slices reads data dimensions backwards
In nexus.py, the function get_detector_module_slices is used to figure out how to slice the raw pixel data into panels.  However, I think this function is interpreting the NeXus specification wrongly.  Compare:

http://download.nexusformat.org/sphinx/classes/applications/NXmx.html

> data_origin: (required) NX_INT
> 
> A dimension-2 or dimension-3 field which gives the indices of the origin of the hyperslab of data for this module in the main area detector image in the parent NXdetector module.
> 
> The data_origin is 0-based.
> 
> The frame number dimension (np) is omitted. The order of indices is i, j or i, j, k, i.e. slow to fast.

To:

https://github.com/cctbx/cctbx_project/blob/master/dxtbx/format/nexus.py#L1638

`Assumes slices are stored in NeXus in fast to slow order.`

This affects the EuXFEL work I've been doing.  I've been writing NeXus files such that dxtbx can understand them, but here it seems we have the order backwards.

The source of this confusion is likely because of nexusformat/definitions#650.

I know DLS is writing NeXus files.  What are the data_origin/data_size fields being created there?  Are the panels square so this wouldn't have mattered?  If so we can fix this without affecting anyone.

phyy-nx:
This is to match spec. Fixes #305.

See http://download.nexusformat.org/sphinx/classes/applications/NXmx.html

### DIFF
--- a/dxtbx/format/nexus.py
+++ b/dxtbx/format/nexus.py
@@ -1186,8 +1186,8 @@ class DetectorFactoryFromGroup(object):
             for module_number, nx_detector_module in enumerate(modules):
                 # Get the depends_on starting point for this module and for this image
                 panel_name = str(os.path.basename(nx_detector_module.handle.name))
-                px_fast = int(nx_detector_module.handle["data_size"][0])
-                px_slow = int(nx_detector_module.handle["data_size"][1])
+                px_fast = int(nx_detector_module.handle["data_size"][-1])
+                px_slow = int(nx_detector_module.handle["data_size"][-2])
                 image_size = px_fast, px_slow
 
                 # Get the trusted range of pixel values
@@ -1406,7 +1406,7 @@ class DetectorFactory(object):
 
         # Construct the detector model
         pixel_size = (fast_pixel_direction_value, slow_pixel_direction_value)
-        image_size = tuple(map(int, nx_module["data_size"][0:2]))
+        image_size = tuple(map(int, nx_module["data_size"][-2:]))
 
         self.model = Detector()
         self.model.add_panel(
@@ -1636,7 +1636,7 @@ def get_detector_module_slices(detector):
     Helper function to read data_origin and data_size from the NXdetector_modules in a
     NXdetector.  Returns a list of lists, where each sublist is a list of slices in
     slow to fast order.
-    Assumes slices are stored in NeXus in fast to slow order.
+    Assumes slices are stored in NeXus in slow to fast order.
 
     """
     # get the set of leaf modules (handles nested modules)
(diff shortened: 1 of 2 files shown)


graeme-winter:
I would be more comfortable if there was a failing test associated with this one! That said the changes look sensible, and (I think) will not impact processing of Eiger data sets which is the only rotation data set use case which would touch these at the moment.

dxtbx/format/nexus.py
```diff
@@ -1406,7 +1406,7 @@ def __init__(self, obj, beam):
 
         # Construct the detector model
         pixel_size = (fast_pixel_direction_value, slow_pixel_direction_value)
-        image_size = tuple(map(int, nx_module["data_size"][0:2]))
+        image_size = tuple(map(int, nx_module["data_size"][-2:]))
```
graeme-winter: Since we are aiming for clarity here I wonder if it is worth renaming `image_size` to `image_size_slow_fast` or similar? Discussion topic, since reviewing.

dxtbx/format/nexus.py
```diff
@@ -1186,8 +1186,8 @@ def set_frame(pg, transformation):
             for module_number, nx_detector_module in enumerate(modules):
                 # Get the depends_on starting point for this module and for this image
                 panel_name = str(os.path.basename(nx_detector_module.handle.name))
-                px_fast = int(nx_detector_module.handle["data_size"][0])
-                px_slow = int(nx_detector_module.handle["data_size"][1])
+                px_fast = int(nx_detector_module.handle["data_size"][-1])
```
graeme-winter: Below we store the size as a tuple and use `map()` - would be cleaner to do the same here?

dxtbx/format/nexus.py
phyy-nx: Oh snap, you found a big bug.  That should have been

`        image_size = tuple(reversed(map(int, nx_module["data_size"][-2:])))`

Since the data is stored slow to fast, [-2:] gives us the image size in slow, fast order.  But dxtbx needs it in fast, slow order.  So if we add the reversed call, we get now match the NeXus spec.  But for one of the Eiger nexus files, print header now shows:

  image_size: {4362,4148}

Which is backwards...

phyy-nx: I didn't catch this because my use case is multipanel.  And since the Eiger is backwards, this backwards code worked...  I think the only fix is to special case it...

phyy-nx: Chat w/ @graeme-winter and he agrees, no choice but to special case it...  Commit incoming.

phyy-nx:
Fixed and special cased.  Merging.

PR #312, opened 2019-03-29
Lazy experiment list

### LINKED ISSUE #307: Need a lazy experiment list
As of #284, when creating an experiment list from a set of image files, the function from_stills_and_crystal is used to populate the set of experimental models from the imageset onto the experiments.  See:

https://github.com/cctbx/cctbx_project/blob/master/dxtbx/model/experiment_list.py#L729-L744

This function uses a for loop to read all the models in and populate the list.  However, this leads to huge start-up times for big hdf5 files due to how long it takes to load the metadata.

ImageSetLazy was developed to solve this issue, such that the models were only read when asked for.  We needs something similar for experiment lists, it appears, and when ready, we need to make FormatNexusStill use ImageSetLazy.

phyy-nx:
This pull request adds lazy loading of experiments.  With this change, the loop in from_stills_and_crystal is greatly sped up as the individual models are not loaded for each experiment.  With this change, the time to load a big dataset greatly improves. This is vital for multiprocessing, since each process cannot be expected to read all the models when it will only handle a small subset of them.

Timing example snippet:

```
from dxtbx.model.experiment_list import ExperimentListFactory
import sys

experiments = ExperimentListFactory.from_filenames([sys.argv[1]])
print len(experiments)
```

With this snippet and an EuXFEL NeXus file with 32960 frames, the time to load the dataset using from_filenames improves from ~34.4s to ~7.7s.

Fixes #307.

Details:
- Add the attribute lazy = false to the Experiment constructor.  If true, the first time the detector, beam, goniometer, or scan are accessed, the function load_models is called. This function uses the experiment's imageset to go load the models.
- Change function from_stills_and_crystal to test if the imageset is an ImageSetLazy. If so, don't read the models up front and instead create the Experiments with lazy=True.

Notes:
- Because functions like Experiment.get_beam can now modify the beam (from None to a loaded beam), get_beam can't be const.  This caused lots of functions to lose the const keyword.
- load_models needs to use the python imageset methods which in this case will be ImageSetLazy methods.  That class uses the python format object load the models and cache them so for that reason load_models calls up and out of the C++ code into the python code.

### DIFF
--- a/dxtbx/format/FormatHDF5SaclaMPCCD.py
+++ b/dxtbx/format/FormatHDF5SaclaMPCCD.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 from dxtbx.format.FormatHDF5 import FormatHDF5
 from dxtbx.format.FormatStill import FormatStill
+from dxtbx.format.FormatMultiImageLazy import FormatMultiImageLazy
 
 # 151028: deepcopying this class causes crash in h5py
 #         temporary fix by closing the file in every methods(!)
@@ -12,7 +13,7 @@ from dxtbx.format.FormatStill import FormatStill
 # 180724: update 'understand' to exclude Rayonix data
 
 
-class FormatHDF5SaclaMPCCD(FormatHDF5, FormatStill):
+class FormatHDF5SaclaMPCCD(FormatMultiImageLazy, FormatHDF5, FormatStill):
     """
     Class to handle multi-event HDF5 files from MPCCD
     preprocessed by Cheetah SFX pipeline at SACLA.
--- a/dxtbx/model/boost_python/experiment.cc
+++ b/dxtbx/model/boost_python/experiment.cc
@@ -23,7 +23,7 @@ namespace dxtbx { namespace model { namespace boost_python {
 
   struct ExperimentPickleSuite : boost::python::pickle_suite {
     static
-    boost::python::tuple getinitargs(const Experiment &obj) {
+    boost::python::tuple getinitargs(Experiment &obj) {
       return boost::python::make_tuple(
         obj.get_beam(),
         obj.get_detector(),
@@ -87,7 +87,8 @@ namespace dxtbx { namespace model { namespace boost_python {
           boost::python::object,
           boost::python::object,
           boost::python::object,
-          std::string>((
+          std::string,
+          bool>((
               arg("beam") = boost::shared_ptr<BeamBase>(),
               arg("detector") = boost::shared_ptr<Detector>(),
               arg("goniometer") = boost::shared_ptr<Goniometer>(),
@@ -96,7 +97,8 @@ namespace dxtbx { namespace model { namespace boost_python {
               arg("profile") = boost::python::object(),
               arg("imageset") = boost::python::object(),
               arg("scaling_model") = boost::python::object(),
-              arg("identifier") = "")))
+              arg("identifier") = "",
+              arg("lazy") = false)))
       .add_property(
           "beam",
           &Experiment::get_beam,
--- a/dxtbx/model/boost_python/experiment_list.cc
+++ b/dxtbx/model/boost_python/experiment_list.cc
@@ -142,17 +142,17 @@ namespace dxtbx { namespace model { namespace boost_python {
   struct experiment_list_indices_pointers {
 
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*beam_type)(
-        const boost::shared_ptr<BeamBase>&)const;
+        const boost::shared_ptr<BeamBase>&);
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*detector_type)(
-        const boost::shared_ptr<Detector>&)const;
+        const boost::shared_ptr<Detector>&);
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*goniometer_type)(
-        const boost::shared_ptr<Goniometer>&)const;
+        const boost::shared_ptr<Goniometer>&);
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*scan_type)(
-        const boost::shared_ptr<Scan>&)const;
+        const boost::shared_ptr<Scan>&);
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*crystal_type)(
         const boost::shared_ptr<CrystalBase>&)const;
     typedef scitbx::af::shared<std::size_t> (ExperimentList::*object_type)(
-        boost::python::object)const;
+        boost::python::object);
 
     static beam_type beam() {
       return &ExperimentList::indices;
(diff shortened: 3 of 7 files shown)


Anthchirp:
What's the drawback? ie. why should this not be the default in all cases?

phyy-nx:
That's a great question.  This lazy mode will be the default for XTC and HDF5, the two multi image still formats I know about.  I don't have any still support set up for Gatan or SER, the other two formats that are multi-image and could therefore be lazy stills.

For sweeps, it likely doesn't do anything, since there is usually only one experiment with one set of models.

For multi-experiment processing (many wedges in h5 files), maybe there's something to wanting lazy support.  However, the time savings only starts kicking in with thousands of experiments.  At EuXFEL we had a 190K image file that simply wouldn't load, which really precipitated this change.  If that were a sweep it would load instantly since there's only a couple models.

phyy-nx:
I tried it, modifying the from_sweep_and_crystal function:

```
$ git diff
diff --git a/dxtbx/model/experiment_list.py b/dxtbx/model/experiment_list.py
index f3630323a..63e964706 100644
--- a/dxtbx/model/experiment_list.py
+++ b/dxtbx/model/experiment_list.py
@@ -715,11 +715,12 @@ class ExperimentListFactory(object):
             [
                 Experiment(
                     imageset=imageset,
-                    beam=imageset.get_beam(),
-                    detector=imageset.get_detector(),
-                    goniometer=imageset.get_goniometer(),
-                    scan=imageset.get_scan(),
                     crystal=crystal,
+                   lazy=True
                 )
             ]
         )
```

But I did get a test failure (test_search_beam_position.py).  I'd want to think about this more carefully before using it for sweeps.

jmp1985:
Hi Aaron,

I've had a look and I have a few of concerns. 

1. It would be desirable to retain const guarantees for accessing the experimental models but this is a minor thing.

2. When we moved the experiment list to C++, the rationale was that we wanted to use thread level parallelism which couldn't be achieved in Python. Since the experimental models were already in C++, we decided that by instantiating the models in advance we could access the experimental models from C++ code that might be in a worker thread. Obviously, the image reading is still done via calling back into python but this couldn't really be helped so it is done in the master thread. By having the possibility of calling back into Python, this makes those calls potentially unsafe.

3. If the experimental models are supplied to the Experiment in the constructor (or otherwise) and lazy is True, then it seems that when you get the model it will be overridden in by the models from the imageset. This might cause subtle problems. For example, lets say we have a refined geometry and the create an experiment list with Lazy == True then when we get the models we will be returned the initial unrefined models - this is the kind of thing that sometimes takes a while to pick up when it goes wrong.

From my understanding, the problem you want to solve is that, during an initial import, when you have many images, the import is done in serial and is too slow. Also, from my understanding, what you would like to do is have no metadata loaded initially and then, when you spawn a load of parallel jobs with a single job for a single image, the metadata is loaded for each image independently, thereby reading the metadata in parallel rather than in serial. 

If that is more or less correct, then the only time we need to be lazy is during import; therefore, it might be better to make a fix in dials.import such that import doesn't load any models and then make a change in stills process to load the models where they are needed. I think this might be preferable because it reduces complexity (having models loaded behind the scenes might causes surprises) and make a change that only affects the bit that needs fixing rather than affecting the behaviour everywhere. 

So for example, we could set all models to None during import and then have a load models from imageset method that we can call manually once the parallel job is started. This obviously requires an additional function call in stills process. 

What do you think? Have I understood the problem correctly? Or are there additional complications?

graeme-winter:
"So for example, we could set all models to None during import and then have a load models from imageset method that we can call manually once the parallel job is started."

@jmp1985 nice idea - if model is None: load model. We just then have a list of "empty" experiments of the same length as the list of data points. At a glance your concerns above are valid - we have been caught out by things like this before.

phyy-nx:
Good comments.  I hadn't considered how this would affect threading.  Lemme try doing it all in dials.stills_process instead.

ndevenish:
I'm also slightly fuzzy on the use case - I can either see
- You haven't got a model yet, in which case you *need* to read the file anyway
- You have read the model from the file in a previous step in which case you've already got the model (serialized experiment is cheap).

The only case I can think of immediately is e.g. importing on one node, building the experiment and then passing to a worker to do the processing - but if that's the case, wouldn't it be simpler to simply pass the filename/image number to the worker for it to do the import itself?

phyy-nx:
The use case is MPI processing where the models (such as wavelength) are
split across many files.  Example: 180K images using a NeXus master linking
to many sub file, processed using 1000 cores.  The wavelength is stored in
a 180K array of values split across many sub files.

Each process should only read 180 pixel arrays and construct 180 beam
models, detector models, etc.

ndevenish:
> Each process should only read 180 pixel arrays and construct 180 beam models, detector models, etc.

Then how does this help? This code seems to assume that you are forced to construct 180k experiments when you only want 180. Can't you just, only construct the 180? I'm not understanding why this is a bottleneck on the master - presumably at the moment this is done on the master and Experiments serialised to the slave nodes. But why can't it just be constructed on the slave node in the first place?

`Experiment` is a simple, well understood object with well defined behaviour. I'd be very reluctant to change that without an extremely good reason.

phyy-nx:
The goal is to make this line:

experiments = ExperimentListFactory.from_filenames('master.h5')

Return as fast as possible.  See:
https://github.com/dials/dials/blob/master/command_line/stills_process.py#L197

My experience with a 30K image file is that it takes 30s.  For a 180K image
file I gave up waiting.  It essentially never returned.

In MPI, each process has to execute this line.  If only one process
executes it and splits the resultant set of experiments using mpi_send,
that's a ton of time wasted for each of the other cores while the master
reads the file and constructs the experiment list.  It's best to delay
model construction as long as possible to maximise usage of the cores.

I'm in agreement though that this pull request doesn't seem to be the best
way.  Alternative inbound.

phyy-nx:
Alternate version is #313

ndevenish:
Ah - did you profile to find what was causing the slowdown? Obviously it should only take 6x as long.

I seem to remember datablock loading had some O(N²) lookup actually building the output list (I think, demultiplexing the models), along with some badly scaling file reading functions which you probably aren't hitting here (the purpose of [this branch](https://github.com/ndevenish/cctbx_project/compare/987b85eeafa3dacf8be3275caa21f8057e06e241...fasterimport) was to fix them, and did, but was put on hold until after the ExperimentList changes went in and the file reading mainly only affected cbf data, which we were moving away from).

I remember @jmp1985 telling me that the ExperimentList functions still just use the datablock code behind the scenes so wouldn't be surprised if it were still there.

phyy-nx:
I didn't profile the code.  Mainly because I figure that it's format class specific.  But yah, the new experiment list code does use the datablock code under the hood.

phyy-nx:
Closing.  Resolved instead with #313.

PR #313, opened 2019-04-02
Add option to delay load models when reading raw data

### LINKED ISSUE #307: Need a lazy experiment list

phyy-nx:
Adds load_models as option to the experiment list factory functions.  If False, skip reading the models from the imageset when constructing experiment lists from raw data files.

Closes #307.

This version isn't a general solution like #312 is, but it works.  A concurrent change will need to be made to dials.stills_process.  See dials/dials#744.

### DIFF
--- a/dxtbx/model/__init__.py
+++ b/dxtbx/model/__init__.py
@@ -435,6 +435,14 @@ class MosaicCrystalSauter2014Aux(CrystalAux, MosaicCrystalSauter2014):
 
         return xl
 
+class ExperimentAux(boost.python.injector, Experiment):
+    def load_models(self, index = None):
+        """ Load the models from the imageset """
+        if index is None: index = 0
+        self.beam       = self.imageset.get_beam(index)
+        self.detector   = self.imageset.get_detector(index)
+        self.goniometer = self.imageset.get_goniometer(index)
+        self.scan       = self.imageset.get_scan(index)
 
 class ExperimentListAux(boost.python.injector, ExperimentList):
     def __repr__(self):
--- a/dxtbx/tests/model/experiment/test_experiment_list.py
+++ b/dxtbx/tests/model/experiment/test_experiment_list.py
@@ -750,6 +750,59 @@ def test_experimentlist_with_identifiers():
     experiments.remove_on_experiment_identifiers(["spam", "jam"])
     assert list(experiments.identifiers()) == ["bacon", "ham"]
 
+def test_load_models(dials_regression):
+    import dxtbx
+
+    filename = os.path.join(
+        dials_regression,
+        "image_examples",
+        "SACLA_MPCCD_Cheetah",
+        "run266702-0-subset.h5",
+    )
+
+    # Test different ways of loading the data
+    waves1, waves2, waves3, waves4, waves5 = [], [], [], [], []
+    oris1, oris2, oris3, oris4, oris5 = [], [], [], [], []
+
+    img = dxtbx.load(filename)
+
+    # Test using dxtbx directly
+    for i in xrange(img.get_num_images()):
+        waves1.append(img.get_beam(i).get_wavelength())
+        oris1.append(img.get_detector(i)[0].get_origin())
+
+    # Test using the imageset clases
+    imageset = img.get_imageset(filename)
+    for i in xrange(len(imageset)):
+        waves2.append(imageset.get_beam(i).get_wavelength())
+        oris2.append(imageset.get_detector(i)[0].get_origin())
+
+    # Test using imageset subsets
+    imageset = img.get_imageset(filename)
+    for i in xrange(len(imageset)):
+        subset = imageset[i:i+1]
+        waves3.append(subset.get_beam(0).get_wavelength())
+        oris3.append(subset.get_detector(0)[0].get_origin())
+
+    # Test using pre-loaded experiments
+    experiments = ExperimentListFactory.from_filenames([filename])
+    for experiment in experiments:
+        waves4.append(experiment.beam.get_wavelength())
+        oris4.append(experiment.detector[0].get_origin())
+
+    # Test using post-loaded experiments
+    experiments = ExperimentListFactory.from_filenames([filename], load_models = False)
+    for experiment in experiments:
+        assert experiment.beam is None and experiment.detector is None
+        experiment.load_models()
+        waves5.append(experiment.beam.get_wavelength())
+        oris5.append(experiment.detector[0].get_origin())
+
+    for w1, w2, w3, w4, w5 in zip(waves1, waves2, waves3, waves4, waves5):
+        assert w1 == w2 == w3 == w4 == w5
+
+    for o1, o2, o3, o4, o5 in zip(oris1, oris2, oris3, oris4, oris5):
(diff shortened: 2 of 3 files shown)


phyy-nx:
Note, only #312 or #313 should be merged.  I'm leaning towards this one (#313).

dxtbx/tests/model/experiment/test_experiment_list.py
```diff
@@ -750,6 +750,59 @@ def test_experimentlist_with_identifiers():
...
+    )
+
+    # Test different ways of loading the data
+    waves1, waves2, waves3, waves4, waves5 = [], [], [], [], []
+    oris1, oris2, oris3, oris4, oris5 = [], [], [], [], []
+
+    img = dxtbx.load(filename)
+
+    # Test using dxtbx directly
+    for i in xrange(img.get_num_images()):
```
Anthchirp:
```suggestion
    for i in range(img.get_num_images()):
```
Python2/3 syntax

dxtbx/tests/model/experiment/test_experiment_list.py
```diff
@@ -750,6 +750,59 @@ def test_experimentlist_with_identifiers():
...
+        waves1.append(img.get_beam(i).get_wavelength())
+        oris1.append(img.get_detector(i)[0].get_origin())
+
+    # Test using the imageset clases
+    imageset = img.get_imageset(filename)
+    for i in xrange(len(imageset)):
```
Anthchirp:
```suggestion
    for i in range(len(imageset)):
```
Python 2/3 syntax

dxtbx/tests/model/experiment/test_experiment_list.py
```diff
@@ -750,6 +750,59 @@ def test_experimentlist_with_identifiers():
...
+        waves2.append(imageset.get_beam(i).get_wavelength())
+        oris2.append(imageset.get_detector(i)[0].get_origin())
+
+    # Test using imageset subsets
+    imageset = img.get_imageset(filename)
+    for i in xrange(len(imageset)):
```
Anthchirp:
```suggestion
    for i in range(len(imageset)):
```
Python 2/3 syntax

dxtbx/model/__init__.py
```diff
@@ -435,6 +435,14 @@ def from_dict(d):
 
         return xl
 
+class ExperimentAux(boost.python.injector, Experiment):
+    def load_models(self, index = None):
+        """ Load the models from the imageset """
+        if index is None: index = 0
```
ndevenish: I think this is going to cause problems

dxtbx/tests/model/experiment/test_experiment_list.py
```diff
@@ -750,6 +750,59 @@ def test_experimentlist_with_identifiers():
...
+    experiments = ExperimentListFactory.from_filenames([filename])
+    for experiment in experiments:
+        waves4.append(experiment.beam.get_wavelength())
+        oris4.append(experiment.detector[0].get_origin())
+
+    # Test using post-loaded experiments
+    experiments = ExperimentListFactory.from_filenames([filename], load_models = False)
+    for experiment in experiments:
+        assert experiment.beam is None and experiment.detector is None
+        experiment.load_models()
```
ndevenish: I think(?) this is wrong. The definition of `load_models` loads imageset index zero if not passed an index, so it's not actually loading the model for the image matching the experiment, just the same one over and over again. A test that the output values don't match would expose this?

ndevenish:
This is better, because it's less intrusive, but I feel it's still applying a bandaid rather than fixing the underlying problem. I worry that it's just offloading more work into the future, and that we will have to start both
- Carrying an index around with, but separate from, an Experiment, in order to load the experiment correctly - it looks like this mistake has been made in even this pull request. It might be better to assign an internal-mangled attribute specifying which index to load - or giving each instance a special `load_models` closure that has captured the index
- Potentially having to check everything every time, e.g.
```python
if ex.beam is None:
    ex.load_models(model_imageset_index)
if ex.beam is None:
    # do something here
```
It also makes it hard to know what to do with these from C++, but I'll accept this as a limitation of "not a general solution"

I'm assuming the problem is that even if you offload the import to the workers, that because the `from_filename` (or equivalent) loads every experiment before narrowing down a subset, and so this necessitates reading every image model, hitting whatever-N² process for large files. I also assume that there's nothing that requires reading everything before selecting a subset of experiments?

How about altering the load interface so that you can ask directly for only a subset of experiments from an imageset e.g. something along the lines of

```python
stills_from_imageset_and_crystal(imageset, crystal, range=(start, end))
```
The master process can load the imageset and introspect the minimal number of expected models, then just pass the imageset source data and range to each worker node, who only needs to load the necessary models.

Would this be a better long-term solution?

ndevenish:
Also; I understand that some of this might be time-critical? If that's the case and it's impossible to set up working from branches in time, this one is unobtrusive enough that I'd be okay with pushing it in now, with the aim to replace with something better later.

phyy-nx:
Hi @ndevenish, thanks for looking at it.  If I understand you, you are concerned about the index parameter to load_models.  I think it would help to review the two functions on master, from_sweep_and_crystal and from_stills_and_crystal:
https://github.com/cctbx/cctbx_project/blob/master/dxtbx/model/experiment_list.py#L711-L742

These are the two functions that from_filenames uses.  In the sweeps version, the models are loaded with no index, since a sweep's imageset should only have one model.  For the stills version, the imageset is sliced up into singletons, so the index is always zero.

These are the only two use cases supported so the default of zero works fine.

I don't understand your function signature.  An experiment can't have a range of models, only one of each kind of model.

As for C++, the lazy image set code is in python because it needs to access the format objects.  Therefore it makes sense for load_models to also be in python.

phyy-nx:
Oh, I understand your function signature now.  You are saying to have from_filenames only read a subset of models.  The problem there is that there would still be latency during startup.  It's better to delay loading until you need to for stills processing.

ndevenish:
> These are the two functions that from_filenames uses. In the sweeps version, the models are loaded with no index, since a sweep's imageset should only have one model. For the stills version, the imageset is sliced up into singletons, so the index is always zero.

Oh, I see, it splits the imageset into single-image imagesets. So, can we have the function not taking an index parameter at all - there seems to be no scenario (certainly no tested scenario) where it's used.

Otherwise, I'm still not entirely sure why this is better than some other approach, but am happy for it to go in and work out something better later.

phyy-nx:
I didn't want to close that door and remove the index parameter entirely.

I'll put in the xrange changes now and merge.  Thanks.

PR #315, opened 2019-04-03
Support multi-image masks in nexus. Resolves #308.

### LINKED ISSUE #308: Need per-image mask support in NeXus
Assuming that per-image mask support is approved in NXmx (see  nexusformat/definitions#651), we need to add it in nexus.py to support the EuXFEL.  Currently, it is assumed the mask provided has dimensionality 1 less than the data array, meaning the mask applies to all the data.  However, the proposal is that if the mask has the same dimensionality as the data, then each image has a separate mask.

Example:
Data shape: nimages x slow x fast

Mask shape 1: slow x fast. This mask applies to the whole dataset.
Mask shape 2: nimages x slow x fast.  This array contains one mask per image.

phyy-nx:
Now that nexusformat/definitions#654 is merged, we can add the multi-image mask to the nexus format class.

### DIFF
--- a/dxtbx/format/FormatHDF5EigerNearlyNexus.py
+++ b/dxtbx/format/FormatHDF5EigerNearlyNexus.py
@@ -348,7 +348,7 @@ class FormatHDF5EigerNearlyNexus(FormatHDF5):
 
         # Get the NXmx model objects
         entry = reader.entries[0]
-        instrument = entry.instruments[0]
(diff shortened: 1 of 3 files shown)


phyy-nx:
Hoping to merge this later today :)

phyy-nx:
Thanks.  I have no more pending dxtbx changes.  I'll start looking at the dxtbx move branch later today.

PR #316, opened 2019-04-05
mmtbx.tls pull request

nicholas-pearce:
Code changes:
- renaming of mmtbx tests not following libtbx convention. 
- enable pytest discovery in mmtbx.
- add missing assert statements in mmtbx tls tests.
- fix typo in mmtbx/tls/analysis.py
New code:
- c++ implementation of TLS-matrix decomposition (mmtbx/tls/decompose*). 
- c++ classes for manipulation of TLS Matrices (mmtbx/tls/utils*). 
- Class and c++ optimiser for fitting arbitrary combinations of Us (e.g. from TLS models) to sets of refined Us (mmtbx/tls/optimise_amplitudes*). 
- new implementations include boost::python wrappers and tests in pytest.


olegsobolev:
After discussion with @bkpoon and @pafonine we suggest to move forward with this pull request as follows:
1. Make the code consistent with mmtbx style and with other files in this pull request:
  - 2 spaces indentation in Python and C++ files (you have both 2 and 4 spaces in this pull request),
  - libtbx style tests instead of pytests,
  - run libtbx.find_clutter, libtbx.find_unused_imports_crude (I spotted IPython's embed not used),
2. I make sure tests work
3. @bkpoon makes sure everything compiles everywhere
4. We merge.
5. @pafonine inspects the new functionality when his schedule permits and probably removes python code path and only use C++ implementation.

nicholas-pearce:
I have made the requested changes. 
- The tests are now "libtbx style", and registered in run_tests, where I have removed pytest discovery. I have changed the test functions to use libtbx functions (e.g. approx -> approx_equal). However, I am still using a pytest function in the tests ("raises()"), which allows for checking Exceptions are raised in 3 lines of code as opposed to 6 lines with "try... excepts..." -- I hope this is acceptable. 
- libtbx.find_clutter and libtbx.find_unused_imports_crude run cleanly for me.
- all tests run cleanly and build runs cleanly for me (on Mac).

bkpoon:
How portable do you need this code to be? It looks like there are variable initializations in some header files and that will not compile on Windows and older versions of gcc. When we move to C++11 as the default, this shouldn't be a problem, but currently not all the tests have been validated to work when compiled with C++11.

If you need this code to work on all platforms, then the implementation should be moved into a `.cpp` file and only do declarations in the header. Also, do you intend for these header files to be included in many other files? Any time a change is made to these header files, any file that includes these headers would need to be recompiled.

The `pytest` import is fine for now. Someone will probably change it later or add the `raises` functionality.

bkpoon:
Some lingering issues remain on Windows to be sorted out.

PR #324, opened 2019-04-26
Invert the BOOST FPE trap defaults.

Anthchirp:
## Executive summary

FPE exceptions in Python inside development installations are no longer trapped by default, making the Python behaviour consistent with an environment from an installer. C++ exceptions continue to be trapped by default.

### How do I restore previous behaviour
To re-enable traces set the environment variable `BOOST_ADAPTBX_ENABLE_TRACE`,
to re-enable FPE crashes set `BOOST_ADAPTBX_TRAP_FPE`,
to re-enable python overflow crashes set `BOOST_ADAPTBX_TRAP_OVERFLOW`.

Traps can be enabled for specific code sections using the context handler:
```python
import boost.python

with boost.python.trapping(overflow=True):
  inf_value = 1e300
  while inf_value != inf_value*inf_value:
    inf_value *= inf_value   # crash
```

### Details
Right now the following two commands behave differently on any development installation:
```libtbx.python -c "import math; print math.ldexp(1.0,1023)*2"```
```libtbx.python -c "import boost.python;import math; print math.ldexp(1.0,1023)*2"```
The first command prints out `inf`, while the second command crashes the python interpreter with an uncatchable FPE exception. When run on a installation from an installer _both_ commands will print `inf`.

Importing the `boost.python` package changes the behaviour of Python (the language) for development installations. This does not necessarily pose a problem within cctbx, as we can work around these cases. However this may cause 3rd party libraries to crash in unexpected ways, including widely used libraries such as pyyaml and pytables. Our work-around usually involves setting the environment variables `BOOST_ADAPTBX_FPE_DEFAULT`and `BOOST_ADAPTBX_SIGNALS_DEFAULT` in dispatchers. This is currently the case for 53 dispatchers in cctbx_project, 18 in dials, and 4 in xia2. This results in code paths within cctbx_project and dials that are not supposed to be executed by dispatchers that do not carry this flag. The code paths are of course not marked as such, and therefore we continue to encounter these FPE issues on a regular basis.

### Current behaviour
To investigate the actual effects of the FPE trap logic I ran the following three commands:
**A**) ```libtbx.python -c "import boost.python; import math; print math.ldexp(1.0,1023)*2"```
**B**) ```libtbx.python -c "import boost.python; print boost.python.ext.divide_doubles(1, 0)"```
**C**) ```libtbx.python -c "import scitbx.array_family.flex as f; print f.int([1,2,3])/0"```

under the following conditions: 
**1** = Developer environment
**2** = Deployment environment
**F** = `BOOST_ADAPTBX_FPE_DEFAULT` set
**S** = `BOOST_ADAPTBX_SIGNALS_DEFAULT` set

with the following results: 

**i** = prints "inf"
**E** = crashes with "Floating point exception"
**T** = crashes with stack trace

. | A | B | C
--- | --- | --- | ---
1  | T | T | T
1F  | i|  i|  T
1S   |E  |E  |E
1FS  |i  |i  |E
2    |i  |i  |T
2F   |i  |i  |T
2S   |i  |i  |E
2FS  |i  |i  |E

#### Observations:
* The crashes only happen on development builds. In fact there is currently no way to enable trace messages on non-development builds. This is counterintuitive, as on a development build we do not necessarily need trace messages as we could attach a debugger, whereas on a non-development build we can not attach a debugger, so this would be where traces are more useful.
* C++ code in flex arrays always crashes, no matter what flags are set.
* Python code that should simply return infinity does so everywhere but on development builds.

### Proposal
My proposal is therefore to invert the trap defaults. Replace the existing `BOOST_ADAPTBX_*` environment variables with 
* `BOOST_ADAPTBX_ENABLE_TRACE` to enable traces
* `BOOST_ADAPTBX_TRAP_FPE` to enable FPE crashes
* `BOOST_ADAPTBX_TRAP_OVERFLOW` to enable python overflow crashes

And with those conditions
**X** = `BOOST_ADAPTBX_ENABLE_TRACE` set
**Y** = `BOOST_ADAPTBX_TRAP_FPE` set
**Z** = `BOOST_ADAPTBX_TRAP_OVERFLOW` set
and
**3** = any environment (development or deployment)

I get the following results:

. | A | B | C
--- | --- | --- | ---
3   | i |  i  | E
3Y  | i | E  | E
3Z  | E |  i  | E
3YZ  | E | E | E
3X   | i | i  | T
3XY  | i | T |  T
3XZ  | T |  i | T
3XYZ | T | T | T

#### In other words:
* C++ code in flex arrays continues to crash (as this is independent of flags)
* Python code works consistently on development and deployment
* all code works out of the box on all occasions, and this includes 3rd party python modules that are not under our control
* With this it becomes possible for the first time to enable the traps on a deployment build (right now this is not possible, setting `*_FPE_DEFAULT` to "" or 0 is treated the same as 1, ie. it disables the flags)
* and to enable those flags separately.

To get the same behaviour as before all people have to do is to set the three flags in their environment, or use the context manager as outlined above to enable the flags selectively as and when needed.

### DIFF
--- a/boost/python.py
+++ b/boost/python.py
@@ -61,27 +61,18 @@ except AttributeError: pass # XXX backward compatibility 2009-11-24
 try: ostream = ext.ostream
 except AttributeError: pass
 
-if ("BOOST_ADAPTBX_SIGNALS_DEFAULT" not in os.environ):
+if os.getenv("BOOST_ADAPTBX_ENABLE_TRACE"):
   ext.enable_signals_backtrace_if_possible()
 
-
 class floating_point_exceptions_type(object):
-
   __shared_state = {'initialised': False}
 
-  def __init__(self, division_by_zero, invalid, overflow):
+  def __init__(self):
     self.__dict__ = self.__shared_state
     if not self.initialised:
-      if "BOOST_ADAPTBX_FPE_DEFAULT" in os.environ:
-        division_by_zero = self.division_by_zero_trapped
-        invalid = self.invalid_trapped
-        overflow = self.overflow_trapped
-      elif "BOOST_ADAPTBX_FE_DIVBYZERO_DEFAULT" in os.environ:
-        division_by_zero = self.division_by_zero_trapped
-      elif "BOOST_ADAPTBX_FE_INVALID_DEFAULT" in os.environ:
-        invalid = self.invalid_trapped
-      elif "BOOST_ADAPTBX_FE_OVERFLOW_DEFAULT" in os.environ:
-        overflow = self.overflow_trapped
+      division_by_zero = bool(os.getenv("BOOST_ADAPTBX_TRAP_FPE", self.division_by_zero_trapped))
+      invalid = bool(os.getenv("BOOST_ADAPTBX_TRAP_INVALID",self.invalid_trapped))
+      overflow = bool(os.getenv("BOOST_ADAPTBX_TRAP_OVERFLOW",self.overflow_trapped))
       ext.trap_exceptions(division_by_zero, invalid, overflow)
       self.initialised = True
 
@@ -118,15 +109,7 @@ class floating_point_exceptions_type(object):
     return locals()
   overflow_trapped = property(**overflow_trapped())
 
-def floating_point_exceptions():
-  import libtbx.load_env
-  if (libtbx.env.is_development_environment()):
-    flag = True
-  else:
-    flag = False
-  return floating_point_exceptions_type(
-    division_by_zero=flag, invalid=False, overflow=flag)
-floating_point_exceptions = floating_point_exceptions()
+floating_point_exceptions = floating_point_exceptions_type()
 
 
 class trapping(object):
--- a/boost_adaptbx/meta_ext.cpp
+++ b/boost_adaptbx/meta_ext.cpp
@@ -139,10 +139,14 @@ namespace {
 "                This crash may be due to a problem in any imported\n"
 "                Python module, including modules which are not part\n"
 "                of the cctbx project. To disable the traps leading\n"
-"                to this message, define these environment variables\n"
-"                (e.g. assign the value 1):\n"
-"                    BOOST_ADAPTBX_FPE_DEFAULT\n"
-"                    BOOST_ADAPTBX_SIGNALS_DEFAULT\n"
+"                to this message, undefine these environment variables\n"
+"                (or assign the value ""):\n"
+"                    BOOST_ADAPTBX_TRAP_FPE\n"
+"                    BOOST_ADAPTBX_TRAP_INVALID\n"
+"                    BOOST_ADAPTBX_TRAP_OVERFLOW\n"
+"                To disable the full trace but keep the exception you\n"
+"                can undefine:\n"
+"                    BOOST_ADAPTBX_ENABLE_TRACE\n"
 "                This will NOT solve the problem, just mask it, but\n"
 "                may allow you to proceed in case it is not critical.\n");
     fflush(stderr);


graeme-winter:
I would like this mostly because a huge number of third party libraries assume that this behaviour is correct (e.g. but not limited to matplotlib) and the current cctbx default is to redefine the language for some arcane reason which does not apply to the vast majority of users. 

Enabling this behaviour for debugging seems sensible - if people want the behaviour by default (i.e. not follow the Python language specification) then they can enable this. It would by default however allow most everyone else to get on with their work and have cctbx be a well behaved Python environment. 

I can see no reason not to merge this. If you want to keep having FPE's on perfectly working code, switch that behaviour on.

nksauter:
This must be discussed in the management meeting before any action is taken.
Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

graeme-winter:
Sure, though as it is rather low level should probably include e.g. lstbx/olex crew in conversation? I.e. this affects more than just dials and phenix

pafonine:
Not a good idea, IMHO.
Pavel

graeme-winter:
@pafonine a more verbose description would be useful. You have a use case where this is beneficial?

ndevenish:
This is a good idea.

As far as I'm concerned, the only times this has ever been useful, all it's done is preventing me from running a single additional command to get the stack trace. And 100% of the time this has been a problem, turning it off has been the solution, for which evidence I provide: The large amount of times it's been turned off.

Whereas there have been many, many times where I've had to disable it when some reasonable program or another has failed to run. Even `matplotlib` now causes this.

The rare `SIGV`s can be handled by a debugger. nan's are sticky. As long as you can turn this on explicitly when you want, when debugging, what harm does turning it off do? Versus annoying people when it stops their program doing perfectly legitimate IEE754 operations.

Has anyone an example of it been beneficial in a way that turning it on manually wouldn't have been?

luc-j-bourhis:
The key initial idea was to get numerical code fail as close as possible to the place where nan's and inf's are produced. The assumption is of course that it is always a bug when that happens. As far as the cctbx is concerned, this a valid assumption. Since the heavy numerical code is in C++, this behaviour of the C++ code shall be left unchanged. Your proposal does that iiuc, so far so good. I agree with your argument that the default setting of those flags keep causing issues with 3rd party libraries. We have the same problem at Bruker and we also disable some of those flags. So, I am inclined to support this proposition.

graeme-winter:
@pafonine did you get a chance to think some more about your objections to this? Would be handy to have an idea what the issues are.

rjgildea:
Has any consensus been reached on this pull request? I believe that the current BOOST FPE trapping behaviour is the cause of the current test failure of scitbx/math/tests/tst_math.py on mac/clang: https://github.com/cctbx/cctbx_project/issues/371

Anthchirp:
there were 3 1/2 in favour, 1 against, 1 for further discussion, but no concrete issues were identified and no further concerns raised here so this can probably be merged.

nksauter:
Please do not merge this until fully discussion between the Berkeley and
Diamond groups.

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

Anthchirp:
We can have the discussion here

graeme-winter:
As mentioned https://github.com/cctbx/cctbx_project/pull/324#issuecomment-487136060 above, happy to discuss but recommend inclusive call

ndevenish:
A group call might work, but it seems like _open_ discussion here about the reasons might be more productive - and document the process so we have something to point to if it comes up in the future.

phyy-nx:
Hi, all.  Spoilers, at the end of this wall of text, the pull request is approved from the LBL side :)

@Anthchirp one of your points in the new code is that

> C++ code in flex arrays continues to crash (as this is independent of flags)

And indeed, `libtbx.python -c "import scitbx.array_family.flex as f; print f.int([1,2,3])/0"` causes a floating point exception on the invert_boost_signal_default branch.

However, again on the invert_boost_signal_default branch, 
```libtbx.python -c "import scitbx.array_family.flex as f; print list(f.log(f.double([0,1,2,3])))"```
prints the output
`[-inf, 0.0, 0.6931471805599453, 1.0986122886681098]`
when I would have expected a crash.  This goes to @luc-j-bourhis's point, that generating nans and infs in the cctbx c++ is always a problem.  So on this branch, the bug @rjgildea found in scitbx.math (#371) wouldn't have been found.   Isn't that a problem?

Regardless of these specific points, we talked about this pull request around the room today.  Discussion included @phyy-nx, @nwmoriarty, @nksauter, @bkpoon, @pafonine, @olegsobolev, and @asmit3.  The general consensus is that the FPE traps have proven useful over and over again in the past and we don't want to disable them.  Especially in my case when running massively parallel jobs where normal testing doesn't uncover all the possible crashes.  Further, there was a design decision made back in the day that nans and infs are unsupported in cctbx, since in scientific code they are unrealistic.  cctbx is not a language, it's job is not to support all the idioms of the language it is written in.  It is a project to produce useful scientific results.

HOWEVER (big HOWEVER, many HOWEVERS)

We all agree that it is super irritating to not be able to use matplotlib on our development builds.  And more importantly, not being able to use new packages is a big concern, and one we can't easily find a solution to.

The example from previous discussion I believe was PyYAML.  Importing it caused a crash if boost.python was imported first.  However, more recently, I've found that numpy 1.16 causes the same crash.  A line in numpy 1.16's init files tests the limits of the current system's floating point operations and generates an inf.  Therefore, if numpy 1.16 is installed with boost, you can crash it with a FPE like this:

`libtbx.python -c "import boost.python; import numpy"`

To me, that's the ball game.  Yes, we could work with the numpy and yaml developers to get them to change their code but that seems counter productive.  We've gotta disable the FPE traps.  Therefore, after further discussion, we support this pull request.  Ok to merge.

HOWEVER HOWEVER (even more HOWEVERS)

We still want the FPE debugging in the core cctbx components.  So at LBL, after merging this pull request, we're going to put together a buildbot builder or an azure pipeline that runs only cctbx_regression.test_nightly (possibly a subset of these tests) with the FPE traps enabled.  Likely we'll need to turn off some tests that cause known problems, but just for this builder.  This is just to ensure new regressions in cctbx itself don't appear.

Thanks all!

graeme-winter:
@phyy-nx thanks for your efforts here 🙂

graeme-winter:
Given that this has been agreed will squash & merge

PR #326, opened 2019-05-01
fix mosaic domain initialization bug

jmholton:
hopefully without breaking Sauter features

### DIFF
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -383,6 +383,7 @@ nanoBragg::init_defaults()
     /* use these to remember "user" inputs */
     user_beam = false;
     user_distance =false;
+    user_mosdomains = false;
 
     /* scattering vectors */
 //    double incident[4] = {0,0,0,0};
@@ -2230,6 +2231,9 @@ nanoBragg::init_mosaicity()
     /* temporary local seed */
     long mseed;
 
+    /* catch nks initialized values here? */
+    if(user_mosdomains) return;
+
     /* free any previous allocations */
     if(mosaic_umats!=NULL) free(mosaic_umats);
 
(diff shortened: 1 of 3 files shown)


nksauter:
Well, this code change doesn't pass my tests!   How can I run the Dagewa tests?  Are they in cctbx_project or elsewhere?  Mine are in github/nksauter/LS49 and require private big data which I could easily make available to everyone concerned.

nksauter:
This pull-requested code does not work for me.  Could you give a little more detail here?  What exactly was the problem you were trying to solve, and how did the updated code accomplish this?  Why do you say there was a bug that needs to be fixed?

nksauter:
James--
Perhaps in the nanoBragg_ext.cpp, the set_mosaic_domains() function should have
```c++
nanoBragg.mosaic_domains = value
```
 after the init_steps()?

graeme-winter:
Re: testing

It occurs to me that if each module had a "what tests to run and how to run them" document work ike this would be much easier - frequently pull requests are caught out with "aha but you did not run _these_ tests" - across the board, not limited to any group or module. 

I appreciate that this is setting a new and higher bar for testing but could make life easier for collaborative coding?

dagewa:
> Well, this code change doesn't pass my tests! How can I run the Dagewa tests? Are they in cctbx_project or elsewhere? Mine are in github/nksauter/LS49 and require private big data which I could easily make available to everyone concerned.

@nksauter I am running simtbx as a user. While developing [my script](https://raw.githubusercontent.com/dials/dials_scratch/master/dgw/sim_ED_stills.py) I discovered a bug, namely that mosaicity could not be set in the intended way via attributes:

https://github.com/dials/dials_scratch/blob/c9a1bc16a3b54b54309a5e478b9ce5db6702f2ed/dgw/sim_ED_stills.py#L194-L195

I reported this to @jmholton, who fixed the bug for my use case. It is now clear that more work needs to be done to ensure all supported ways of setting the mosaicity work properly. I would be happy to work on a test to cover this, or turn my script into such a test, but it is not ready yet. I'm still exploring the use of simtbx here.

jmholton:
I believe I have cancelled this pull request and created a new one.

I edited Nick's test so that it initializes parameters in the right 
order, and added a new test for the more traditional functionality.

I could use some help in general figuring out if I'm setting up these 
tests right.

Cheers,

-James

PR #328, opened 2019-05-01
Advisory: Partially revert xfel messagepack changes

ndevenish:
**Note: Just an advisory pull request, not compulsory**

We've partially reverted the messagepack-as-default changes in [dials](https://github.com/dials/dials/commit/33fa5472920dbca6332b2391ea2951bf2a96b7c3) and [xia2](https://github.com/xia2/xia2/commit/082c6442f56cecc1ccd2933d169ee47a90aa4504) while we are reconsidering and handling a couple of related bugs relating to https://github.com/dials/dials/issues/770 that have stopped processing.

I believe this pull reverses, if you want, the parts of the cctbx/xfel code that was changed that is responsible for reading and writing only mpack files. I've attempted to leave in the parts of the code that handled both types. The idea is to re-revert these reversions once the underlying problem is fixed.

I don't know how urgently you need to use this with the latest dials so don't know how necessary this is. There appears to be no xfel tests so I can't check to see if these are working like we did with dials and xia2.

### DIFF
--- a/xfel/command_line/cspad_cbf_metrology.py
+++ b/xfel/command_line/cspad_cbf_metrology.py
@@ -268,7 +268,7 @@ def write_combine_phil(params, all_exp, all_ref):
 
 def refine(params, merged_scope, combine_phil):
   print "Combining experiments..."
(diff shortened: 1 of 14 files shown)


phyy-nx:
Thanks, I'll review today.

Again, the xfel tests are in xfel_regression and are run nightly at LBL.  We test cspad.cbf_metrology, PRIME, cxi.merge, and if psana is available, xtc_process and FormatXTC.  We also test some legacy code.  The cspad.cbf_metrology test includes all the joint metrology stuff we worked on for Brewster 2018.

phyy-nx:
Sounds reasonable.  Thanks for doing this.

PR #329, opened 2019-05-02
better fix for mosaic initialization bug

jmholton:
I added a test for the "normal" mosaic functionality and edited NKS's test to make it set the mosaic spread before setting the number of domains.  Doing the reverse order does not work because of parameter sanity updates.  Specifically, if the mosaic_spread_deg is zero, then the number of domains is always one.  This build passes the tst_gaussian_mosaic.py test.

It was actually a lot of effort to get these on-the-fly sanity checks to work, but I am now growing more convinced that I should trash this whole interface and start over.

In the interim, does anyone have a problem with these changes?  I'm not really all that sure I'm doing this right.


dagewa:
> It was actually a lot of effort to get these on-the-fly sanity checks to work, but I am now growing more convinced that I should trash this whole interface and start over.

I do find the current interface quite awkward. I would be happy with a higher-level factory method defined in Python in which everything is provided at construction time. Something like:

```python
nanoBragg.from_experiments_and_parameters(experiments, params)
```
where `experiments` is a dxtbx `ExperimentList` and params is a PHIL scope with documented definitions for everything that can be set for a `nanoBragg` instance. The factory method would take care of calling underlying setters in the appropriate order. There would be no need to trash the low level interface to provide this alternative - sometimes you may need the low level access even from Python, but I guess most cases would be satisfied with a high level interface like this.

jmholton:
Well, the difficulty is that the "proper order" depends on what you want 
to do.  These simulations can take such a long time to run that I 
decided sanity checks should be applied before the call to 
add_nanoBragg_spots().  Initially, I thought the boost interface would 
provide an user-friendly way to do this, but in the end I learned that 
it can't.  Things like when a user provides a SASE spectrum or edits a 
mosaic domain cannot be captured and managed. So it would seem, as you 
suggest, that I should abandon the object-oriented philosophy and revert 
back to a linear setup-test-run program flow.

Is there a general CCTBX philosophy for dealing with invalid or 
contradictory settings?  The mosaic_spread_deg vs mosaic_domains is just 
one example, but there are plenty of others where I'm trying to avoid 
dividing by zero, etc.  I suppose an example would be like in 
phenix.refine when you tell it to refine the occupancy of a certain side 
chain but then have another declaration that turns it off.

Also, can you point me to an example of a JSON-like interface?  I was 
told PHIL was being deprecated.

As for documentation, have the docstrings attached to every nanoBragg 
variable been helpful at all?

Cheers,

-James

jmholton:
OK.  Pushing the button...

dagewa:
I don't mean to say that the low level interface should be abandoned, just that simple use cases (like mine) could be handled cleanly with a high level interface that takes care of the details for the user.

I don't know of any general CCTBX philosophy for contradictory settings. A reasonable approach may be to emit a warning if a default setting is changed in order to accommodate something else the user has requested, and to fail with `Sorry` if the user explicitly requests incompatible settings. I doubt we actually follow this approach rigorously though. 

The docstrings are indeed useful, but for the user (and forgetful developer) the ability to run e.g. `dials.refine -c -a2 -e2|less` and immediately see the program's parameters with help strings and type declarations is very efficient.

PHIL has a few idiosyncrasies, but in general I find it very well suited to the task for which it was designed. I have not heard anything about it being deprecated. If that is true, I would be worried as DIALS and xia2 make heavy use of it. JSON serves a different purpose. Types, help strings and other attributes don't come for free and would have to be bolted on in some specification derived from JSON. No user wants to be typing bits of JSON manually in order to run a script.

dagewa:
Thanks for merging

dermen:
For what its worth, I think ConfigArgParse

https://github.com/bw2/ConfigArgParse

could be a fine alternative to phil for simtbx, where a parameter list is
handled only once and not really passed to other functions downstream..

PR #331, opened 2019-05-09
Avoid OSError when cctbx.python is run with uwsgi

gbunkoczi:
With uwsgi, sys.prefix seems to be the name of the virtual environment, and os.path.split return the empty string for dirname. os.samefile than raises and exception, since stat fails for dirname.


nwmoriarty:
Thanks. Hi from all of us here @ Berkeley.

PR #338, opened 2019-05-22
check which entries in a flex.vec3_double array are equal to a vec3_double

asmit3:
The comparison is done on an element by element basis in C++ and returns a flex bool array. Currently the == operator for flex.vec3_double returns a single bool value which is why this new function was implemented.

### DIFF
--- a/scitbx/array_family/boost_python/flex_vec3_double.cpp
+++ b/scitbx/array_family/boost_python/flex_vec3_double.cpp
@@ -184,6 +184,18 @@ namespace {
     return result;
   }
 
+  // Checks which elements in flex vec3_double is equal
+  // to a vec3_double and returns a flex.bool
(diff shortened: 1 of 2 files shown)


ndevenish:
> Currently the == operator for flex.vec3_double returns a single bool value

Is this not fixable?

asmit3:
@rjgildea I have added the __eq__ and __ne__ functions now with the new commits. Thanks for the input.

asmit3:
Overloading of the __eq__ and __ne__ operator cause other issues. For example the following code fails with a SCITBX_ASSERT error
```
x=flex.vec3_double((1,2,3))
[None, x].count(True)
```
This probably has to do with how None is handled in the C++ layer ? Regardless I am putting this branch on hold for now. Will try to get back if/when time permits.

PR #353, opened 2019-07-01
Fix npy detector format for Python 3

Anthchirp:
one more


Anthchirp:
agreed. Once we move to python 3 only we should remove the six dependency anyway.

PR #358, opened 2019-07-08
fix dxtbx deprecation warnings

Anthchirp:
requires recent dxtbx sources (after 12th of June) and a run of libtbx.configure.


Anthchirp:
It's been a month, so by now everyone should have update the sources. Unless anyone tells me otherwise I'll merge this next week.

PR #369, opened 2019-07-23
adding print_number_of_steps

JiriCernyJC:


ndevenish:
- `out` appears to not refer to any valid name - does this code work?
- `print` implicitly converts via `__str__`, I think
- I assume `”num_steps”` is always guaranteed to exist if the file loaded

PR #374, opened 2019-08-06
Gpu to master

dermen:
Requesting permission to bring nanoBraggCUDA.cu into master, further added time logger functionality from @RBolotovsky  and cuda device Id control for running GPU simulations multi GPU nodes. I just tested @nksauter  LS49 repository tests and the simtbx tests on new  CORI build and all tests passed.


dermen:
postponing, need to clean up and change into multiple pull requests

PR #376, opened 2019-08-07
Conda doc

dermen:
Updates to main project readme for help with conda based installation. Also links to examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Greetings, earthling. We see you intend to embark on the arduous quest of becomi
 
 * [Examples](#Examples)
 
-* [Additional information] (#additional)
+* [Additional information](#additional)
  
 
 <a name="intro"></a>
@@ -405,11 +405,11 @@ libtbx.configure newproject
 ```
 
 #### Compiling code during development
-If your new modules involve code that needs to be compiled, after configuring , go to the relevant build folder and run make 
+If your new modules involve code that needs to be compiled, go to the relevant build folder and run make 
 
 ```
 # after updating C code in one of the modules
-cd ~/crystal/build. # or whatever your build dir is named
+cd ~/crystal/build # or whatever your build dir is named
 make
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,32 @@
 # Computational Crystallography Toolbox
 
-## Introduction
+Greetings, earthling. We see you intend to embark on the arduous quest of becoming an elite computational crystallographer. Many before you have dared to try. We hope the CCTBX package grants you success along your journey.
```
graeme-winter: @dermen while I enjoy the D&D / Zork (or whatever, not sure I get the reference) theming I am not sure that the enjoyment would be universal? 😉 have experience of people adding "funnies" which later on causes more annoyance than it was worth (DIALS `HalRaiser` I am looking at _you_)

README.md
```diff
@@ -15,13 +41,378 @@ The cctbx is designed with an open and flexible architecture to promote extendab
...
+``` 
+
+Make ```bootstrap.py``` an executable and look at the options (we'll just use a few)
+
+```
+chmod +x bootstrap.py
+./bootstrap.py --help
+```
+
+Note, in the past bootstrap required that the first python binary in your path be python version 2, or else it will fail. Verify this by running
```
graeme-winter: Does `bootstrap.py` not run under Python 3? 🤔 @bkpoon I am pretty sure this was fixed up in the CCTBX Python 3 migration work?

graeme-winter:
@dermen thanks - adding some detailed instructions on getting started from conda is a clear improvement

bkpoon:
@graeme-winter These instructions were written a few months ago, so I'll update the text to describe the latest behavior and also make other general edits.

clacri:
Hi,  this is slightly related to this PR as it refers to the README but I can post it elsewhere. I noticed today that the CONTRIBUTING.md is not referred to in the README. Is this intentional, or would that be a desired improvement? (in which case I am happy to modify it and add a paragraph, and in this way, get more used to the cctbx contribution workflow!)

Issue #380, opened 2019-08-15
Converter method from vec3_double flex arrays to numpy arrays


m4tt-willi4ms:
As in the case of flex.double's `as_numpy_array()` method, it would be helpful to have a similar convenience method for vec3_double arrays. 
@rjgildea @luc-j-bourhis: will be submitting a pull request shortly; the implementation is much different (and likely less efficient) than the one for double arrays, but I thought I'd share nonetheless.

PR #381, opened 2019-08-15
Small change to flex.py for converting vec3_double arrays to numpy ar…

m4tt-willi4ms:
…rays
To address issue #380


phyy-nx:
Hi Matt, we've opened #384 as an alternate to this which has your test in a different place.  Look it over?  Thanks!

PR #383, opened 2019-08-28
Register scitbx flex arrays as random-seedable

Anthchirp:
Declares the random seeder in scitbx so that it can be controlled for testing.
(Feature mentioned in dials/dials#890)

Am I doing this right, @ndevenish?


Anthchirp:
Confirmed works, so this can be merged.

PR #384, opened 2019-08-28
Add flex.vec3_double().as_numpy_array()

### LINKED ISSUE #380: Converter method from vec3_double flex arrays to numpy arrays

phyy-nx:
Alternate version of #381 with test in tst_flex.py.  Fixes #380 and resolves #381.

Thanks Matt!

### DIFF
--- a/scitbx/array_family/flex.py
+++ b/scitbx/array_family/flex.py
@@ -576,3 +576,12 @@ def sum(flex_array, axis=None):
       slices[axis] = slice(i,i+1)
       flex_array_sum += flex_array[slices]
     return flex_array_sum
+
+def as_numpy_array(self):
+  """
+  A short extension method for converting vec3_double arrays to numpy arrays.
+  """
+  if isinstance(self, type(vec3_double())):
+    return self.as_double().as_numpy_array().reshape(-1, 3)
+
+vec3_double.as_numpy_array = as_numpy_array
--- a/scitbx/array_family/flex.py
+++ b/scitbx/array_family/flex.py
@@ -577,11 +577,11 @@ def sum(flex_array, axis=None):
       flex_array_sum += flex_array[slices]
     return flex_array_sum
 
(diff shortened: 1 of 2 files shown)


scitbx/array_family/flex.py
```diff
@@ -576,3 +576,12 @@ def sum(flex_array, axis=None):
       slices[axis] = slice(i,i+1)
       flex_array_sum += flex_array[slices]
     return flex_array_sum
+
+def as_numpy_array(self):
```
rjgildea:
```suggestion
def _vec3_double_as_numpy_array(flex_array):
```

1. Prefix the function name with an underscore to make it clear it's a private function
2. Rename to make clear that it's specific to `flex.vec3_double`
3. Change the argument from `self` to `flex_array`

scitbx/array_family/flex.py
```diff
@@ -576,3 +576,12 @@ def sum(flex_array, axis=None):
...
+  """
+  A short extension method for converting vec3_double arrays to numpy arrays.
+  """
+  if isinstance(self, type(vec3_double())):
+    return self.as_double().as_numpy_array().reshape(-1, 3)
+
+vec3_double.as_numpy_array = as_numpy_array
```
rjgildea:
```suggestion
vec3_double.as_numpy_array = _vec3_double_as_numpy_array
```
Rename as above.

Anthchirp:
This has been ready to be merged for a while. Will leave it to @m4tt-willi4ms and/or @phyy-nx to push the button (suggest squash merging)

PR #386, opened 2019-08-29
Remove boost.python.injector

Anthchirp:
This is known to not work in Python 3 and its continued availability
makes debugging harder than necessary.

Any places still refering to `boost.python.injector` should use
the `@boost.python.inject_into` decorator instead.

Merging this will break code that still uses the old method - that is by design.
Broken places can be fixed following the instructions [given in the file](https://github.com/cctbx/cctbx_project/blob/464c54b72525bb5a60a358662630336343412e57/boost/python.py#L187-L200)

### DIFF
--- a/boost/python.py
+++ b/boost/python.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
-import sys, os
+import os
 import re
+import sys
 from libtbx import cpp_function_name
 symbol_not_found_pat = re.compile(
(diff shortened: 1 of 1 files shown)


Anthchirp:
Although nobody commented on this I will change the approach here.

There is a nicer way of achieving the same aim by introducing `DeprecationWarning`s first. This will make it easier for developers relying on that functionality. I will commit that to master shortly, then update this pull request, and then we can remove the function at some point in the future.

PR #387, opened 2019-08-29
Write the libtbx eggs to the build directory

hattne:
Tell setuptools to put the libtbx eggs in the build directory and eliminate the chdir around the setup() invocation.  This allows libtbx.refresh to run on a custom Python where write-access to site-packages directory is restricted.

### DIFF
--- a/libtbx/pkg_utils.py
+++ b/libtbx/pkg_utils.py
@@ -235,6 +235,28 @@ def define_entry_points(epdict, **kwargs):
   finally:
     os.chdir(curdir)
 
+  # With the entry points installed into build/lib check for any legacy entry
+  # point definitions in both build/ and base/ and remove them where possible.
+  # Code block can be removed April 2020
+  try:
+    legacy_path = abs(libtbx.env.build_path / 'libtbx.{}.egg-info'.format(caller))
+    if os.path.exists(legacy_path):
+      print("Removing legacy entry points from", legacy_path)
+      import shutil
+      shutil.rmtree(legacy_path)
+  except Exception as e:
+    print("Could not remove legacy entry points:", str(e))
+  try:
+    import site
+    for spp in site.getsitepackages():
+      legacy_path = os.path.join(spp, 'libtbx.{}.egg-link'.format(caller))
+      if os.path.exists(legacy_path):
+        print("Removing legacy entry points from", legacy_path)
+        os.remove(legacy_path)
+  except Exception as e:
+    print("Could not remove legacy entry points:", str(e))
+
+
 def _merge_requirements(requirements, new_req):
   # type: (List[packaging.requirements.Requirement], packaging.requirements.Requirement) -> None
   """Merge a new requirement with a list.
--- a/libtbx/pkg_utils.py
+++ b/libtbx/pkg_utils.py
@@ -212,17 +212,16 @@ def define_entry_points(epdict, **kwargs):
   for ep in epdict:
     print("  {n} entries for entry point {ep}".format(ep=ep, n=len(epdict[ep])))
 
-  # Temporarily change to build/ directory. This is where a directory
-  # named libtbx.{caller}.egg-info will be created containing the
-  # entry point info. Create libtbx.{caller}.egg-link in
-  # {libtbx.env.build_path}/lib, which must be in sys.path.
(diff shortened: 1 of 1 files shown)


Anthchirp:
Tested this in a clean conda environment using the pull request #383 as a test case(*):
```bash
$ python bootstrap.py hot update base build --git-ssh --nproc=4 --use-conda 
$ . build/setpaths.sh 
$ cd modules/cctbx_project/
$ git checkout -b hattne-master master
$ git pull git://github.com/hattne/cctbx_project.git master
$ git merge origin/random
$ mkdir ../../wrong
$ cd ../../wrong
$ libtbx.configure ../modules/cctbx_project
$ cd ..
$ libtbx.python -c "import pkg_resources;print(list(pkg_resources.iter_entry_points('pytest_randomly.random_seeder')))"
```
and in a clean standard base build environment:
```bash
$ python bootstrap.py hot update base build --git-ssh --nproc=4
$ . build/setpaths.sh
$ cd modules/cctbx_project/
$ git checkout -b hattne-master master
$ git pull git://github.com/hattne/cctbx_project.git master
$ git merge origin/random
$ mkdir ../../wrong
$ cd ../../wrong
$ libtbx.configure ../modules/cctbx_project
$ cd ..
$ libtbx.python -c "import pkg_resources;print(list(pkg_resources.iter_entry_points('pytest_randomly.random_seeder')))"
```

While both versions print the expected output, the egg-info directories end up in `/wrong`, with the eg. `lib/libtbx.scitbx.egg-link` files pointing to that location.
This is not the case with the current code in `master`, where the `/wrong` directory is kept clean.

So it looks to me like this patch does not achieve the stated aims?


(*) I later noticed that this extra step was not necessary for verification, as `dxtbx` is configured by default and that already declares entry points. However it did allow me to confirm that PR #383 works as intended.

hattne:
OK, the previous iteration would only work if one refreshed in the build directory; looks like --script-dir does not change the destination of the egg.

Current iteration was tested in a conda environment and "standard base build"; .egg-info goes into build directory and .egg-link points to the right place.

Anthchirp:
Alas this doesn't work either.
Running the conda variant fails at the start of the `bootstrap.py build` step with:
```Updating entry points for dxtbx
  117 entries for entry point dxtbx.format
TEST FAILED: /scratch2/wra62962/files/temp/c387/conda/build/lib does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /scratch2/wra62962/files/temp/c387/conda/build/lib

and your PYTHONPATH environment variable currently contains:

    ''

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://setuptools.readthedocs.io/en/latest/easy_install.html#custom-installation-locations


Please make the appropriate changes for your system and try again.
Process failed with return code 1
```

Anthchirp:
Instructions I used, starting from just an empty directory:
```
wget https://raw.githubusercontent.com/cctbx/cctbx_project/master/libtbx/auto_build/bootstrap.py
python bootstrap.py hot update  --git-ssh --use-conda --nproc=4
cd modules/cctbx_project
git checkout -b hattne-master master
git pull git://github.com/hattne/cctbx_project.git master
cd ../..
python bootstrap.py base build  --git-ssh --use-conda --nproc=4
```

Anthchirp:
also fails with a normal base build without conda with the same error (just takes longer to get to that point)

hattne:
OK, now I'm confused.  I assume this is setuptools failing a sanity check–if build/lib won't eventually be in path, I've got a lot of understanding to do still.

hattne:
Next idea: how about using egg_info to write an egg to the build directory directly, and omit the egg-link entirely?

Anthchirp:
@hattne any comments on the change? Otherwise I will just merge in a few days.

Issue #291, opened 2019-02-19
Synchronize dependencies?


Anthchirp:
Good morning @bkpoon,

I just saw commit 478310afa787ce67df86a86fd466440fa0740d94 and wondered whether you are considering using these conda builds for the next phenix release?
The reason I am asking is that the dependencies therein differ from those currently declared in [package_defs](https://github.com/cctbx/cctbx_project/blob/master/libtbx/auto_build/package_defs.py). I don't mind having newer dependencies on conda, but would just like to point out that we only run tests on the base installation, so as far as we are concerned the newer dependencies are untested and DIALS can't really support them until they are declared in package.defs. So if you plan to distribute phenix conda builds including DIALS this might potentially lead to difficult to reproduce user support issues for us.

The most obvious one is of matplotlib, where we support version 2.0
https://github.com/cctbx/cctbx_project/blob/3b1b2d9e91624a118062096353a5dd677441a37d/libtbx/auto_build/package_defs.py#L124
yet the conda build now runs on 2.2.3
https://github.com/cctbx/cctbx_project/blob/478310afa787ce67df86a86fd466440fa0740d94/libtbx/auto_build/conda_envs/cctbx_py27_linux-64.txt#L117
but I also noticed a few others where conda has now outpaced the base installation:

package | base | conda
---------- | ------ | -----
numpy | 1.13.3 | 1.15.4
scipy | 1.0.0 | 1.2.1
scikit-learn | 0.19.1 | 0.20.2
pillow  | 4.2.1 | 5.4.1
biopython | 1.68 | 1.73
matplotlib | 2.0.0 (*) | 2.2.3
sphinx | 1.7.7 | 1.8.4
cython | 0.28.6 | 0.29.5
psutil | 5.4.8 | 5.5.1
tabulate | 0.8.2 | 0.8.3

(*) version is apparently 2.2 on Windows, #191

bkpoon:
Hi @Anthchirp 

Those updates are for a wider test of conda and of the dependencies. Once these versions have been validated, I can update the versions in `package_defs.py`. We've been running nightly builds for the dials and dials_regression tests with conda dependencies and I have one issue to sort out with macOS before these updated versions can be used generally. I would say that testing should be done in about 2 weeks.

bkpoon:
[...]

PR #389, opened 2019-09-10
Dependency updates

Anthchirp:
not very urgent at this time, but probably a good idea to consider these

* Cython 0.29
* HDF5 1.10.5
* OpenSSL 1.0.2s
* pytest 4.6.5 and pytest-xdist


Anthchirp:
just stumbled across this comment https://github.com/cctbx/cctbx_project/issues/291#issuecomment-470021918 so may need to downgrade cython again

Anthchirp:
builds passed, so I will merge this in a few days unless someone disagrees?

bkpoon:
Sounds good, looks like everything builds and no new failures.

PR #390, opened 2019-09-17
Speed up DIALS .mtz export

Anthchirp:
Instead of running O(n) calls to _add_batch(), which runs an O(n) check
and then runs another C library function running another O(n) check for
a total complexity of O(2n²)...

...run one custom function that inserts all batch information at the
same time for O(n) cost in time.

Relevant ticket: https://github.com/dials/dials/issues/924

### DIFF
--- a/iotbx/mtz/object.cpp
+++ b/iotbx/mtz/object.cpp
@@ -272,15 +272,12 @@ namespace iotbx { namespace mtz {
 
     CMtz::MTZBAT* p = ptr()->batch;
     CMtz::MTZBAT* p_tail = p;
-    int max_batch_number = 0;
-    int i_batch = 0;
-    int n_batches = image_range[1] - image_range[0] + 1;
-    for(;;i_batch++) {
-      if (p == 0) break;
-      max_batch_number = std::max(max_batch_number, p->num);
+    while(p != 0) {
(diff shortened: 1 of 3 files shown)


iotbx/mtz/object.cpp
```diff
@@ -246,6 +247,166 @@ namespace iotbx { namespace mtz {
     return result;
   }
 
+  void
+  object::add_dials_batches(int dataset_id,
```
rjgildea: Does the name _really_ need to have "dials" in?

Anthchirp: We thought this to be sensible as this function takes away a lot of the options for setting values and replaces them with DIALS-specific defaults.

Anthchirp: Also - if DIALS ever needs to write different things we'd want to be free to modify the function signature, so really don't want anyone else to rely on this function.

Anthchirp:
@rjgildea right now looking at moving the function into the dials repository instead (which would be a more appropriate place indeed)

iotbx/mtz/object.cpp
Anthchirp: cf https://github.com/dials/dials/pull/936

PR #391, opened 2019-09-18
Optionally avoid auto_anomalous check in iotbx.mtz.object.as_miller_arrays()

rjgildea:
- Add an `anomalous=None` parameter to `iotbx.mtz.object.as_miller_arrays()`.
- If this is set to `True` or `False`, then the call to `milller.set.auto_anomalous()` in `iotbx.mtz.column_group()` is avoided. This call can potentially be expensive with particularly large reflection files with millions of observations.
- In `iotbx.merging_statistics` pass the value of the anomalous parameter to
`iotbx.mtz.object.as_miller_arrays()`.
- Add anomalous parameters to other `as_miller_arrays()` methods

### DIFF
--- a/iotbx/cif/__init__.py
+++ b/iotbx/cif/__init__.py
@@ -113,7 +113,8 @@ class reader(object):
                        crystal_symmetry=None,
                        force_symmetry=False,
                        merge_equivalents=True,
-                       base_array_info=None):
+                       base_array_info=None,
+                       anomalous=None):
     if base_array_info is None:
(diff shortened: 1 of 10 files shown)


graeme-winter:
This is a useful addition as the calculations are not necessary in many of the cases we have.

graeme-winter:
I understand that this continues to cause problems with failing tests, some of which we have access to others which we do not. Propose that we revert this change set and instead copy the implementation to dials so that we don't have any unforeseen effects on phenix code.

Issue #195, opened 2018-05-15
Add pickling support for scitbx.sparse matrices


rjgildea:
Currently scitbx.sparse matrices are not pickleable, which means they can't be returned by libtbx.easy_mp.parallel_map. This would be useful in several parts of DIALS, e.g. dials.refine, dials.cosym, and scaling.

See e.g.:
https://github.com/dials/dials/blob/master/algorithms/refinement/refiner.py#L597
(multiprocessing currently disabled if SparseLevMar refinement engine selected)

https://github.com/dials/dials/blob/master/algorithms/symmetry/cosym/target.py#L193
(currently uses scipy.sparse as that is pickleable)

Also @jbeilstenedmands would like to use it to parallelise his scaling code.

dagewa:
+1 this would allow us to explore the use of multiprocessing non-linear least squares refinement for the XFEL case where the problem can be very sparse

jbeilstenedmands:
This would be very helpful for me, there are several places in my scaling code where I do calculations that return large sparse matrices, and being able to parallelise this would be very helpful for performance.

graeme-winter:
LBL folks - any objections to someone over this side taking a look at this?

PR #396, opened 2019-09-24
Add pickling support for sparse matrices. For #195

jbeilstenedmands:
This pull requests adds pickling support for sparse matrices, which allows them to be used in multiprocessing, for example in sparse refinement code.

The state of the sparse matrix is stored as a list of tuples of row_index, column_index, value. I'm unsure how to reliably test the speeds of various implementations, but this seems the obvious choice to me.

### DIFF
--- a/scitbx/sparse/boost_python/matrix.cpp
+++ b/scitbx/sparse/boost_python/matrix.cpp
@@ -117,6 +117,42 @@ struct matrix_wrapper
     return m.permute_rows(p);
   }
 
+  struct SparseMatrixPickleSuite : boost::python::pickle_suite {
+
+    typedef vector<T, copy_semantic_vector_container> column_type;
+    typedef typename column_type::const_iterator const_row_iterator;
+    typedef typename column_type::index_type index_type;
+    typedef T value_type;
+
+    static boost::python::tuple getinitargs(const matrix<T> &obj){
+      return boost::python::make_tuple(
+        obj.n_rows(),
+        obj.n_cols());
+    }
+
+    static boost::python::list getstate(const matrix<T> &obj){
+      boost::python::list state;
+      /// state is given by (row_index, col_index, value) for nonzero elements
+      for (index_type j=0; j < obj.n_cols(); j++) {
+        for(const_row_iterator p = obj.col(j).begin(); p != obj.col(j).end(); p++) {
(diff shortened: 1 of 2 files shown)


luc-j-bourhis:
Preliminary remark: don't forget to call the method 'compact' on the matrix before doing anything. Your plan is the so-called Coordinate List format, and it is pretty inefficient. If that matters, you should use the so-called Compressed Row Storage, described [here](http://netlib.org/linalg/html_templates/node91.html). Considering scitbx.sparse matrices are stored by columns, it might be better to adapt that to a Compressed Column Storage but I am not sure. Being stored by columns means that it is way more efficient to iterate over the elements of a column than over those of a row.

dagewa:
scitbx sparse matrices are stored as columns, where each column is a sparse vector (or equivalently, a `sparse.matrix_column`). The natural Python representation for a sparse vector is a dictionary of elt:val, from which it can be constructed
```python
v = sparse.vector(10, {1:3, 4:7})
```
while a sparse.matrix can be constructed via a list of such dictionaries 
```python
>>> sparse.matrix(10, 2, [dict(v), dict(v)])
sparse.matrix(rows=10, columns=2,
              elements_by_columns=[ { 1: 3, 4: 7 },
                                    { 1: 3, 4: 7 }, ])
```
I was thinking that we could lean on these facts to pickle both vectors and matrices. The data already have complete Python representations which are pickleable, as long as we keep track also of the dimensions of the vector or matrix.

luc-j-bourhis:
David, yes, you could do that but storing to and restoring from dictionaries would be even less efficient than the Coordinate List in that pull request. I did add support for dictionaries only because it made writing test code easier! But I don't know at all what is the intended use and whether efficiency matters, so take all my remarks with a grain of salt.

jbeilstenedmands:
Thanks for the feedback, I'll add a call to compact(), thanks for the reminder.

Regarding the storage, the compressed column storage would be a bit better, however I think we'd only see an improvement of up to a third in binary size/serialisation time (dependent on sparsity, dimensions etc). In getstate, I'm using the sparse vector iterator so only the nonzero elements are returned to be stored in the state list - so the only potential saving is in fewer stores of the column index. In setstate, I think it shouldn't make any difference whether the matrix elements are added individually or first constructed into vectors and then added as a list of vectors.

PR #400, opened 2019-10-02
Remove prehistoric backwards-compatibility

rjgildea:
These date back over 3 years to pre-github times, so removing them is highly
unlikely to cause any issues.

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -2621,64 +2621,6 @@ def unpickle():
   env = pickle.load(libtbx_env)
   if (env.python_version_major_minor != sys.version_info[:2]):
     env.raise_python_version_incompatible()
-  # XXX backward compatibility 2009-04-06
-  if (not hasattr(env.build_options, "boost_python_bool_int_strict")):
-    env.build_options.boost_python_bool_int_strict = True
-  # XXX backward compatibility 2009-04-27
-  if( not hasattr(env.build_options, "use_environment_flags") ):
-    env.build_options.use_environment_flags = False
-    env.build_options.env_cxxflags = ""
-    env.build_options.env_cflags = ""
-    env.build_options.env_cppflags = ""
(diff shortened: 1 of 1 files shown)


libtbx/env_config.py
```diff
@@ -2621,67 +2591,6 @@ def unpickle():
...
-  # XXX backward compatibility 2016-01-08
-  if not hasattr(env, 'explicitly_requested_modules'):
-    env.explicitly_requested_modules = set()
-  # XXX backward compatibility 2016-06-06
-  if not hasattr(env, 'extra_command_line_locations'):
-    env.extra_command_line_locations = []
-  # XXX backward compatibility 2017-11-03
-  if not hasattr(env.build_options, "python3warn"):
-    env.build_options.python3warn = 'none'
   # XXX backward compatibility 2018-12-10
```
phyy-nx: Why remove the soft boot backwards compatibility?  libtbx.configure uses this pickle file to save the options, and if you have an old build that you update, libtbx.refresh will now fail to run.  Granted, 2009 is pretty old, but 2016 isn't.

rjgildea: > Why remove the soft boot backwards compatibility? libtbx.configure uses this pickle file to save the options, and if you have an old build that you update, libtbx.refresh will now fail to run. Granted, 2009 is pretty old, but 2016 isn't.

cctbx moved to git/GitHub in November 2016 (judging by last commits at https://sourceforge.net/p/cctbx/code/25775/log/?path=/trunk). All the backwards compatibility options removed (with the exception of the `python3warn` option pre-date this move. I very much doubt it would be possible to bring a 2016-era libtbx environment up-to-date without a full bootstrap/re-build.

phyy-nx:
Gotcha.  That's reasonable.  Sounds good to me then.

Anthchirp:
The build failures on the PR look to be derived from the master branch. Merging.

Issue #395, opened 2019-09-23
Fix missing anomalous=None in as_miller_array functions


phyy-nx:
Several tests in phenix_regression are failing since #391 was merged.  These are the tests:

modules/phenix_regression/mmtbx/bulk_sol_and_scaling_fast/tst.py
modules/cctbx_project/cctbx/regression/tst_miller_statistics.py
modules/phenix_regression/iotbx/tst_french_wilson.py
modules/phenix_regression/mmtbx/tst_xmanip.csh
modules/phenix_regression/refinement/tst_089.py
modules/phenix_regression/iotbx/tst_reflection_file_reader.csh
modules/phenix_regression/misc/tst_hyss_etc.csh
modules/phenix_regression/refinement/autobuild/tst.py
modules/phenix_regression/gui/tst_hyss_gui.py
modules/phenix/phenix/regression/tst_xmanip.py

Fixing the tests generally entails adding anomalous=None to a few classes in iotbx implement the as_miller_arrays function.  This issue can be closed when that has been done.

FYI, to help catch these failures in the future:
1. phenix_regression and phenix_examples have been added to the cctbx_project Azure pipelines, so test modules/cctbx_project/cctbx/regression/tst_miller_statistics.py which expects phenix_examples now fails(before it was skipped) (see for example https://dev.azure.com/cctbx/cctbx_project/_build/results?buildId=756&view=logs).
2. The remaining tests in phenix_regression and phenix may be moved into iotbx after this week's phenix workshop.
3. Phenix itself will have an Azure pipeline set up at some point.  Its tests may be added as checks on cctbx branches like the cctbx_project pipeline checks that are present now, but it depends on how long the phenix builds take.  We're still figuring how many resources from Azure we can commit to each pipeline.

phyy-nx:
@rjgildea can you look into this?  If you can't get to it, can we revert #391 until the whole API can be patched in one go?  We need the API working for current development work.  Thanks for your attention.

rjgildea:
Sorry, I started looking at this the other day and then got distracted and forgot about it. I'll try to look at it again tomorrow.

phyy-nx:
Thanks!

phyy-nx:
Tests still failing:

- [ ]  modules/phenix_regression/misc/tst_hyss_etc.csh
- [ ]  modules/phenix_regression/iotbx/tst_iotbx_reflection_file.csh

rjgildea:
[...]

PR #401, opened 2019-10-02
Fix treatment of scale_range in hklf export

benjaminhwilliams:
A recent change (https://github.com/cctbx/cctbx_project/commit/4a7e1484d8db4c31d9d0189c7be0a9cf8cdb4e64) by @randyjread introduced a regression in the behaviour of `iotbx.shelx.hklf.miller_array_export_as_shelx_hklf`.  In introducing the `full_dynamic_range` argument, it seems he inadvertently deactivated the existing behaviour when `full_dynamic_range=False` and `scale_range` is not `None`.

This PR contains a simple fix that reinstates the option to run with `scale_range` set to a non-`None` value.  It also adds a test for that behaviour, to prevent future regressions, and also a basic test to exercise the `full_dynamic_range=True` behaviour introduced in https://github.com/cctbx/cctbx_project/commit/4a7e1484d8db4c31d9d0189c7be0a9cf8cdb4e64.

- Make it so that non-`None` values of `scale_range` are no longer ignored in `iotbx.shelx.hklf.miller_array_export_as_shelx_hklf`.
- Expose the argument `scale_range` of `iotbx.shelx.hklf.miller_array_export_as_shelx_hklf` in the method `export_as_shelx_hklf` of `miller_array` objects.
- Add a test to prevent regressions of behaviour when `scale_range` is not `None`.
- Add a test for existing behaviour of `export_as_shelx_hklf` with argument `full_dynamic_range=True`.

This also fixes https://github.com/xia2/xia2/issues/334.

### DIFF
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -4462,7 +4462,8 @@ class array(set):
   def export_as_shelx_hklf(self,
         file_object=None,
         normalise_if_format_overflow=False,
-        full_dynamic_range=False):
+        full_dynamic_range=False,
+        scale_range=None):
     """
     Write reflections to a SHELX-format .hkl file.
     """
@@ -4471,7 +4472,8 @@ class array(set):
       miller_array=self,
       file_object=file_object,
       normalise_if_format_overflow=normalise_if_format_overflow,
-      full_dynamic_range=full_dynamic_range)
+      full_dynamic_range=full_dynamic_range,
+      scale_range=scale_range)
 
   def export_as_cns_hkl(self,
         file_object,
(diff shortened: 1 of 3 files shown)


benjaminhwilliams:
There's currently an Azure Pipelines outage in the US (https://status.dev.azure.com/) which may account for all those cryptic failures of the PR check builds.  I'll re-trigger them when the outage is resolved, in the hope that they give more meaningful info.

benjaminhwilliams:
Having re-run the checks now that the Azure Pipelines system is back up and running, I can see that there is a single failing test causing all the check failures.  It seems that the failure is a known problem in master since the merge of #391, addressed in #395 & #402.  Since it is out of scope for this PR, is addressed elsewhere, and since the changes in this PR have @randyjread's approval, I'm going to merge this.

benjaminhwilliams:
Thanks for the review @randyjread.

PR #403, opened 2019-10-03
Add DIALS-specific conda environment

Anthchirp:
Following the lead of phenix dials now also includes conda environment definitions.

### DIFF
--- a/libtbx/auto_build/install_conda.py
+++ b/libtbx/auto_build/install_conda.py
@@ -156,7 +156,9 @@ class conda_manager(object):
                             platform=conda_platform[platform.system()])),
     'xfel': default_file,
     'labelit': default_file,
-    'dials': default_file,
+    'dials': os.path.join('dials', '.conda-envs',
+      default_format.format(builder='dials', version=version,
+                            platform=conda_platform[platform.system()])),
     'external': default_file,
     'molprobity': default_file,
     'qrefine': default_file,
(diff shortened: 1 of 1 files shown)


graeme-winter:
Corresponding conda environments added at https://github.com/dials/dials/compare/456d331e75ed...52b3fcd9828c

bkpoon:
Looks great! I do have a few comments,

- `matplotlib` pulls in `pyqt`
https://github.com/conda-forge/matplotlib-feedstock/blob/5af510232c2fa984e981e3eeb18128a6cab41b6d/recipe/meta.yaml#L69-L79
To get `matplotlib` without `pyqt`, use `matplotlib-base`
- Different channels are not always [compatible](https://conda-forge.org/docs/user/tipsandtricks.html#using-multiple-channels). And by not choosing a channel for everything, you will be affected by whatever channel priority is in the user's `~/.condarc` file. By default, the `defaults` channel takes priority. Compatibility is better with the newer packages, though. And I was able to run `pytest --regression` in `dials` where `conda_base` was constructed with the `defaults` channel having priority over the `conda-forge` channel.
- I'm assuming that for releases, you'll lock down specific versions and builds for reproducibility. You can actually backup those specific packages anywhere since the explicit version of the specification list is just a list of URLs and optionally the MD5 hash. I've been using other channels on [Anaconda.org](https://anaconda.org), but each channel has a 3 GB limit.
- And just a heads up, it looks like `orderedset` does not have a version for Python 3.7. There is a [pull request](https://github.com/simonpercivall/orderedset/pull/21) to update the metadata, so the code will probably work, it's just not tested and there is no Python 3.7 package on `conda-forge`.

Anthchirp:
Thank you. I will amend the matplotlib dependency, and read up on the channel stuff.

As for orderedset, since we only support up to 3.6 for the time being this should be fine. Given that the PR has been sitting for 5 months and there has not been any other movement in the package since it looks to be somewhat abandoned for the moment. I expect that in the long term we will get rid of orderedset again and replace it with a native implementation using dictionary keys again. Python 3.7+ (and CPython 3.6+) dictionary keys are guaranteed to be ordered.

PR #404, opened 2019-10-03
Fix necessary for compiling phaser_tng

Oeffner:
We have had this problem for more than a year and I think this change may solve the problem.
It fixes compiling phaser_tng outside of conda on Windows with the following command:

`python bootstrap.py --builder=phaser_tng`

Without it the following error is reported:

`
RuntimeError: Multiple sources for dispatcher:
  target file:
    "c:\users\phaserbuilder\phasertng2\build\bin\phaser_tng.build_dag_node"
  source files:
    "c:\users\phaserbuilder\phasertng2\modules\phaser_tng\phaser_tng\command_line\build_dag_node.py"
   =relocatable_path(anchor="c:\users\phaserbuilder\phasertng2\build", relocatable="..\modules\phaser_tng\phaser_tng/command_line\build_dag_node.py")
    "c:\users\phaserbuilder\phasertng2\modules\phaser_tng\phaser_tng\command_line\build_dag_node.py"
   =relocatable_path(anchor="c:\users\phaserbuilder\phasertng2\build", relocatable="..\modules\phaser_tng\phaser_tng\command_line\build_dag_node.py")
`

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -1316,7 +1316,6 @@ Wait for the command to finish, then try again.""" % vars())
       elif (source_file.isfile()
             and (   not hasattr(os.path, "samefile")
                  or not reg.samefile(source_file))):
-        #import code, traceback; code.interact(local=locals(), banner="".join( traceback.format_stack(limit=10) ) )
         raise RuntimeError("Multiple sources for dispatcher:\n"
           + "  target file:\n"
           + "    %s\n" % show_string(abs(target_file))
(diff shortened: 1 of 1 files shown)


libtbx/env_config.py
```diff
@@ -1316,6 +1316,7 @@ def write_dispatcher(self,
       elif (source_file.isfile()
             and (   not hasattr(os.path, "samefile")
                  or not reg.samefile(source_file))):
+        #import code, traceback; code.interact(local=locals(), banner="".join( traceback.format_stack(limit=10) ) )
```
phyy-nx: Errant comment?  Otherwise, looks good to me too.

graeme-winter:
Completely agree with @nksauter and @phyy-nx - if we could look for similar cases in the code base I think it would be _a good thing_ to remove any os-specific code.

Oeffner:
Thanks Aaron. I just removed the superfluous debug comment you pointed out. this means we can now finally build phaser_tng on Windows. I wonder if this bug was exposing a deeper problem in our code.

PR #408, opened 2019-10-25
bootstrap: migrate some CCI packages to GitHub

bkpoon:
Due to the potential for more power outages at Berkeley. The
following packages were migrated to the CCTBX organization
on GitHub.com.

- annlib
- ccp4io_adaptbx
- clipper
- eigen
- gui_resources
- tntbx

The existing ccp4io repository is used instead of the one at Berkeley.
The official SCons 3.0.0 package is used instead of the copy at Berkeley.

### DIFF
--- a/libtbx/auto_build/bootstrap.py
+++ b/libtbx/auto_build/bootstrap.py
@@ -691,21 +691,17 @@ class SourceModule(object):
 # On Windows due to absence of rsync we use pscp from the Putty programs.
 class ccp4io_module(SourceModule):
   module = 'ccp4io'
-  anonymous = ['curl', [
-    'http://cci.lbl.gov/repositories/ccp4io.gz',
-    'https://drive.google.com/uc?id=1EF6AqowSrVnse7pRtRmIsvhS6Q0dsSLT&export=download',
-  ]]
-  authentarfile = ['%(cciuser)s@cci.lbl.gov', 'ccp4io.tar.gz', '/net/cci/auto_build/repositories/ccp4io']
-  authenticated = ['rsync', '%(cciuser)s@cci.lbl.gov:/net/cci/auto_build/repositories/ccp4io/']
+  anonymous = ['git',
+               'git@github.com:cctbx/ccp4io.git',
+               'https://github.com/cctbx/ccp4io.git',
+               'https://github.com/cctbx/ccp4io/archive/master.zip']
 
 class annlib_module(SourceModule):
   module = 'annlib'
-  anonymous = ['curl', [
-    'http://cci.lbl.gov/repositories/annlib.gz',
-    'https://drive.google.com/uc?id=1YD_KDXrfhJ5ryT97j4yxmbAPoecGLjg0&export=download',
-  ]]
-  authentarfile = ['%(cciuser)s@cci.lbl.gov', 'annlib.tar.gz', '/net/cci/auto_build/repositories/annlib']
-  authenticated = ['rsync', '%(cciuser)s@cci.lbl.gov:/net/cci/auto_build/repositories/annlib/']
+  anonymous = ['git',
+               'git@github.com:cctbx/annlib.git',
+               'https://github.com/cctbx/annlib.git',
+               'https://github.com/cctbx/annlib/archive/master.zip']
 
 class scons_module(SourceModule):
   module = 'scons'
@@ -728,11 +724,6 @@ class afitt_class(SourceModule):
   authenticated = [
     'scp',
     '%(cciuser)s@cci.lbl.gov:/net/cci-filer2/raid1/auto_build/externals/'+afitt_version+'.gz']
-
-class libsvm_module(SourceModule):
-  module = 'libsvm'
-  anonymous = ['curl', 'http://cci.lbl.gov/repositories/libsvm.gz']
-  authenticated = ['rsync', '%(cciuser)s@cci.lbl.gov:/net/cci/auto_build/repositories/libsvm/']
 
 # Core CCTBX repositories
 # These must all provide anonymous access.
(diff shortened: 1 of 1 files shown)


graeme-winter:
@bkpoon +1 to the concept

Does this mean you can run bootstrap with LBL offline? Would assume so and big help if so. 

Also recognise that this continual power disruption must be a nightmare 😕

graeme-winter:
I had a quick look but asked @Anthchirp to take a look as he's done similar things for the DIALS release builds...

bkpoon:
Yes, this is desired solution for making bootstrap work when LBL is down. The Google Drive links were a stopgap solution since the svn server was down. The power is off at LBL now, so the Google Drive links will eventually take over if you're willing to wait for all the retry attempts. The Azure pipelines are working with those links and it takes a little more than 1.5 hours to download the `modules` directory (https://dev.azure.com/cctbx/cctbx_project/_build/results?buildId=1163)

bkpoon:
@Anthchirp I added the master branch from the DIALS repositories as `releases/lbl-tarball` so that the hashes from your versions exist in these repositories. Future DIALS branches can be created in the dials organization from the current `master` branches. An example can be found in this [fork of `tntbx`](https://github.com/bkpoon/tntbx) which has the DIALS branches and a `master` branch with the history.

I also recreated the `boost` repository in the `cctbx` organization. I do not remember the discussion about moving it to the dials organization. Also, it looks like it is now archived and read-only. Since `conda` is not the only way to provide the dependencies for `cctbx`, the option of building the subset of Boost libraries from source is still necessary. I may also switch the `boost` repository to use submodules from the official Boost repository.

Lastly, there are DIALS release branches in `boost` and `annlib_adaptbx`. Please make future DIALS release branches in the dials organization so that they are consistent with the other repositories. Thanks!

Anthchirp:
* We don't use hashes for anything, so the `releases/lbl-tarball` branches are quite unnecessary and can be deleted.
* By overwriting the repository forwards you broke the DIALS 1.1-1.3 bootstrap scripts, but given that these are now over 4 years old I don't think this matters much.
* Since DIALS 1.4 the DIALS bootstrap script uses the `dials/boost` repository, so you do not need to keep the DIALS release branches in `cctbx/boost`.
* `cctbx/annlib_adaptbx` is only used by the DIALS 2.2 and 3.0 bootstrap, so all other release branches can be removed from there. Please retain the 2.2 and 3.0 branches.
* Since @ndevenish has taken over responsibility for the DIALS release processes for versions 3.0+ tagging him here for information.

bkpoon:
It looks like `annlib`, `ccp4io`, and `scons` were also redirecting to the `dials` organization. I added `dials-1.1`, `dials-1.2`, and `dials-1.3` to the `annlib` and `ccp4io` repositories. The `scons` repository is not that important so that can still redirect to the `dials` organization. The error I am getting now for `dials-1.1` is for https://github.com/xia2/cctbx.git which redirects to the `cctbx` organization https://github.com/cctbx/cctbx_project.git.

Here is my command for getting the sources for `dials-1.1`.
```
wget https://raw.githubusercontent.com/dials/cctbx/dials-1.1/libtbx/auto_build/bootstrap.py
python2 bootstrap.py hot udpate --builder=dials
```
and here is the output
```
Performing actions: hot update
===== Running in .: deleting dist, tests, tmp
===== Downloading https://github.com/cctbx/annlib/archive/dials-1.1.zip: 537 kB
===== Installing modules/annlib-dials-1.1.zip into modules/annlib
===== Downloading https://github.com/cctbx/boost/archive/dials-1.1.zip: 17310 kB
===== Installing modules/boost-dials-1.1.zip into modules/boost
===== Downloading https://github.com/cctbx/scons/archive/dials-1.1.zip: 813 kB
===== Installing modules/scons-dials-1.1.zip into modules/scons
===== Downloading https://github.com/cctbx/ccp4io/archive/dials-1.1.zip: 1499 kB
===== Installing modules/ccp4io-dials-1.1.zip into modules/ccp4io
===== Downloading https://github.com/dials/cbflib/archive/dials-1.1.zip: 12015 kB
===== Installing modules/cbflib-dials-1.1.zip into modules/cbflib
===== Running in modules: git clone -b dials-1.1 https://github.com/xia2/cctbx.git cctbx_project
Cloning into 'cctbx_project'...
fatal: Remote branch dials-1.1 not found in upstream origin
Process failed with return code 128
```
And because of the redirect back to the `cctbx` organization, the cloning will fail because there is no `dials-1.1` branch. Why were all these redirects set up in the first place?

> By overwriting the repository forwards you broke the DIALS 1.1-1.3 bootstrap scripts, but given that these are now over 4 years old I don't think this matters much.

My intent is not to break earlier DIALS builders, so we can add the DIALS 1.1-1.3 branches to these dependency repositories (but not `cctbx_project`). But it looks like the `cctbx` repository in the `xia2` organization would need to be recreated. If you really don't want to support those versions of DIALS, I can also remove the DIALS 1.1-1.3 branches that I just added. Then I'll go ahead and remove the DIALS branches from `boost` and the DIALS branches (except for 2.2 and 3.0) from `annlib_adaptbx`.

Anthchirp:
I don't see a point in going out of our way to fix 1.1-1.3 bootstrap versions. Trying to avoid breaking stuff is one thing, but putting in effort to ensure they continue working is quite another. Right now the only supported versions of DIALS are 2.2 and 3.0, and anyone wanting to use other versions can either use the binary installers or fix up the bootstrap scripts themselves.

bkpoon:
Alright, I will merge this next week on 8/18 at around 11 am Pacific.

It does look like I can build `dials-1.2` with the branches now, but I will remove the DIALS release branches from all the repositories in the `cctbx` organization if they are not needed anymore. Just to confirm. the `dials-2.2` and `dials-3.0` branches in `annlib_adaptbx` are the only ones you want to keep?

Anthchirp:
I'm on leave and not in a position to verify this. I am confident that what I wrote was correct.

bkpoon:
@graeme-winter But 3.1 should not be in the `cctbx` organization.

@Anthchirp I just tried the instructions for DIALS 1.2 on CentOS 6
```
$ wget https://raw.githubusercontent.com/dials/cctbx/dials-1.2/libtbx/auto_build/bootstrap.py
$ python bootstrap.py --builder=dials hot update base build
```
And that finished successfully. As for DIALS 1.1, the error is with a redirect from the `xia2` organization since http://github.com/xia2/cctbx redirects back to `cctbx` organization. When I modify `bootstrap.py` to use the `cctbx` repository from the `dials` organization, all the sources are downloaded successfully. So the DIALS 1.1-1.3 bootstrap.py script behavior can be preserved even though it's not ideal to have DIALS release branches in the `cctbx` organization.

Issue #172, opened 2018-04-12
Update Boost to 1.65


Anthchirp:
This will involve fixing a number of deprecation warnings

Anthchirp:
Required work will include #70

Anthchirp:
guess this will be picked up eventually but is not a priority. closing.

PR #409, opened 2019-11-05
Update conda environments (Python 3.7 and 3.8)

### LINKED ISSUE #70: Boost 1.65: boost::python::numeric removed
As per http://www.boost.org/users/history/version_1_65_1.html:
> The boost::python::numeric API has been removed, as it is being obsoleted by boost::python::numpy.

This is used in (at least) `scitbx/.../numpy_bridge.hpp` and thus anything that depends on it. I tried literally just renaming the interface and providing compatibility macros (https://github.com/ndevenish/cctbx_project/commit/1d25c89dc9bfd0f2a2af961e294db9bbcc10897d was the very rough first attempt), and that wasn't sufficient, but didn't get time to delve any deeper.

More of an "in the future" problem since we don't use it at the moment - since this was just discussed I thought I'd add it here to track. I don't think this is the only issue with Boost 1.65.

The new numpy API was only added in Boost 1.63, so potentially moving to that as a minimum requirement first would mean that we don't have to conditionally adhere to both APIs, but I don't know how realistic that is.

bkpoon:
Test updated dependencies and add Python 3.7 (and maybe 3.8) support

## List of changes
1) Add `cctbxlite` environments (not yet enabled)
2) Add `pyside2`, `pyzmq` and `websocket-server` for @Oeffner to `cctbx` environments 
2) Versioned `conda` environments for linux, macOS, and Windows for Python 2.7, 3.6, 3.7, and 3.8
3) Updated CI build to build all versions of Python (12 builds)
a) Windows
b) CentOS 6
c) macOS 10.14 (Xcode 11.3.1)
4) Update Full build to build all versions of Python (40 builds)
a) Windows
b) CentOS 6, 7, 8
c) Ubuntu 14.04, 16.04, 18.04 (will replace 14.04 with 20.04 when it's released later in April)
d) macOS 10.14 Xcode 11.3.1 and 10.3, and macOS 10.15 Xcode 11.4
5) Update Boost to 1.70 (source build), closes #70 
a) Match current version for `conda`
b) Enables Python 3.7 and later support
c) Enables support for latest versions of CUDA
6) Bump macOS minimum deployment version from 10.7 to 10.9
a) Match version for `conda` packages
7) Add `--python` flag to `bootstrap.py` for selecting Python version
a) Just passes argument to `libtbx.install_conda`
8) `scons` moved into `conda` environment
9) Add `--no-boost-src` flag to `bootstrap.py` to prevent downloading of Boost into modules during the `update` step.
10) Add weekly builds of where `-ffast-math` (`/fp:fast` for Windows) is removed and `-O0`, `-O1`, `-O2`, and `-O3` (`/O1` and `/O2` for Windows) are tested on the Full build (160 builds).

@Oeffner Can you double check the Windows builds are working for you?
@luc-j-bourhis @pcxod @Oeffner Are there any changes that you would like to the `cctbxlite` environments? Currently, these are the dependencies in `cctbxlite`
```
biopython
cython
docutils
future
ipython
libsvm
libsvm_py
matplotlib-base
mrcfile
nomkl
numpy
numpydoc
openblas
pillow
pip
psutil
pyopengl
pytest
reportlab
six
sphinx
wxpython
python.app # [osx]
py2app  # [osx]
```
We can probably cut down the number of dependencies for `cctbxlite`.


Anthchirp:
I had a go at this and ran
```bash
# copy bootstrap from newpy branch
$ python bootstrap.py --python3 --nproc=4 --git-ssh hot update --use-conda
$ cd modules/cctbx_project ; git checkout newpy ; cd -
$ python bootstrap.py --python3 --nproc=4 --git-ssh base build --use-conda
$ . build/setpaths.sh 
$ mkdir tests ; cd tests
$ libtbx.run_tests_parallel module=scitbx nproc=4
```

<details>
  <summary>and I get a bunch of <tt>TypeError</tt>s</summary>

```
libtbx.python "newpy/modules/cctbx_project/scitbx/stl/tst_set.py" [FAIL] 0.1s
  Time:  0.10
  Return code: 1
  OKs: 0
  Standard error:
      File "newpy/modules/cctbx_project/scitbx/stl/tst_set.py", line 50, in <module>
        exercise()
      File "newpy/modules/cctbx_project/scitbx/stl/tst_set.py", line 46, in exercise
        exercise_stl_string()
      File "newpy/modules/cctbx_project/scitbx/stl/tst_set.py", line 39, in exercise_stl_string
        s = set.stl_string(["a", "b", "c"])
    TypeError: No registered converter was able to produce a C++ rvalue of type std::string from this Python object of type str
```
</details>

<details>
  <summary>and <tt>Boost.Python.ArgumentError</tt>s</summary>

```
libtbx.python "newpy/modules/cctbx_project/scitbx/math/tests/tst_error_propagation_matrix_inverse.py" [FAIL] 0.7s
  Time:  0.70
  Return code: 1
  OKs: 1
  Standard error:
      File "newpy/modules/cctbx_project/scitbx/math/tests/tst_error_propagation_matrix_inverse.py", line 202, in <module>
        test_B_matrix()
      File "newpy/modules/cctbx_project/scitbx/math/tests/tst_error_propagation_matrix_inverse.py", line 147, in test_B_matrix
        Bmat = create_Bmat()
      File "newpy/modules/cctbx_project/scitbx/math/tests/tst_error_propagation_matrix_inverse.py", line 43, in create_Bmat
        return matrix.sqr(Crystal(*vecs, space_group_symbol = "P 1").get_B())
    Boost.Python.ArgumentError: Python argument types in
        Crystal.__init__(Crystal, rec, rec, rec)
    did not match C++ signature:
        __init__(boost::python::api::object, scitbx::mat3<double> A, std::string space_group_symbol, bool reciprocal=True)
        __init__(boost::python::api::object, scitbx::mat3<double> A, cctbx::sgtbx::space_group space_group, bool reciprocal=True)
        __init__(boost::python::api::object, scitbx::vec3<double> real_space_a, scitbx::vec3<double> real_space_b, scitbx::vec3<double> real_space_c, std::string space_group_symbol)
        __init__(boost::python::api::object, scitbx::vec3<double> real_space_a, scitbx::vec3<double> real_space_b, scitbx::vec3<double> real_space_c, cctbx::sgtbx::space_group space_group)
        __init__(_object*, dxtbx::model::Crystal)
libtbx.python "newpy/modules/cctbx_project/scitbx/glmtbx/tests/tst.py" [FAIL] 0.2s
```
</details>
I expect this is due to us still not properly supporting boost 1.65+ (#70, #172, #184, ...)

Using scons from the conda-forge package works rather nicely. However I don't think the eigen library is picked up correctly.
<details>
  <summary>Although no warnings or errors are shown the <tt>scitbx.examples.bevington</tt> extension does not appear to be built</summary>

```
$ libtbx.python -c "import scitbx.examples.bevington"
  File "newpy/modules/cctbx_project/boost/python.py", line 31, in import_ext
    try: mod = __import__(name)
ModuleNotFoundError: No module named 'scitbx_examples_bevington_ext'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "newpy/modules/cctbx_project/scitbx/examples/bevington/__init__.py", line 4, in <module>
    ext = boost.python.import_ext("scitbx_examples_bevington_ext")
  File "newpy/modules/cctbx_project/boost/python.py", line 42, in import_ext
    + ["  "+p for p in sys.path]))
ImportError: __import__("scitbx_examples_bevington_ext"): No module named 'scitbx_examples_bevington_ext'
  sys.path:
   (...)
```
</details>
I could not find a failing test that covers this extension, however its absence does break a test in DIALS.

bkpoon:
Which operating system are you using? The `boost` conda package is built with gcc 7, so it cannot link properly with anything below gcc 5.1 due to an ABI change in gcc (https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html). If you're using CentOS 6 or 7, or Ubuntu 14.04, you have to replace `boost` and `boost-cpp` with the packages from the `cctbx-old-abi` channel. To make the change,

```
conda activate ./conda_base
conda remove boost boost-cpp
conda install -c cctbx-old-abi boost boost-cpp
conda deactivate
```
Then I would delete the `build` directory and build again. With these changes, the `scitbx` tests pass for me on CentOS 6. This is not incorporated into `bootstrap.py` yet, but it is done for Azure Pipelines (in the `newpy` branch).

https://github.com/cctbx/cctbx_project/blob/420f0287b667cb82d4071b9bd554ed6ecef7b7db/.azure-pipelines/unix-conda-build.yml#L67-L72

But, this may break any `conda` packages in your environment that depends on the `boost` or `boost-cpp` packages (as 

On macOS, `conda-forge` uses clang 9, so a more recent Xcode (like 10 or 11) is needed. On my macOS 10.14.6 system with Xcode 11.3.1, the `scitbx` tests pass.

Alternatively, you can also try removing the `boost` and `boost-cpp` conda packages and putting the regular Boost 1.70 source package into `modules` as `boost`. The `scitbx` tests will pass as well (I only tested on CentOS 6).

For the `conda` package for `cctbx`, I think it'll be a good idea to include all the tests just to make sure things are working. You can run the test with the current package from the `cctbx-dev` channel, but you'll have to point to the test in some other location.

```
conda create -n newpy_test -c cctbx-dev python=3.8 cctbx
conda activate newpy_test
python <modules directory>/cctbx_project/scitbx/math/tests/tst_error_propagation_matrix_inverse.py
```

Let me know if you're still getting errors.

As for the `eigen` issue, I haven't gotten around to updating the code to build with the `eigen` files from `conda`. But the problem is that the `SConscript` for building only looks for `eigen` in `modules/eigen`.

Anthchirp:
Yes, I tried building on RHEL 7. Just to double-check - you're not planning to drop CentOS/RHEL 7 support, right?

I also had a look at running with the conda compilers but could not get that to work at all. Just gave me plenty of undefined variables errors in `libtbx/SConscript` etc.

bkpoon:
That's right, we are continuing to support CentOS/RHEL 6 and 7.

The conda compiler for linux is just regular GCC 7, so you can try that. But, the `devtoolset-7` package does not accept the new ABI (https://bugzilla.redhat.com/show_bug.cgi?id=1546704), so you'll still have to use the packages from the `cctbx-old-abi` channel.

Oeffner:
I'll test this pull request today on Windows. But I have to throw a spanner in the works and ask if we could have an additional module, [websockets](https://websockets.readthedocs.io/en/stable/), installed as well. It is python3 only and needed for the HKLviewer as it is more mature and stable than the websockets_server module.

luc-j-bourhis:
@bkpoon @pcxod My answer shall be two-fold. As a basic cctbx, useful for demos, and basic stuff, I think most of the current dependencies make sense. But there are a couple I don't know about:
```
libsvm
libsvm_py
mrcfile
nomkl
```

Then `biopython` will only be useful for macromolecular crystallographers but with conda, the bar is much lower than before in term of ease of install, so I'd say why not!

Then there is the question of what Olex2 needs. Oleg cooks his own sauce at the moment, so the point might be moot actually, unless we decide to take advantage of the conda install in the future. Anyway, for the record, please correct me Oleg, but the following aren't needed for Olex2 I don't think
```
biopython
ipython
matplotlib-base
pillow
pip
psutil
pyopengl
pytest
reportlab
six
wxpython
python.app # [osx]
py2app  # [osx]
```

libtbx/auto_build/bootstrap.py
```diff
@@ -775,7 +775,7 @@ class geostd_module(SourceModule):
 
 class boost_module(SourceModule):
   module = 'boost'
-  anonymous = ['git',
+  anonymous = ['git', '-b 1.70',
```
Anthchirp, line 778: Would suggest updating the repository on merge, rather than explicitly pointing to a branch

libtbx/auto_build/bootstrap.py
```diff
@@ -2653,9 +2672,16 @@ def run(root=None):
...
+  python_args.add_argument('--python',
+                    default='27', type=str, nargs='?', const='27',
+                    choices=['27', '36', '37', '38'],
+                    help="""When set, a specific Python version of the
+conda environment will be used. This only affects environments selected with
+the --builder flag. This cannot be used with --python3.""")
+  python_args.add_argument("--python3",
                     dest="python3",
-                    help="Install a Python3 interpreter. This is unsupported and purely for development purposes.",
+                    help="Install a Python3 interpreter. This is unsupported and purely for development purposes. This cannot be used with --python.",
```
Anthchirp, line 2684:
```suggestion
                    help=argparse.SUPPRESS,
```
I don't think there is a need to advertise this option any more. The help string is incorrect in any case (it is supported).
I think this option can either be hidden or even be removed entirely.

bkpoon:
@luc-j-bourhis So for dependencies, `libsvm` is a support vector machine library that is used for identifying ions in `mmtbx` and `libsvm_py` is the Python interface to that library. The Python interface is in the `cctbx` channel only and I haven't had time to contact the `libsvm` package managers to incorporate the Python interface. `mrcfile` is for opening cryo-EM maps and it's something you don't need. And `nomkl` is a way for selecting `openblas` instead of `mkl` as the basic math library.

As for the other dependencies, when the `cctbxlite` builder was added, it still used the dependencies for the `cctbx` builder. This is an opportunity to strip away anything that you, @pcxod, and @Oeffner might not want, so `cctbxlite` could be the minimal amount of dependencies for building.

Longer term, and this doesn't need to be resolved in this pull request, is that `conda` lets developers and users have finer-grained control of the dependencies. We can have runtime environments and developer environments. The starting point for building anything is a standard `conda` environment (path can be the argument to the `--use-conda` parameter in `bootstrap.py`). In the standard environment files I construct, the dependencies are from `conda-forge` and the version and build numbers are baked in for reproducibility, but any channel can be used as the source of the dependencies. And once the `conda-forge` package is available, an environment with the latest versions can be created by
```
conda create -n new_env -c conda-forge --only-deps cctbx
```
By having a minimal set of dependencies for building, other developers can add their dependencies to the minimal environment for their projects. For example, one way `mrcfile` and other dependencies can be added to the environment above is by defining a standard `conda` YAML environment file and running
```
conda install -n new_env --file more_dependencies.yml
```
There can be multiple `--file` arguments, so for Phenix, I can have a separate YAML file for each submodule that is not a separate `conda` package, and construct the final environment in pieces. Alternatively, I can have a Phenix `conda` package and the `conda create` command can create a fresh environment for Phenix builds.

bkpoon:
@Oeffner Sure, we can use websockets. So to confirm, the Python 2 environment will still have websocket-server, and the Python 3 environments will only have websockets instead of websocket-server? I can make the changes next week.

libtbx/auto_build/bootstrap.py
bkpoon: Sure, I can remove the option. How about the wxpython4 one?

libtbx/auto_build/bootstrap.py
bkpoon: That works, but once this pull request is approved, we should decide on a date to merge and to advertise that date so that developers are not caught off guard about this change. People will probably have to rebuild completely or delete their build/lib/*boost* files.

libtbx/auto_build/bootstrap.py
Anthchirp: wxpython4 is implied by python 3 environments, so the flag is not needed there.

However the flag is also automatically set on python 2 when running on MacOS 10.14+.
Removing the option entirely may break those platforms. (I'm fine with it being removed as we no longer support those cases on master)

Anthchirp: having said that - the command line option of course is not needed. So we could instead move the MacOS check to the point where the wxpython 3 or 4 decision is made.

Oeffner:
Yes that that makes sense. HKLviewer has no need for websocket-server in python3. Applies to the branch I'm working in which will soon be merged in to master.

Oeffner:
Hi @bkpoon I just build this branch on my Windows10 workstation with the commandline
`python3 bootstrap.py --builder=cctbx --nproc=10 --python3 --use_conda` and it builds fine. But the cctbx.HKLviewer doesn't as no stub file is created. So I added a change to bootstrap.py so that cctbx.HKLviewer works with a plain cctbx build. Committed in e520f1902c43d3a104ca2a75f067a3a4c74a9eb8

bkpoon:
@Oeffner I don't normally download a completely new set of repositories. For this branch, since it changes dependencies, I'll make a symbolic link to an existing `modules` directory, check out the `newpy` branch in `cctbx_project` and the `1.70` branch in `boost`,and then run `bootstrap.py` with the `base` and `build` options.
```
python bootstrap.py base build --use-conda <extra options>
```

pcxod:
Hi @bkpoon , @luc-j-bourhis. I do not use conda for building cctbx for Olex2 - I rely on the old fashioned way that seems to have been working for a while bow. After getting the bundle (which is apparently missing on this page https://cctbx.github.io/installation.html#id3, but as I remember last time I got it with bootstrap), I typically use 'configure iotbx smtbx fast_linalg' and have to leave with the dependencies that come along them. But these are as close as a minimal set.

Oeffner:
Not sure if this is the right forum but in porting python modules in our phaser repo to python3 I have encountered problems with scitbx.suffixtree module. The regression test, cctbx_project\scitbx\suffixtree\test\tst_single.py,  fails in a python 3 build with lots of errors like `OverflowError: Python int too large to convert to C long` . The test passes fine in a python2 build. If anyone has an idea of how to fix this let me know.

bkpoon:
@Oeffner It's been failing on Windows, but shows up as a warning because of the stderr output. I looked briefly into it a while ago, but wasn't able find a fix. Also, the test is marked as an expected failure for newer versions of GCC. There is a segmentation fault. @gbunkoczi would know more about that code.

Anthchirp:
I raised a question in our group meeting today: The next DIALS release is coming up in May. For planning purposes only - do you have a date in mind when you plan to merge this?

bkpoon:
@Anthchirp  How about next Friday, May 8th at 3 pm Pacific? I'll make the changes to just use the `master` branch of `boost` before merging.

I made the changes to remove the `--python3` and `--wxpython4` command-line options for `bootstrap.py`, but `install_base_packages.py` should still be able to build Python 3.7 and install wxPython 4. And I added a flag to not download the Boost source package. It's `--no-boost-src`, but the flag name can be changed. This flag will be useful for building `conda` packages where `boost` will come from `conda-forge`.

@pcxod @luc-j-bourhis I'll try to construct a minimal set of dependencies. It may just be `boost`, `numpy`, `six`, and `future`. Maybe `matplotlib-base` for plotting. It'll be useful for building a `cctbx-base` `conda` package that is barebones and then the regular `cctbx` package with all the other useful options for general users.

Anthchirp:
The suggested timestamp works well and leaves us plenty of time to iron out any potential issues before the release. Thank you!

PR #416, opened 2019-11-20
libtbx: fix for Windows stdout redirection

bkpoon:
- New Windows console in Python 3.6 and greater does not work with os.dup2
- https://bugs.python.org/issue30555
- https://github.com/python/cpython/pull/1927
- https://stackoverflow.com/questions/52373180/python-on-windows-handle-invalid-when-redirecting-stdout-writing-to-file


bkpoon:
@Oeffner Here the changes to get cctbx to compile on Windows for Python 3. The header changes just picks an earlier version of msinttypes. The stdint.h from Microsoft seems to be still causing issues with VC++ 14.0.

bkpoon:
Rolled into #409

PR #423, opened 2019-11-27
Update SCons to 3.0.5

Anthchirp:
We are running into build issues with SCons on Python 3. SCons 3.0 doesn't properly support Python 3, and a lot of fixes have been made to that version over the past 2 years, including one we need.
This updates SCons to the most recent 3.0 release by checking out the version directly from source, as suggested in https://github.com/cctbx/cctbx_project/pull/408#pullrequestreview-308331401

(This is currently holding up build-related work)

### DIFF
--- a/libtbx/auto_build/bootstrap.py
+++ b/libtbx/auto_build/bootstrap.py
@@ -705,12 +705,10 @@ class annlib_module(SourceModule):
 
 class scons_module(SourceModule):
   module = 'scons'
-  anonymous = ['curl', [
-    'http://cci.lbl.gov/repositories/scons.gz',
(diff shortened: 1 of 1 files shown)


Anthchirp:
On the other hand the repository is faster to update (`git checkout 3.1.1`) :smiley: 
I'm not too invested in it though, so am making archive download the default.

Regarding the error: I get some non-helpful `TypeError` in a parsing module, no trace though, so can't pinpoint exactly what is going on. I'll run a build with 3.1.1 to see how that goes.

Anthchirp:
build passes for me, so happy to merge this

bkpoon:
Looks good!

PR #425, opened 2019-12-11
Longstanding bugfix for cluster.unit_cell

phyy-nx:
Use radians, not degrees.  Caught by @nksauter.

The dials.cosym tests still pass.  Looking at them, I don't think they are clustering differing unit cells. 
 test_cluster_unit_cell.py also still passes.  Looking at the output, the dendrogram produced by test_cluster_unit_cell.py is the same along x, meaning the same cells are associated with each other, but different along Y, meaning their relative scores are different.

If I run cluster.unit_cell on the 757 integration results in xfel_regression/prime_test_data/myo/integration, I get different numbers of singletons, so this change will affect scientific results.

@rjgildea, this is likely relevant to your interests.  Let me know if you want me to hold this.  Otherwise, I'll merge it tomorrow morning.

### DIFF
--- a/xfel/clustering/singleframe.py
+++ b/xfel/clustering/singleframe.py
@@ -307,9 +307,9 @@ class SingleFrame(InputFrame):
       a = uc[0] ** 2
       b = uc[1] ** 2
       c = uc[2] ** 2
-      d = 2 * uc[1] * uc[2] * math.cos(uc[3])
-      e = 2 * uc[0] * uc[2] * math.cos(uc[4])
-      f = 2 * uc[0] * uc[1] * math.cos(uc[5])
+      d = 2 * uc[1] * uc[2] * math.cos((math.pi/180.)*uc[3])
+      e = 2 * uc[0] * uc[2] * math.cos((math.pi/180.)*uc[4])
+      f = 2 * uc[0] * uc[1] * math.cos((math.pi/180.)*uc[5])
       return [a, b, c, d, e, f]
 
 class SingleDialsFrame(SingleFrame):
--- a/xfel/clustering/singleframe.py
+++ b/xfel/clustering/singleframe.py
@@ -303,13 +303,18 @@ class SingleFrame(InputFrame):
 
   @staticmethod
   def make_g6(uc):
-      """ Take a reduced Niggli Cell, and turn it into the G6 representation """
+      """ Take a reduced Niggli Cell, and turn it into the G6 representation. This is
+          similar but not identical to the metrical matrix.  See
+          doi:10.1107/S0567739473001063 Gruber (1973)
+          doi:10.1107/S0108767388006427 Andrews and Bernstein (1988)
+          doi:10.1107/S1600576713031002 Andrews and Bernstein (2014)
+      """
       a = uc[0] ** 2
       b = uc[1] ** 2
       c = uc[2] ** 2
-      d = 2 * uc[1] * uc[2] * math.cos((math.pi/180.)*uc[3])
-      e = 2 * uc[0] * uc[2] * math.cos((math.pi/180.)*uc[4])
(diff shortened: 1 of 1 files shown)


rjgildea:
Good spot. Perhaps this could have been avoided in the first place by using a `uctbx.unit_cell` as this function (incorrectly) duplicated the functionality of `ucbtbx.unit_cell.metrical_matrix()`. Could this whole function just not be replaced with e.g.:
`return uctbx.unit_cell(uc).metrical_matrix()`

In fact, the metrical matrix, a.k.a. "g6" matrix is already cached as an attribute of the image object, so the existing code was incorrectly recalculating something it already had access to!

https://github.com/cctbx/cctbx_project/blob/master/xfel/clustering/singleframe.py#L336

https://github.com/cctbx/cctbx_project/blob/master/xfel/clustering/singleframe.py#L360

nksauter:
Richard,

Your statements are not correct.  G6 is different from the metrical matrix
and it is defined specifically in Herbert Bernstein's paper,
doi:10.1107/S1600576713031002,
equation (2).

Nick

Nicholas K. Sauter, Ph. D.
Berkeley, CA 94720
(510) 486-5713

rjgildea:
I see - there is an extra factor of 2 in the off-diagonal components compared to the metrical matrix.

rjgildea:
Nevertheless this _could_ still be calculated _via_ the metrical matrix, as in the cctbx implementation of the fast minimum reduction:
https://github.com/cctbx/cctbx_project/blob/fcc5184b86bf39b0af639c5ec8fd4f3445311e8a/cctbx/uctbx/fast_minimum_reduction.h#L78-L84
https://github.com/cctbx/cctbx_project/blob/fcc5184b86bf39b0af639c5ec8fd4f3445311e8a/cctbx/uctbx/gruber_1973.py#L119-L125

phyy-nx:
Hi, thanks.  I just tested the metrical matrix version just now and it's a
bit slower, possibly because I also would have to create a unit cell object
where make_G6 doesn't take one as an argument.  I'm happy with this
implementation as is.

xfel/clustering/singleframe.py
```diff
@@ -307,9 +307,9 @@ def make_g6(uc):
       a = uc[0] ** 2
       b = uc[1] ** 2
       c = uc[2] ** 2
-      d = 2 * uc[1] * uc[2] * math.cos(uc[3])
-      e = 2 * uc[0] * uc[2] * math.cos(uc[4])
-      f = 2 * uc[0] * uc[1] * math.cos(uc[5])
+      d = 2 * uc[1] * uc[2] * math.cos((math.pi/180.)*uc[3])
```
rjgildea:
```suggestion
      d = 2 * uc[1] * uc[2] * math.cos((math.radians(uc[3]))
```

xfel/clustering/singleframe.py
```diff
@@ -307,9 +307,9 @@ def make_g6(uc):
...
+      e = 2 * uc[0] * uc[2] * math.cos((math.pi/180.)*uc[4])
```
rjgildea:
```suggestion
      e = 2 * uc[0] * uc[2] * math.cos((math.radians(uc[4]))
```

xfel/clustering/singleframe.py
```diff
@@ -307,9 +307,9 @@ def make_g6(uc):
...
+      f = 2 * uc[0] * uc[1] * math.cos((math.pi/180.)*uc[5])
```
rjgildea:
```suggestion
      f = 2 * uc[0] * uc[1] * math.cos((math.radians(uc[5]))
```

phyy-nx:
Thanks, I added the suggestions.

PR #428, opened 2019-12-20
Strip trailing directory separators from module names.

### LINKED ISSUE #427: libtbx.configure breaks on module names with trailing directory separators
I have fallen into this trap:
```
libtbx.configure dials_scratch/
```
which ultimately fails with an obscure error
```
Processing: "/home/fcx32934/sw/cctbx/modules/dials_scratch/libtbx_refresh.py"
Updating entry points for 
error: Invalid distribution name or version syntax: libtbx.-0.0.0
```
This kind of situation is easy to get into, if for example using tab-complete to fill in the module name from its directory.

Fix incoming to strip trailing directory separators from module names

dagewa:
This seems to be a suitable place to strip possible trailing directory separators from module names passed to `libtbx.configure`.

Fixes #427.

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -789,6 +789,8 @@ Wait for the command to finish, then try again.""" % vars())
             '  resulting target = "%s"' % (
               module_name, redirection, dist_path))
         self.command_line_redirections[module_name] = dist_path
+      # remove any trailing dir separators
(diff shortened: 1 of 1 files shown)


dagewa:
Thanks. I did think about that, but then I remembered that Windows seems to accept either direction of slash in some circumstances and I thought better catch them both at all times.

PR #431, opened 2020-01-14
Importing Carbon.Appearance fails

graeme-winter:
On wx 4.0.6 at least, which I am currently using, there is no Carbon.Appearance so work around - does not seem to have a measurable impact.

Maniftested in dials/dials#1091 which this commit will fix


ndevenish:
`Carbon` is a python2 interface to an obsolete OSX API: https://docs.python.org/2/library/carbon.html - so this won't be doing anything at all very shortly

PR #432, opened 2020-01-23
Correct base pair definitions in bondlength_defaults.py

keitaroyam:
Only single hydrogen bonds were defined for symmetric homo purine/pyrimidine pairs.


keitaroyam:
@olegsobolev Can I merge these changes? I guess they were missed when transferring data from mmtbx/secondary_structure/base_pairing.data

olegsobolev:
Sorry for the delay. 
If I understand your PR correctly, you want to define "mirror" information for some bond lengths, e.g. to define them not only for N6-N1, but also for N1-N6. If so, this looks like duplicating information. Did you take a look at 
`def get_basepair_target_distance(atom1, atom2, which_basepair)`
just below this dictionary? This function is meant to take care of this kind of situation instead of increasing the number of constants in the dictionary. Although this function is not used in production, the dictionary itself is used. If this function does not provide the information you want or you have different inputs, I would rather add another wrapper function instead of changing the dictionary. 
Here is some example code:
```
import iotbx.pdb
from mmtbx.monomer_library.bondlength_defaults import get_basepair_target_distance

h = iotbx.pdb.input("6gpi.pdb").construct_hierarchy()
asc = h.atom_selection_cache()

atom1 = h.select(asc.selection("name N1")).atoms()[0]
atom2 = h.select(asc.selection("name N6")).atoms()[0]
print atom1.id_str()
print atom2.id_str()
print get_basepair_target_distance(atom1, atom2, 1)
print get_basepair_target_distance(atom2, atom1, 1)
```

Output:
```
model="   1" pdb=" N1  F7H A   1 "
model="   1" pdb=" N6  TCY A   9 "
(2.99, 0.14, 0.07)
(2.99, 0.14, 0.07)
```

keitaroyam:
@olegsobolev Thank your for the reply. I meant the symmetric homo purine/pyrimidine pairs (classes I-IV, XII-XV) should have two hydrogen bonds for each, but only one is defined with the current code. It is duplicated information as you pointed out although this is an easy fix, so I will look for another place to be fixed.

Issue #434, opened 2020-01-28
missing scipy dependency


Anthchirp:
Since approximately https://github.com/cctbx/cctbx_project/commit/0bec3ec293aeb66efc8e67a37c0ebe8a6ea88175 iotbx and cctbx depend on scipy. However scipy is not installed as a dependency in install_base_packages.py (unless one uses the dials installer, https://github.com/cctbx/cctbx_project/blob/master/libtbx/auto_build/install_base_packages.py#L279-L282)

pafonine:
Are we absolutely sure we need to bring in the entire package (scipy) 
just to do interpolation? I doubt..


/Users/pafonine/phenix/modules/cctbx_project/cctbx % libtbx.find_files 
-g scipy
./xray/structure_factors/from_scatterers_direct_parallel.py:     1. 
numpy package (http://numpy.scipy.org) version 1.5.1 OK
/Users/pafonine/phenix/modules/cctbx_project/cctbx % cd ../mmtbx/
/Users/pafonine/phenix/modules/cctbx_project/mmtbx % libtbx.find_files 
-g scipy
./validation/rama_z.py: from scipy import interpolate
./validation/grid2d.py: from scipy import interpolate

olegsobolev:
I'll work to get rid of it soon.

olegsobolev:
@Anthchirp 
I removed the dependency: 415fdc8aa8789a0472634788a842ea93ed88df01

Anthchirp:
This has been fixed

PR #435, opened 2020-01-29
Fix some corner cases in the 'counting_sorted' binning algorithm

benjaminhwilliams:
Working with small molecule diffraction data, with few reflections, has highlighted an issue with the `'counting_sorted'` binning algorithm of `cctbx.miller.set` with implications for `iotbx.merging_statistics.dataset_statistics`, amongst other things.

The problem occurs in cases where there are few reflections overall, and relatively many symmetrically-nonequivalent reflections with coincidentally close _d_-spacings, as is not uncommon in small molecule work.  The current algorithm can result in the creation of empty, zero-width bins.  This PR introduces the following slight modifications to the existing behaviour:
- In `cctbx.miller.set.setup_binner_counting_sorted`, bin edges are determined based on the set of _d_-values unique to within a tolerance specified with the existing parameter `d_tolerance`, rather than all the _d_-values, thereby avoiding empty bins.
- The way `cctbx.miller.set.setup_binner_counting_sorted` treats the parameters `reflections_per_bin` and `n_bins` is changed slightly so that if both are not `None`, then `reflections_per_bin` is treated as minimum number of reflections per bin, and `n_bins` is treated as the maximum number of reflections per bin.  Previously if `n_bins` and `reflections_per_bin` were both provided, then only `n_bins` was used and `reflections_per_bin` was ignored.
- A new optional parameter `reflections_per_bin` in `iotbx.merging_statistics` exposes the corresponding parameter in `cctbx.miller.set.setup_binner_counting_sorted`.  It is ignored for other binning algorithms, where it has no meaning.

New and updated tests for this behaviour are included.

With corresponding modifications in DIALS, this is a partial fix for https://github.com/dials/dials/issues/1097.


benjaminhwilliams:
I'm trying to find out if this PR can be responsible for any of the unsuccessful CI checks but it's not easy to decipher the tangle of failures.  I can't see any of these failing tests that can possibly have been affected by the changes introduced here.  Quite a lot of them are simply `ImportError: No module named scipy`.  @bkpoon, is this normal for the CCTBX CI checks?  I'll merge master into this branch to see if that makes any difference.

Anthchirp:
these are due to https://github.com/cctbx/cctbx_project/issues/434

benjaminhwilliams:
OK, I guess that accounts for the missing SciPy imports, but perhaps not the other failing tests here, such as `cctbx_project/smtbx/refinement/tests/tst_least_squares.py` failing with an incorrect expectation value on Ubuntu 14.04.  Hopefully @bkpoon can provide some guidance as to how many red crosses is too many red crosses for a PR to be merged.

bkpoon:
The Base builds will fail because of the missing `scipy` imports and there are some unstable tests (like the one in `smtbx`). Nothing new seems to be failing.

benjaminhwilliams:
Great, thanks @bkpoon, that's good to know.

@rjgildea, could you suggest a suitable reviewer?  I don't know who's best placed.

PR #436, opened 2020-01-29
Repair cspad.cbf_metrology

biochem-fan:
286a084 passes a wrong file name to `dials.refine` after the first level.


phyy-nx:
Thanks.  It is tested in xfel_regression, so I'm curious why this error wasn't hit before.  I'll look into that later today, but for now I'll merge this.

phyy-nx:
Ok, so cspad.cbf_metrology has two methods, hierarchical and expanding.  Hierarchical wasn't tested, which is where this bug was found.  Thanks again @biochem-fan.  I've added a test for the hierarchical  method to xfel_regression.

PR #437, opened 2020-02-11
updates to tls/uij magnitude optimisation

nicholas-pearce:
-- added lasso and ridge regression penalties
-- added/expanded tests for new functions
-- fix white-space/indentation to be consistent 2-space throughout
-- rename variable names for clarity

all tests run on linux cleanly
libtbx.find_clutter runs cleanly
libtbx.find_unused_imports_crude runs cleanly
should be python3-compatible but not tested

### DIFF
--- a/mmtbx/tls/optimise_amplitudes.py
+++ b/mmtbx/tls/optimise_amplitudes.py
@@ -11,60 +11,10 @@ from mmtbx_tls_optimise_amplitudes_ext import *
 from libtbx import adopt_init_args
 
 import collections
-_OptimisationWeights = collections.namedtuple(
+OptimisationWeights = collections.namedtuple(
   typename = 'OptimisationWeights',
   field_names = ('sum_of_amplitudes', 'sum_of_squared_amplitudes', 'sum_of_amplitudes_squared'),
 )
-
-class OptimisationWeights(_OptimisationWeights):
-  """
-  Weights object for OptimiseAmplitudes class.
-
-  Parameters
-  ----------
-  sum_of_amplitudes : float
-    weight for the lasso-like term (constrains sum of amplitudes)
-  sum_of_squared_amplitudes : float
-    weight for the ridge-regression-like term (restrains each amplitude to zero)
-  sum_of_amplitudes_squared : float
-    weight for the square of the lasso-like term (restrains sum of amplitudes to zero)
-  """
-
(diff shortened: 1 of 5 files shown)


bkpoon:
I took a quick look. There is one Python 3 syntax change ("iteritems" to "items").

https://github.com/cctbx/cctbx_project/blob/de1a55ecfd34182cec36dc9a9cd5ef5d71e4885d/mmtbx/tls/optimise_amplitudes.py#L52

I am able to run the test in Python 3.7 after that change.

 I'm currently at SACLA, but I can take a closer look near the end of next week. Maybe @olegsobolev can take a look as well.

One comment is that can the `OptimisationWeights` be a subclass of `dict`? Or have a parent class that has the same properties of `dict`? Then you won't need to have separate functions for `generic` types and `dict` types.

mmtbx/tls/optimise_amplitudes.py
```diff
@@ -11,77 +11,208 @@
...
+    Transfer weights from an input object (dict or other OptimisationWeights object)
+    """
+    if weights is None:
+      return self
+    elif isinstance(weights, dict):
+      return self.update_from_dict(weights)
+    elif isinstance(weights, type(self)):
+      return self.update_from_dict(weights.__dict__)
+    else:
+      raise Exception('Unknown type for `weights`')
```
Anthchirp: Should be `ValueError`

mmtbx/tls/optimise_amplitudes.py
```diff
@@ -11,77 +11,208 @@
...
+    """
+    self = cls()
+    found_none = True
+    for k in self.__dict__.keys():
+      v = getattr(obj, k, None)
+      if v is None: continue
+      found_none = False
+      self.__dict__[k] = v
+    if found_none is True:
+      raise Exception('No overlapping attributes were found between objects')
```
Anthchirp: `ValueError`

olegsobolev:
Tests work in my hands. Looks good to me.

nicholas-pearce:
OK — great. Thanks to all for checking and for feedback. 

I’ll change the “not-a-dict” class as suggested — a named tuple seems appropriate. That commit will come early next week, probably.

Thanks,
Nick

bkpoon:
Thanks for making the changes!

PR #438, opened 2020-02-17
Ensure conda dependencies are met for DIALS installations

Anthchirp:
This is the conda-equivalent of the existing package management with pip
for base-ful installations.

Right now this only affects DIALS-native setups (`bootstrap.py --installer=dials` and equivalent) and is skipped for other cctbx bootstrapped conda environments, as it requires
1. the package `conda` to be installed as part of the conda environment (as far as I know this is not the case for any cctbx bootstrapped environments), and
2. the conda environment to be consistent (same)

Dependencies are declared in `libtbx_config` under the key `"conda_required"`,
the only packages declaring such dependencies at this time are DLS-internal packages.

The same exclusions apply as with the pip setup, that is if either
1. the environment variable `LIBTBX_DISABLE_UPDATES` is set, or
2. the installation is base-less

then the update check is skipped and no updates are attempted or performed

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -1841,7 +1841,10 @@ selfx:
     # Resolve python dependencies in advance of potential use in refresh scripts
     # Lazy-load the import here as we might not have an environment before this
     from . import pkg_utils
-    pkg_utils.resolve_module_python_dependencies(self.module_list)
+    if self.build_options.use_conda:
+      pkg_utils.resolve_module_conda_dependencies(self.module_list)
+    else:
+      pkg_utils.resolve_module_python_dependencies(self.module_list)
 
     for path in self.pythonpath:
       sys.path.insert(0, abs(path))
@@ -1880,6 +1883,8 @@ selfx:
 class module:
   """
   Attributes:
+    conda_required (List[str]):
+      List of conda package requirement specifiers. Should match PEP508-style.
     python_required (List[str]):
       List of python package requirement specifiers. Should match PEP508-style.
   """
@@ -1897,6 +1902,7 @@ class module:
       self.names = [name, name + mate_suffix]
       if (dist_path is not None):
         self.dist_paths = [dist_path, None]
+    self.conda_required = []
     self.python_required = []
 
   def names_active(self):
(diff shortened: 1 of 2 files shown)


graeme-winter:
If no further feedback would like to merge this tomorrow to prepare for Diamond shutdown.

Anthchirp:
You can disable this by setting `LIBTBX_DISABLE_UPDATES`, which is the same logic (and environment variable) as with python pip packages and what we decided on in https://github.com/cctbx/cctbx_project/issues/151

I also note that the issue title I picked is misleading - this change will not enable automatic updating of the conda environment, but will ensure that all package dependencies are met. It will only update packages where this is required to satisfy dependency declarations. I will change the issue title to reflect this.

graeme-winter:
> Sorry, I was traveling, but now I'm back.
> 
> Can the current default behavior of no automatic updates for conda be kept? There was a `LIBTBX_UPDATE_CONDA` environment variable that was checked previously. That would keep dependency management via conda contained to a step before any build configuration is done. Developers can choose to enable automatic updates

Since the functionality is fundamentally there (following the prior art in #151) I'm happy that this is already addressed

In discussion with @Anthchirp we believe that there should be no effect at this time for anyone resulting from merging this PR - it is purely to enable other packages we have automatically installing their own dependencies, as we had with the `pip` mechanism before.

bkpoon:
I disagree, the discussion in #151 was specifically about `pip` packages, not `conda` packages.

from @graeme-winter 
> OK, some thoughts -
> 
> * if you have a baseless install, how will installing dependencies in base step help?
> * not aware of your conda based installation, but most installations are not conda based
> 

Since `conda` is different from `pip` and is much more comprehensive (which is why we are migrating to `conda`), the default should be different.

> In discussion with @Anthchirp we believe that there should be no effect at this time for anyone resulting from merging this PR - it is purely to enable other packages we have automatically installing their own dependencies, as we had with the `pip` mechanism before.

If this is a DLS system administration tool, then the default should not be to enable this for everyone else. This seems to be enabled for any environment that contains the `conda` package and any repository with `conda_required` in the `libtbx_config` file. We will likely be using `conda` to add code from other collaborators to user installations of Phenix, so we will also be including `conda` in our environments. The Phenix builds contain `dials` and `dxtbx`, which I suspect will have `conda_required`, so all Phenix developers will be affected. Is the plan to limit this feature to only DLS repositories?

PR #441, opened 2020-02-21
Fix get_gcc_version so that it doesn't return 'None' when gcc emits warnings into stderr

JBlaschke:
The current `get_gcc_version` function (in `libtbx/env_config.py`) deliberately returns `None` when `stderr` is non-empty. However, sometimes a compiler will output warnings in `stderr` (and still a valid version number in`stdout`). In these cases, the returned `None` can interfere with the build process.

This version still outputs `None`, if it fails to find a valid version number in `buffer.stdout`... so if your compiler is actually screwy, this is where the build system breaks (I guess that's what the `len(buffer.stderr)` business was all about).

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -129,16 +129,19 @@ def is_llvm_compiler(gcc='gcc'):
     return False
 
 def get_gcc_version(command_name="gcc"):
+  # run command in shell subprocess
   from libtbx import easy_run
   buffer = easy_run.fully_buffered(
     command="%s -dumpversion" % command_name)
-  if (len(buffer.stderr_lines) != 0):
-    return None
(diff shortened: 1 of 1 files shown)


JBlaschke:
@bkpoon bumping this (if you have time)

bkpoon:
Does this fix the problem on Cori or on Summit? Which system should I test?

JBlaschke:
> Does this fix the problem on Cori or on Summit? Which system should I test?

@bkpoon building on Cori exposed this problem. But it fixes a general issue: if `gcc` puts anything in `stderr` (like `libreadline` warnings), then the build fails because we're not processing the version number.

bkpoon:
Thanks!

PR #450, opened 2020-03-19
Smtbx fp,fdp refinement

dwpaley:
This was created jointly with @pcxod and adds refinement of inelastic form factors to the smtbx. A new constraint for shared groups of fp or fdp was added. A command-line script smtbx.anom_refine is provided to control the refinements in the common use case where all sites, ADPs, etc are fixed and the fps and fdps are refined alone. (for example see http://nuckolls.chem.columbia.edu/system/files/197/original/Electron_Cartography_in_Clusters.pdf ) Some tests are added (including an integration test for smtbx.anom_refine), and some additions to flex.xray_scatterer and cctbx.xray.structure permit shaking, setting, and extracting fp/fdp values for testing purposes.


dwpaley:
Updated tst_commandline_anomrefine to avoid using subprocess.check_output, which caused problems for both Windows and Python3.

bkpoon:
@olegsobolev and @pafonine Can you take a look at these changes? They modify scatterers and the flex array of scatters. The tests pass.

@dwpaley Are there any licensing issues with Columbia University? Also, I see implementations of `linearise` in `smtbx/refinement/constraints/shared.cpp`. Are they used or tested in these changes or somewhere else?

dwpaley:
Hi @bkpoon, 
There are no licensing issues. 
For the shared_***::linearise functions, there isn't a low-level test, but the shared_fp and shared_fdp constraints are exercised in `smtbx/refinement/constraints/tests/tst_constrained_structure.py` in a new test case `constrained_fpfdp_test_case`. I did confirm that test will fail if I break one of the linearise functions. It doesn't look like the other functions shared_u_star::linearise and so on are tested specifically, but I can put something together if you prefer.
Thanks,
Dan

bkpoon:
@dwpaley Alright, sounds good. I can merge this on Monday if @olegsobolev and @pafonine do not raise any issues.

Do you want to be a member of the cctbx organization so that you can check in bug fixes without a pull request?

dwpaley:
@bkpoon That sounds great and I'm happy to become a member so I can maintain this bit as needed.
Thanks
Dan

bkpoon:
@dwpaley I just sent an invite. Welcome!

olegsobolev:
Looks fine to me. They just add some functions to modify already existing
properties of scatterers. Scatterer class itself is untouched. I don't see
any issues.

bkpoon:
@dwpaley @pcxod Thanks!

PR #453, opened 2020-03-26
When using conda compilers ensure the setting sticks

Anthchirp:
and they are used again for subsequent builds.

When bootstrapping with `--compiler=conda` we need the environment activated. If a subsequent `make`/`make reconf` does not have the environment activated you will end up with build errors. A simple solution is to ensure the environment is activated right before the build step, ie. inside `Makefile`.
With this patch all you have to do is run the usual `. build/setpaths.sh` before running `make`, and there is no need to explicitly activate the environment.

Not sure what the effect of this is when using `conda build`.

Prerequisite for dials/dials#1201


Anthchirp:
To clarify - this has nothing to do with Diamond system administration (we do not use DIALS for this) and everything to do with dials/dials#1201 and ensuring compatibility between DIALS and other conda-forge packages. We no longer distribute (nor support, outside of the 2.2 LTS branch) DIALS using a libtbx base installation, and we never supported base-less installations. We only support DIALS on a conda-forge enviroment. Naturally, we should therefore be using conda compilers exclusively, and I want to get to that point that we have an environment that ensures conda compilers are used.

So I don't view using conda compilers as 'advanced' usage at all, but as the future default for DIALS development and distribution environments. As such stacking environments is not helpful for either scenario. The developer should have the compilers as part of their environment, because .. that is what they're there for. A user may not want to have to install compilers, but they should really just install a precompiled DIALS conda-forge package anyway, and this is outside the scope of this PR.

In short, what I am aiming for is to ensure we pick up the conda compilers every single time and eliminate the possibility of us erroneously picking up system compilers. Part of this will be to remove the need for the developer/user to run `. build/setpaths.sh` and some additional `activate` command.
Whether we achieve this by 
1. activating the conda environment in `setpaths.sh`,
2. eliminating the need for sourcing `setpaths.sh` entirely and relying only on environment activation, or
3. ensuring the correct compilers are picked on compilation
I don't mind. This PR laid out the path for 3, as I considered that less invasive. But I'm also perfectly happy to go for one of the other two. Alternatively we could have another setting for 3, eg. `--compiler=force-conda`, for this particular purpose.

What would you suggest?

phyy-nx:
Hi all, I think it's important to note we are trying to support your use case.  Using a standard set of compilers seems useful.  However, from what I understand, using the conda compilers outside of `conda-build` (the program that creates conda packages from recipes), is difficult.  `conda-build` creates its own environment every time it is run with all the packages it needs before it even tries to build cctbx, and specific build dependencies need to be added using yaml files (see for example [CDT and libGL](https://conda-forge.org/docs/maintainer/knowledge_base.html), which @bkpoon is [working on for cctbx](https://github.com/bkpoon/staged-recipes/blob/cctbx/recipes/cctbx/yum_requirements.txt)).  Without `conda-build`, there can be ABI mis-matches, specific libraries needed, and so forth.  See for example @bkpoon's point where he said that gltbx is failing to build using the conda compilers on CentOS 6/7.  In other words, making the conda compilers the default for any cctbx derived program, outside of the program conda-build, may not be trivial.

bkpoon:
Leaving aside the difficulties in getting the `conda` compilers to work for a moment, how about a 4th proposal? For XFELs, Aaron and Nick have been writing separate `setup.sh` scripts that activate the environment so that they can access programs like `mysql` or `mpirun`.  What if we codified creating these scripts into bootstrap.py if you are using `conda`? After the configuration step in `bootstrap.py`, an additional script is written that simplifies things (`bash`/`csh` and maybe `bat` versions too).

Let's give two use cases. First, given a clean linux system without a `conda` installation and the working folder, `$WORKING`, the following script would be created, if you supply `--use-conda` with no arguments.
```
#!/bin/bash
source $WORKING/mc3/etc/profile.d/conda.sh
conda activate $WORKING/conda_base
source $WORKING/build/setpaths.sh
```
Second, given an existing conda environment that can be detected by bootstrap using `--use-conda $PATH_TO_CONDA_ENV`, `bootstrap.py` would would write a script that looks like this:
```
#!/bin/bash
source $PATH_TO_CONDA_INSTALLATION/etc/profile.d/conda.sh
conda activate $PATH_TO_CONDA_ENV
source $WORKING/build/setpaths.sh
```
We can implement this and submit another pull request.  Now, in the case of using the `conda` compilers where they are not already available, the commands would look like this:
```
python bootstrap.py --use-conda hot update base
source conda, conda activate conda_base, conda install compilers
python bootstrap.py --use-conda build --config-flags="--compiler=conda"
source setup.sh
cd build
make nostop # again, because gltbx doesn't work yet
```
In your case, since you already have the `conda` compilers installed in `conda_base`, the setup script would do what you want. And the flexibility of manually managing conda environments is preserved.

So before running `make`, you would need to run `setup.sh` (or whatever name for the new script will be).

Some notes that may be helpful for building using the `conda` compilers on macOS.  The documentation suggests that the macOS 10.9 SDK is needed: https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#macos-sdk.

bkpoon:
I added the changes in the `conda_setpaths` branch. I'll test that the `conda` compilers work before creating a separate pull request.

Issue #145, opened 2018-03-01
adding base/bin to path in setpaths.sh/csh


Anthchirp:
Are there any concerns adding the ```base/bin``` directory to $PATH?
I suspect the answer is 'no', but wanted to ask the question anyway.

(We need to be able to access the h5* executables of the installed hdf5 library, instead of ancient ones that may be available from the system or elsewhere.)

graeme-winter:
+1 I would very much like this

Oeffner:
On Windows the HDF5 programs live in base\HDF5-1.8.16\bin

bkpoon:
Would making dispatchers to the h5 executables in build/bin work? It would be similar to how build/bin/python is done (for development builds, not builds from installers).

The CCTBX runtime environment tries to be as least disruptive as possible, so the scripts for setting up the environment generally just adds build/bin to $PATH and maybe a bunch of variables for CCTBX locations (setpaths_all.sh).

If we add base/bin to $PATH, everything in base/bin would take precedence, which on Linux would include a bunch of GUI stuff that can be problematic for the user. Also, since the h5 executables dynamically link to hdf5.so in base/lib, we would also have to prepend base/lib to $LD_LIBRARY_PATH, which would also add a bunch of other libraries that would take precedence and potentially cause unintended consequences. The dispatchers avoid these issues by only setting up the library paths in its own shell process.

Lastly, there are methods of configuring CCTBX without a base directory (e.g. xfel with psana), so dispatchers would avoid that issue as well (or add a check for base before prepending $PATH).

Anthchirp:
What do you mean by "bunch of GUI stuff"?
Point about libraries well received, some redirectors clearly needed.
Last bit not really a concern I think, adding a non-existing path to $PATH is harmless.

Oeffner:
base packages gtk, xrender, png and such like is a bunch of GUI stuff I think

bkpoon:
Also, fontconfig, cairo, gdk, wxwidgets, and glib have binaries in base/bin. These dependencies are also built with the expectation that they are only used for CCTBX programs, not as a general replacement. For example, there is no JPEG support since that option is explicitly disabled for gdk (otherwise, we would have to provide libjpeg as a dependency).

[...]

PR #454, opened 2020-04-02
conda: add extra environment script

bkpoon:
From the discussion in #453 

When the `--use-conda` flag is used in `bootstrap.py`, an extra script is written that will activate the `conda` environment used for building before sourcing the standard setpaths script.

By default, there will be variations of the setpaths and unsetpath scripts named `conda_setpaths.sh/csh` and `conda_unsetpaths.sh/csh` for linux and macOS, and `conda_setpaths.bat` and `conda_unsetpaths.bat` for Windows. The prefix is `conda`, but that or the general filename can be changed.

This makes the packages (e.g. compilers, mpi, mysql, etc.) installed in the `conda` environment available along with the standard CCTBX build in a single script.

There may be issues as discussed in #145, but `conda` environments are more comprehensive than what we build with `install_base_packages.py`, and there is much more user testing in the `conda` community, so this should be okay.

### DIFF
--- a/libtbx/auto_build/bootstrap.py
+++ b/libtbx/auto_build/bootstrap.py
@@ -1806,6 +1806,11 @@ environment exists in or is defined by {conda_env}.
         description="permit execution of config_modules.sh",
       ))
 
+    # write extra setpaths script for conda
+    if self.use_conda is not None:
+      self.add_command('libtbx.install_conda', args=['--write_setpaths'],
+                       description='Writing additional setup scripts for conda.')
+
   def add_make(self):
     self.add_command('libtbx.scons', args=['-j',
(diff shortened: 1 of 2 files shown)


Anthchirp:
I haven't yet had a chance to try this out, and would still like to do so (unless you want to merge it for other reasons)

Anthchirp:
I have now built a local solution in the dials repository (dials/dials#1235) that is loosely inspired by this and my other pull request. This new solution is stand-alone, so does not depend on any changes in `cctbx_project`.

If this PR is useful to you then feel free to merge, otherwise feel free to close this pull request.

Thank you

phyy-nx:
Thanks.  Merging.

PR #456, opened 2020-04-11
nanoBragg handedness

dermen:
For now this code supports all use cases I have come across, including CSPAD and Jungfrau16M models. 

Things to consider:

0. Can we think of a better handedness test other than ```dot_product(pix0_vector, odet_vector)```? We have to compute the dot product regardless, so maybe this is fine as is.
1. The function ```init_beamcenter``` will sometimes recompute the ```odet_vector``` . If it does so then it checks the boolean ```detector_is_righthanded```, and rescales accordingly. 
2. Python-visible methods ```set_roi``` and ```set_beam_convention``` both call ```init_beamcenter``` under the hood. Could this get us into trouble ? 
3. Should we worry about the user that sets ```odet_vector``` manually, and unknowingly switches the hand ?

### DIFF
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -96,6 +96,12 @@ nanoBragg::nanoBragg(
     Fclose = Xclose = -dot_product(pix0_vector,fdet_vector);
     Sclose = Yclose = -dot_product(pix0_vector,sdet_vector);
     close_distance = distance =  dot_product(pix0_vector,odet_vector);
+    if (close_distance < 0){
+        if(verbose)printf("WARNING: dxtbx model seems to be lefthanded. Inverting odet_vector.\n");
+        vector_rescale(odet_vector, odet_vector, -1);
+        close_distance = distance =  dot_product(pix0_vector,odet_vector);
+        detector_is_righthanded = false;
+    }
 
     /* set beam centre */
     scitbx::vec2<double> dials_bc = detector[panel_id].get_beam_centre(beam.get_s0());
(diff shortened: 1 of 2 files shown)


simtbx/nanoBragg/nanoBragg.cpp
```diff
@@ -96,6 +96,12 @@ nanoBragg::nanoBragg(
     Fclose = Xclose = -dot_product(pix0_vector,fdet_vector);
     Sclose = Yclose = -dot_product(pix0_vector,sdet_vector);
     close_distance = distance =  dot_product(pix0_vector,odet_vector);
```
dagewa, line 98: I think `odet_vector` is equivalent to `detector[panel_id].get_normal()`. Also, I think `close_distance` is equivalent to `detector[panel_id].get_directed_distance()`. This does the dot product over on the dxtbx side. I prefer using the dxtbx library for calculations like this, because it then keeps application code shorter.

simtbx/nanoBragg/nanoBragg.cpp
```diff
@@ -384,6 +390,7 @@ nanoBragg::init_defaults()
     Xclose=NAN;Yclose=NAN;close_distance=NAN;
     Fclose=NAN;Sclose=NAN;
     ORGX=NAN;ORGY=NAN;
+    detector_is_righthanded=true;
```
phyy-nx, line 393: It may be worth defining what right handed means here.  Namely, if the panel's normal vector (fast cross slow) is in the direction of expected propagation of photons through the material, then the detector is 'right handed'.

simtbx/nanoBragg/nanoBragg.cpp
dermen: I ultimately think it would be a good idea to write a test which verifies everything internal in nanoBragg lines up with dxtbx, if indeed one uses the dxtbx nanoBragg constructor.

nksauter:
Approved.

PR #457, opened 2020-04-13
WIP: Polarization anisotropy for fp and fdp

dwpaley:
This adds refinement of polarization anisotropy for f' and f" using smtbx tools. For now I'd like to ask for input on anything that should be significantly refactored. 

The biggest changes are the following:

- In iotbx/shelx/hklf.h, add a reader (actually an overloaded constructor for existing reader) that considers a .p4p and .raw file and constructs vectors for the polarization of incident and diffracted beams. 

- In cctbx/xray/observations.h, add data members to hold those polarization vectors

- In cctbx/xray/scatterer_flags.h, add flags use_fp_fdp_aniso, grad_fp_aniso, and grad_fdp_aniso. The bits in the existing data structure (a long unsigned int) were all used up, so these are implemented as bools.

- In cctbx/xray/scatterer.h, add fp_star and fdp_star. As suggested by the name these are meant to be analogous to u_star.

- Many changes in smtbx/structure_factors/direct/standard_xray.h. Where possible, the changes are separated out: a new heir to one_scatterer_one_h::base called in_any_sg_with_polarization, and overloads for one_h::base::compute and one_h::modulus_squared::compute. I believe the existing code doesn't ever touch these. A few new data members were added to one_h::base and one_scatterer_one_h::base

Existing tests pass, but there is essentially no test coverage yet for the new code. That is the biggest reason for the WIP but I will be checking the correctness, stability etc of the calculations and adding some usability features as well. The major changes that touch existing/common code are 99% or 100% finished.

@bkpoon @pcxod @olegsobolev @luc-j-bourhis I think some or all of you will want to look at this.

Thanks,
Dan Paley

### DIFF
--- a/cctbx/array_family/boost_python/flex_ext.cpp
+++ b/cctbx/array_family/boost_python/flex_ext.cpp
@@ -42,6 +42,7 @@ namespace {
     tuple_mapping_fixed_size<cctbx::grid_point<> >();
     tuple_mapping_fixed_size<cctbx::hendrickson_lattman<> >();
     tuple_mapping_fixed_size<cctbx::miller::index<> >();
+    tuple_mapping_fixed_size<cctbx::miller::index<double> >();
   }
 
   void register_cctbx_c_grid_conversions()
--- a/cctbx/array_family/boost_python/flex_miller_index.cpp
+++ b/cctbx/array_family/boost_python/flex_miller_index.cpp
@@ -122,6 +122,10 @@ namespace {
         first_index, (
           bp::arg("miller_index")))
     ;
+
+    flex_wrapper<miller::index<double> >::ordered(
+        "miller_index_double", flex_root_scope)
+    ;
   }
 
 }}} // namespace scitbx::af::boost_python
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -2068,21 +2068,37 @@ class array(set):
       assert sigmas.size() == self.indices().size()
     self._sigmas = sigmas
 
-  def e_incidents(self):
-    return self._e_incidents
+  def u_incs(self):
+    return self._u_incs
 
-  def set_e_incidents(self, e_incidents):
-    if e_incidents is not None:
-      assert e_incidents.size() == self.indices().size()
-    self._e_incidents = e_incidents
+  def set_u_incs(self, u_incs):
+    if u_incs is not None:
+      assert u_incs.size() == self.indices().size()
+    self._u_incs = u_incs
 
-  def e_scattereds(self):
-    return self._e_scattereds
+  def u_scats(self):
+    return self._u_scats
 
-  def set_e_scattereds(self, e_scattereds):
-    if e_scattereds is not None:
-      assert e_scattereds.size() == self.indices().size()
-    self._e_scattereds = e_scattereds
+  def set_u_scats(self, u_scats):
+    if u_scats is not None:
+      assert u_scats.size() == self.indices().size()
+    self._u_scats = u_scats
+
+  def v_scats(self):
+    return self._v_scats
+
+  def set_v_scats(self, v_scats):
+    if v_scats is not None:
+      assert v_scats.size() == self.indices().size()
+    self._v_scats = v_scats
+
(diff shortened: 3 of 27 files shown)


bkpoon:
@dwpaley It looks like there is some instability with GCC 4. There are occasional error messages that look like 
```
double free or corruption (fasttop):
```
when running a refinement test in `smtbx`. I do see some code where you do something special with GCC 4. Can you try lowering the optimization level to `-O2` and/or removing `-ffast-math`? For linux, I think these are the lines to edit
https://github.com/cctbx/cctbx_project/blob/7447555d44f2ef2ed89394587c194ddb495d2a3f/libtbx/SConscript#L636-L637
Alternatively, is it possible to set `grad_site` to `(0, 0, 0)` in the `compute_anisotropic_part` function in `smtbx/structure_factors/direct/standard_xray.h` instead of the GCC-specific fix? Explicitly setting `grad_site` in the function should hopefully prevent the compiler from putting nonsense in the array.

dwpaley:
Hi @bkpoon, it looks like the compiler magic belongs to @luc-j-bourhis:

`commit 9069261897c1ffd230f58b36a176577bde5d33b2
Author: Luc J. Bourhis <luc_j_bourhis@mac.com>
Date:   Thu Jul 30 19:47:25 2009 +0000

    work-around for the bug in gcc 4.0.0 discovered on Fedora Core 4 by nightly tests
    (utterly hideous)`

I just copy-pasted it into my new linearization in_any_sg_with_polarization, which follows the logic of the two existing ones.

For now as an experiment I tried setting grad_site inside compute_anisotropic_part as you suggest.

bkpoon:
Oh, maybe the magic only applies to GCC 4.0.0 and not the versions in CentOS 6 and 7, and Ubuntu 14.04. I'll have some time next week to help debug this. The tests are very unstable.

dwpaley:
@bkpoon I will spend some time on it today. I can reproduce it on OSX 10.14, but only in a debug build. It crashes with a few different messages, either the "double free" one or this equivalent one:

```
Python(16005,0x70000c055000) malloc: *** error for object 0x1290d2000: pointer being freed was not allocated
Python(16005,0x70000bf4f000) malloc: *** error for object 0x1290d2000: pointer being freed was not allocated
Python(16005,0x70000c055000) malloc: *** set a breakpoint in malloc_error_break to debug
Python(16005,0x70000bf4f000) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

or a segfault 11. It tries to free the same address up to 4 times. It seems like a separate issue from the GCC 4.0 one.

The crash happens when calling build_normal_eqns in smtbx/refinement/least_squares.py, line 84. I can reproduce consistently using smtbx test data (smtbx/regression/test_data) with 

`$ for f in f f f f f f f f f f; do smtbx.refine thpp.cif thpp.hkl=hklf4 out.cif --overwrite --max-cycles=1 >/dev/null; echo run; done` 

This gives about 5 crashes per set of 10 runs. If build_normal_eqns is called with may_parallelise=False, it never crashes.

The build_normal_eqns call goes to build_normal_equations in smtbx/refinement/least_squares.h.

I'll update with whatever else I find. Thanks for offering to help.

Dan

dwpaley:
@bkpoon Ok, I pushed a fix along with some other unrelated stuff. I don't understand what the problem was in full detail, but in short when accumulating to the normal equations in smtbx/refinement/least_squares.h, there was a check of observations.u_incs().size() where u_incs() is returning a shared array from cctbx::xray::observations.  The reflection accumulation is multithreaded and I think maybe the reference count was getting mixed up. 

Anyway the problem is now avoided by checking a bool observations::has_polarisation instead of passing around the whole array.

Thanks for pointing this out, I hope it will be more stable now.

bkpoon:
@dwpaley That seems to have done the trick. Thanks!

dwpaley:
This won't be ready to merge in until I can get some test data, which depends on beam time and will probably be a few months. I will close this for now. Everything will remain available on my fork https://github.com/dwpaley/cctbx_project/tree/fpfdp_aniso. Thanks @bkpoon for your input. Dan

PR #463, opened 2020-04-18
libtbx: remove -ffast-math flag for MacOS+conda

Anthchirp:
I found a failing test in `scitbx/lbfgs/tst_lbfgs_fem.py` while trying out compiling dials with conda compilers (dials/dials#1235 [(comment)](https://github.com/dials/dials/pull/1235#issuecomment-614516788)) and then saw the fix in 7976f7c16d7b941a236311d8c8fad36d36f04e24 coming through.

This fix however applies when using the system-native compiler, so I've also removed the flag when running with `compiler=conda` on MacOS.

I don't know if this is the best approach, and suspect @bkpoon will probably have a more well-informed opinion here. I didn't want to tie the flag to an xcode version because the compiler will be independent of that (and on the Mac that I'm running on there seems to be no xcode installed at all? Or maybe I'm just holding it wrong)

### DIFF
--- a/libtbx/SConscript
+++ b/libtbx/SConscript
@@ -791,8 +791,12 @@ int main() {
     else:
       env_etc.ccflags_base = ["-fPIC", "-fno-strict-aliasing"] + \
         gcc_common_warn_options() + \
-        ["-DNDEBUG", "-O3", "-ffast-math", "-funroll-loops"] + \
+        ["-DNDEBUG", "-O3", "-funroll-loops"] + \
         ["-DBOOST_ALL_NO_LIB"]
+      if env_etc.compiler != 'darwin_conda':
+        env_etc.ccflags_base.append("-ffast-math")
+        # -ffast-math changes scitbx/lbfgs/tst_lbfgs_fem.py behavior in Xcode 11.4
+        # minimization steps may be different than fortran standard
       env_etc.cxxflags_base = []
       env_etc.shlinkflags = ['-shared']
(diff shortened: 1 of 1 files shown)


Anthchirp:
Apparently removing the flag solves one issue but causes problems elsewhere, dials/dials#1239

bkpoon:
I'm getting the error for `test/command_line/test_symmetry.py` with and without the flag on macOS 10.15.4 and Xcode 11.4.1 (my version got automatically updated) for Python 2.7 and 3.6. The other tests seem to pass without the `-ffast-math` flag on Xcode 11.4.1  and fails with the flag. I'll double-check my 4 builds tomorrow. This is before @rjgildea fixes in https://github.com/dials/dials/commit/cde05fa01cf534e72b87ed3bac7f609f2328a0bd.

Generally, I would expect that dropping the `-ffast-math` flag should provide more consistent output since `-ffast-math` may result in math that is not standards compliant.

bkpoon:
I swapped the results. Removing the flag causes 3 new failures. I can attach my log files if that's helpful. The `test/command_line/test_symmetry.py` fails with and without the flag.

bkpoon:
@Anthchirp I think the number of developers on macOS that is not using Xcode or not using the `clang` from conda will be large enough to justify the effort to generalize this, so this fix is fine. If there is large enough demand, we can make a general check against `env_etc.clang_version` since I suspect this is a `clang` issue, not a macOS issue.

PR #467, opened 2020-04-22
EuXFEL AGIPD: py3-compatibility

kirienko:
`py3`-compatibility changes in `agipd_cxigeom2nexus.py`

@phyy-nx please have a look


phyy-nx:
Thanks!

Issue #458, opened 2020-04-14
boost python namespace conflict


picca:
Hello, I am working on a Debian package for dials and cctbx.

but when I tryed to install the package, I got this error

```
~/Debian/dials/dials$ LANG=C sudo debi
(Reading database ... 474720 files and directories currently installed.)
Preparing to unpack dials-doc_2.2.1+dfsg2-1_all.deb ...
Unpacking dials-doc (2.2.1+dfsg2-1) over (2.2.1+dfsg2-1) ...
Preparing to unpack dials_2.2.1+dfsg2-1_amd64.deb ...
Unpacking dials (2.2.1+dfsg2-1) over (2.2.1+dfsg2-1) ...
Preparing to unpack python3-dials_2.2.1+dfsg2-1_amd64.deb ...
Unpacking python3-dials (2.2.1+dfsg2-1) ...
dpkg: error processing archive python3-dials_2.2.1+dfsg2-1_amd64.deb (--install):
 trying to overwrite '/usr/lib/python3/dist-packages/boost/__init__.py', which is also in package libboost-mpi-python1.67.0 1.67.0-17+b1
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Setting up dials-doc (2.2.1+dfsg2-1) ...
dpkg: dependency problems prevent configuration of dials:
 dials depends on python3-dials; however:
  Package python3-dials is not installed.

dpkg: error processing package dials (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 python3-dials_2.2.1+dfsg2-1_amd64.deb
 dials
debi: error: debpkg -i dials-doc_2.2.1+dfsg2-1_all.deb dials_2.2.1+dfsg2-1_amd64.deb python3-dials_2.2.1+dfsg2-1_amd64.deb subprocess returned exit status 1
```

Indeed boost-mpi-python already provide a boost package. So there is a conflict between cctbx and boost-mpi-python.

It seems to me that you should not preempt the boost namespace and rename your boost package with another name or move it somewhere else below cctbx maybe.

thanks for considering.

phyy-nx:
Thanks for pointing this out.  Discussed this with @bkpoon this afternoon.  To summarize, the following current use case is not ideal as we shouldn't be claiming the `boost` namespace:

```
$ conda install -c cctbx-dev cctbx
$ python
>>> import boost 
```
(`import boost` imports our module named `boost` which includes convenience wrappers for loading our boost modules)

[...]

PR #470, opened 2020-04-26
Move code from boost.* to boost_adaptbx.*, #458

Anthchirp:
provide redirects that throw a `FutureWarning` on import to aid in
migration.

`boost.python.injector` was not migrated as it is deprecated anyway
(#386). The `DeprecationWarning` was upgraded to a `FutureWarning` instead.

This is step 1 towards resolving the boost python namespace conflict (#458).

I will prepare a changeset for existing code in `cctbx_project` as well.


Anthchirp:
Ran tests for scitbx, libtbx, iotbx, cctbx, rstbx before and after changeset.
FutureWarnings were displayed as expected and tests passed quietly after applying the changeset.

Anthchirp:
will merge this next week if no objections

Anthchirp:
Hmm. Are you _really, really_ sure about this?

I would just like to point out that what you call "messiness" in the `boost_adaptbx` directory means with this patch it will gain 5 `.py` files for a grand total of 9 `.py` files (for comparison: `libtbx` contains 81).

This "cleanliness" will come at a high price though; Instead of
```python
import boost_adaptbx.python
boost_adaptbx.python.something
```
we have to write
```python
import boost_adaptbx.boost.python
boost_adaptbx.boost.python.something
```
everywhere.
That translates into **a lot** of `boost`ing: 475 * 2 = **950** extra `boost.`s in the `cctbx_project` repository alone. That's 6 kB worth of `boost.` and that is just for all instances of `boost.python`. I didn't count the others and didn't look in any other repository. It also translates into extra effort when reading or writing any of the 233 affected files.

In contrast, the `boost_adaptbx` and `boost` directories have only ever seen 415 commits (43 in the last year, a lot of then Python 3 related), so it's not like this code sees a lot of activity, so I assert that the mental load saved by a cleaner `boost_adaptbx` directory is negligible.

Personally, I find redundancy in the import namespace name quite annoying and redundant and I'm very much onboard with [PEP 20](https://www.python.org/dev/peps/pep-0020/) on this:

> Flat is better than nested.

As for the `DeprecationWarning`s - well, it's the point of `DeprecationWarning`s to flag the use of deprecated APIs :smile:, plus they only show up in tests (does libtbx testing even show them? I honestly don't know). I'm happy to disable them in the PR though and enable them a week later. Announcing the change before the PR is merged makes no sense because people won't be able to make the change yet.

bkpoon:
> This "cleanliness" will come at a high price though; Instead of
```
import boost_adaptbx.python
boost_adaptbx.python.something
```
>we have to write
```
import boost_adaptbx.boost.python
boost_adaptbx.boost.python.something
```
> everywhere.

That is not quite true. If people want to shorten it, they can do what [`numpy`](https://numpy.org/devdocs/user/quickstart.html#an-example) and [`matplotlib`](https://matplotlib.org/tutorials/introductory/pyplot.html) do with `import something as something_else`. For example,
```
import numpy as np
import matplotlib.pyplot as plt
import boost_adaptbx.boost.python as bp

bp.something
```

I made updates to your branch in `fix-boost` and created a new pull request (#516) that moves the `boost` directory back to `boost_adaptbx`.

Anthchirp:
I offered to address this in this PR and was waiting for an answer to my question for the past three months. I don't see why you needed to open a new PR.

bkpoon:
Sorry, I did not think you wanted to make the changes since it sounded like you were arguing that adding 6 kB of "boost" to a 214 MB repository (110 MB without the .git directory) was too much. It was also not clear how it translates to extra effort since the changes are done programmatically as a search and replace.

I made a new branch since I wanted to keep your contributions but not modify your existing branch, so I made a new pull request.

Issue #478, opened 2020-05-18
Treatment of NaN in flex histogram


graeme-winter:
It seems that NaN's confuse flex histogram...

```
    ratio = flex.log10(t_i.select(t_nn) / r_i.select(r_nn)).as_double()

    print("Min: %f" % flex.min(ratio))
    print("Number < -5: %d" % (ratio < -5).count(True))
    
    r_hist = flex.histogram(ratio, data_min=-10, data_max=10, n_slots=10)

    for c, v in zip(r_hist.slot_centers(), r_hist.slots()):
        print(c, v)
```

It turns out there are 6,338 NaN's in ratio, which all end up in the smallest bin:

```
Min: -5.173720
Number < -5: 2
-9.0 6338
-7.0 0
-5.0 9
-3.0 59
-1.0 481347
1.0 343104
3.0 61
5.0 1
7.0 0
9.0 0
```

I believe `flex.histogram` should not be treating NaN in this way...

graeme-winter:
```
      std::size_t
      get_i_slot(ValueType const& d_)
      {
        std::size_t i_slot = 0;
        ValueType d = d_ - data_min_;
        if (d != 0 && d >= slot_width_) {
              i_slot = static_cast<std::size_t>(d / slot_width_);
          if (i_slot >= slots_.size()) i_slot = slots_.size() - 1;
        }
        return i_slot;
      }

    protected:
      void
      assign_to_slot(ValueType const& d)
      {
        slots_[get_i_slot(d)]++;
      }
```

in `scitbx/histogram.h` - guess

```
if (not isnan(d)) {
        slots_[get_i_slot(d)]++;
}
```

would fix this, opinions?

graeme-winter:
Update: this is not limited to `flex.histogram` - `flex.min` gives some weird results:

```
from scitbx.array_family import flex
import math

x = flex.double(range(-10, 100)) + 0.5
l = flex.log10(x)

for _ in l:
    if math.isnan(_):
        continue
    print(_)
    break

print(flex.min(l), flex.max(l))
```

gives

```
-0.3010299956639812
1.9800033715837464 1.9978230807457256
```

i.e. `flex.min()` gives _the wrong answer_ if there are NaN's in the input list

graeme-winter:
To be clear following the last comment: _either_ NaN or min(!NaN) values should be returned, not some other random element. Returning NaN would be consistent with `std::min` and the true min value with `fmin` from `math.h`

[...]

PR #479, opened 2020-05-18
Fix treatment of NaN in histogram

graeme-winter:
Includes test to assert this behaviour. Means histogram behaves as expected
for #478 however issues with min() and max() remain.


ndevenish:
This has a bug where ignored nan's don't contribute to the `n_out_of_slot_range_` count - they just disappear. It also seems vaguely whether `isnan` always works properly, even under `-ffast-math`? Though perhaps that part doesn't matter..

PR #483, opened 2020-05-28
Add tests for fixed merging statistics logic

benjaminhwilliams:
https://github.com/cctbx/cctbx_project/pull/477 changed the logic to decide whether data are already merged.  This PR introduces changes to the tests to accommodate the new behaviour.

The test data (`https://github.com/cctbx/cctbx_project/pull/477`) for the relevant tests are merged anomalous data.  When running `iotbx.command_line.merging_statistics.run` with the default parameter `anomalous=False`, we should therefore be able to calculate the merged non-anomalous data.  By contrast, with `anomalous=True` we should find that an exception (`Sorry`) is raised since the anomalous data are already merged.  These changes exercise that behaviour appropriately.

Also, I've tidied up `iotbx/regression/tst_merging_statistics.py` a little by removing some redundant parentheses and unused variables.

### DIFF
--- a/iotbx/merging_statistics.py
+++ b/iotbx/merging_statistics.py
@@ -541,11 +541,18 @@ class dataset_statistics(object):
     self.crystal_symmetry = crystal_symmetry
     i_obs = i_obs.customized_copy(
       crystal_symmetry=crystal_symmetry).set_info(info)
-    if (assert_is_not_unique_set_under_symmetry and
-        i_obs.as_anomalous_array().is_unique_set_under_symmetry()):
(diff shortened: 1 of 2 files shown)


benjaminhwilliams:
Apologies, I should have rebased this branch onto the current state of `master` before opening the PR.  In any case, I intend to squash-merge this as a single commit, so the branch history won't end up getting polluted.  I guess this way you can at least see both the commit that changes the tests and the commit that changes the behaviour in `iotbx.merging_statistics.dataset_statistics`.

phyy-nx:
Ok to merge after the CI checks finish.

PR #484, opened 2020-05-28
gltbx: use system libraries as fallback

Anthchirp:
When compiling with conda compilers in a conda environment the system
includes and libraries are not considered by default. This works fine
for 99% of the build, however `libGL.so` is not included in a conda
installation.

gltbx has a check whether `libGL` can be compiled against, and if it can't
then it won't be built. Here we add a fallback solution: if `libGL` can't
be compiled against **AND** we are using conda compilers **AND** we are on linux
then try explicitly to compile against system libraries. If this is not
successful then skip as before, otherwise build gltbx using the system
libraries.

I haven't checked yet what happens on MacOS platforms

### DIFF
--- a/gltbx/SConscript
+++ b/gltbx/SConscript
@@ -57,9 +57,41 @@ int main() { std::cout << GL_POINT << std::endl; return 0; }
   conf.Finish()
   if (flag and len(output.strip()) != 0) or libtbx.env.build_options.use_conda:
     conf = env.Configure()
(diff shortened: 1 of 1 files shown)


Anthchirp:
It seems that MacOS does not need any special treatment.

Anthchirp:
tests all passed, planning to merge this shortly

Anthchirp:
Cool, thank you!

PR #488, opened 2020-06-12
Remove pytest launcher

Anthchirp:
remove the `pytest` and `py.test` dispatchers.

I added this back when we started using pytest so that we could run 'pytest' commands directory even on environments where the base `bin/` directory was not added to the path. As we are now moving to regular python environments where the original pytest dispatcher will be available this creates an unnecessary conflict. As cctbx decided to not use pytest anywhere I suspect nobody will mind if I remove this?


phyy-nx:
FYI mmtbx and smtbx have pytest tests. I think removing this is fine though, since as you say the pytest launchers will still be available from conda and libtbx.pytest will continue unchanged.

Anthchirp:
after updating and a `make reconf`/`libtbx configure` you may see something along the lines of
```
$ pytest --help
-bash: ..../build/bin/pytest: No such file or directory
```
which is due to the removed dispatcher. This will only happen in terminals where the environment was loaded before the reconfiguration. You can fix the affected terminal by running `hash -r` or load the environment in a new terminal.

PR #489, opened 2020-06-18
Fixes to suffixtree

gbunkoczi:
This fixes segfaults with Python2 and Python3 (tried with gcc 4.8 and 8.1 with debug and release modes).

### DIFF
--- a/scitbx/suffixtree/boost_python/edge_exports.hpp
+++ b/scitbx/suffixtree/boost_python/edge_exports.hpp
@@ -7,14 +7,16 @@
 #include <boost/python/list.hpp>
 #include <boost/python/errors.hpp>
 #include <boost/python/def.hpp>
-#include <boost/python/iterator.hpp>
 
+#include <boost/noncopyable.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/functional.hpp>
 
-#include <scitbx/boost_python/iterator_range.hpp>
+#include <scitbx/suffixtree/boost_python/iterator_wrapper.hpp>
 #include <scitbx/suffixtree/edge.hpp>
 #include <scitbx/suffixtree/iterator.hpp>
+
+#include <iostream>
 
 namespace scitbx
 {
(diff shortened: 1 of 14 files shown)


bkpoon:
Let me run the Weekly build this weekend which tests different optimization levels.
@Oeffner Do the phaser tests pass? Also, I've set up a preliminary nightly test of `phaser_regression` here (https://dev.azure.com/phenix-project/nightly_builds/_build). I just have clean up some of the caching of builds.

Oeffner:
The phaser_py3 source I used is not a branch but a different repo. Doing regression tests on the current phaser repo compiled for python2 should work. But will fail if build for python3. The phaser_py3 repo will work now with Gabor's changes. I'll switch the phaser_py3 over on Monday once our Linux machines are up running again.

Oeffner:
Hi @gbunkoczi. The regression test cctbx_project/scitbx/suffixtree/test/tst_single.py now comes out with errors like
```
  File "..\..\phaserpy3python2build\modules\cctbx_project/scitbx/suffixtree/test/tst_single.py", line 501, in test_single_leaf
    cl = single.const_edge.from_edge( leaf )
AttributeError: 'module' object has no attribute 'const_edge'
```
I guess these are benign error but if you have time would you be able to rewrite that test so it passes?

gbunkoczi:
Hi @Oeffner,

I actually did (and it shows in the repository). Has it not propagated to
the pull request?

Oeffner:
@gbunkoczi yes you're right. I confused my source directories on my local PC

ndevenish:
There's a lot of change here. What was causing the SEGV?

gbunkoczi:
I honestly do not know. I suspected the shared_ptr to shared_ptr< const >
conversions, but removing these did not fix the issue. Eventually, I
replaced the iterator export with a custom one, and  it suddenly started to
work. I am not 100% happy with the current implementation, but it should be
good enough for now, so that people are not held up by this bug. I continue
looking at the problem, and hopefully find the root issue at one point.

Oeffner:
Thank you Gabor for fixing this, our regression test of sculptor_new, sculpt_ensembler and MRage now pass without crashes. Only issue is some curious difference of where MSE is changed to a MET residue when running as a python3 build.
https://www-structmed.cimr.cam.ac.uk/Local/PhaserNightly/contents.html

gbunkoczi:
I realise that this is probably outside the scope of this discussion, but
can you point me to the actual test that fails with the MSE->MET conversion?

Oeffner:
@gbunkoczi It's the test 
```
libtbx.python "/home/phaserbuilder/PhaserNightlyBuildPy3/Current/modules/phaser_regression/additional_regressions/sculptor_new/tst_sculptor_new.py"
```
You have to build phaser for python3. I use the command line
```
"/usr/bin/python" bootstrap.py --builder=phaser base hot update --use_conda --python=36 --nproc=7
```
Before the actual compilation I replace the phaser repo with https://git.uis.cam.ac.uk/cimr-phaser/phaser_py3.git
and then build it with command line:
```
"/usr/bin/python" bootstrap.py --builder=phaser build --use_conda --python=36 --nproc=7
```

PR #490, opened 2020-06-19
Look for conda eigen library as fallback

Anthchirp:
The logic in https://github.com/cctbx/cctbx_project/blob/2ce53dc81382c4761d8cf18d71b7276077e981fb/scitbx/examples/bevington/SConscript#L6 looks for eigen sources.
When `boost` is in `modules/` then it checks `modules/eigen`, however if conda boost sources are used it attempts to find the sources in `modules/cctbx_project/eigen` - it starts looking relative to `modules/cctbx_project/boost`.

Add a fallback that checks if the conda include path contains an `eigen3` directory, if so then use that one instead.

Related: dials/dials#1307


Anthchirp:
planning to merge this shortly

phyy-nx:
👍

PR #491, opened 2020-06-22
Updated try_all_readers()

keitaroyam:
Do not give up when encountered non-mtz/pickle binary file (for example, gzipped cif files can be treated)

### DIFF
--- a/iotbx/reflection_file_reader.py
+++ b/iotbx/reflection_file_reader.py
@@ -91,7 +91,6 @@ def try_all_readers(file_name):
     except KeyboardInterrupt: raise
     except Exception: pass
     else: return ("cctbx.miller.array", content)
-    return (None, None)
   try: content = cns_reflection_reader.cns_reflection_file(
     open(file_name))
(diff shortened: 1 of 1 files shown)


bkpoon:
Looks good to me. The `try_all_readers` function is able to open a compressed CIF file with reflections with this change.
```
iotbx.fetch_pdb -x 1yjp
bzip2 1yjp-sf.cif
libtbx.python -c "from iotbx.reflection_file_reader import try_all_readers; a = try_all_readers('1yjp-sf.cif.bz2'); print(a)"
```
returns `('cif', iotbx.cif.reader object at <somewhere>)`. Without this change, I get `(None, None)`.

If no one else has any comments, I can merge this on Monday. Thanks!

olegsobolev:
As usual, it would be great to add a test to make sure this will not regress over the time. Especially since it is already in @bkpoon comment.

bkpoon:
@olegsobolev We'll have to use an existing compressed cif file. We cannot assume that Windows will have bzip2 or gzip available.

PR #495, opened 2020-07-08
Clean up cbflib_adaptbx/exercise_compression

graeme-winter:
Set a random seed to make test more stable (though should be independent of this), fix some typos and clean up / make consistent variable names

Inspired by 87ba4692a010bc3b4736be9cfef7f68baf1e27f1


graeme-winter:
@bkpoon I think this test _should_ be stable - was worried by it not being - but would be interested if setting the seed resolves this - have also cleaned the test some as highlighted by your diffs

bkpoon:
Sure, let's try. For that past 30 days, the failure rate is ~1%. Let's reevaluate in the beginning of August.

PR #496, opened 2020-07-08
Enable C++11 for DIALS

Anthchirp:
DIALS now builds using C++11, so update the CCTBX bootstrap to reflect
this change, cf. https://github.com/dials/dials/pull/1235

Will merge when tests pass.


bkpoon:
Just FYI, none of the pipelines run the `dials` builder, so the tests should pass.

Anthchirp:
That's sort of what I was hoping for :slightly_smiling_face:

PR #499, opened 2020-07-11
Some code cleanup

Anthchirp:
remove some unnecessary compatibility constructs for unsupported python versions.

In particular the one in `wxtbx` caught my attention, as this (likely accidentally) globally disables `DeprecationWarning`s and `UserWarning`s based on the patch level of the python interpreter, ie. warnings are enabled on 2.7.4, disabled on 2.7.6, enabled on 3.6.5, disabled on 3.6.6, enabled on 3.7.3, etc...

The rest is mainly Python 2.2 and 2.6 compatibility stuff. The `sys.api_version()` used at one point has been a constant value (`1013`) since 2006.

### DIFF
--- a/cctbx/adp_restraints/tst_ext.py
+++ b/cctbx/adp_restraints/tst_ext.py
@@ -8,12 +8,10 @@ from cctbx.adp_restraints import adp_restraint_params
 import scitbx
 from scitbx import matrix
 import libtbx.load_env
-import math, os, sys
+import math, os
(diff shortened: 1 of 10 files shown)


Anthchirp:
If there are no concerns I'll merge this on the coming weekend

bkpoon:
I think these changes look fine.

Anthchirp:
Thank you!

PR #500, opened 2020-07-14
Add max_bins & min_bin_size keywords

rjgildea:
Allow these parameters to be set directly rather than having to construct and pass a phil params object. If a phil params object is passed, then values passed via that will take precedence.

### DIFF
--- a/cctbx/french_wilson.py
+++ b/cctbx/french_wilson.py
@@ -311,6 +311,8 @@ def french_wilson_scale(
       params=None,
       sigma_iobs_rejection_criterion=None,
       merge=False,
+      min_bin_size=40,
+      max_bins=60,
       log=None):
   from cctbx.array_family import flex
   if not miller_array.is_xray_intensity_array():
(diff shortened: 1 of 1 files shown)


cctbx/french_wilson.py
```diff
@@ -336,10 +338,10 @@ def french_wilson_scale(
...
       "a bug in one of the data processing and/or conversion programs.") %
       miller_array.data()[0])
-  if params == None:
-    params = master_phil.extract()
-  if (params.max_bins is None) : # XXX reset in case of user error
-    params.max_bins = 60
+  if params and params.max_bins:
+    max_bins = params.max_bins
+  if params and params.min_bin_size:
+    min_bin_size = params.min_bin_size
```
bkpoon, line 344: Maybe avoid checking `params` twice.
```suggestion
if params is not None:
  if params.max_bins is not None:
    max_bins = params.max_bins
  if params.min_bin_size is not None:
    min_bin_size = params.min_bin_size
```
Also, please add a comment that the default PHIL parameters are set in `master_phil` and that any changes to those default values should be reflected in the function definition and vice versa.

rjgildea: I've left this as it is, as personally I find the version with only two `if` statements and one level of nesting more readable.

PR #505, opened 2020-07-29
Update scitbx to the current state of the universe

Anthchirp:
SI definitions have changed on the 20th of May 2019:
https://en.wikipedia.org/wiki/2019_redefinition_of_the_SI_base_units
On that day the Planck constant and elementary electric charge were redefined
to be the exact values contained in this pull request.


Anthchirp:
👍 
Wanted to put this in as a PR to check that this doesn't cause any tests to fail.

PR #510, opened 2020-08-05
Trivial fix: typo in numer

graeme-winter:
PR in case anything depends on existing text output


graeme-winter:
All checks have passed -> merging

PR #511, opened 2020-08-06
Encapsulate gaussdev

nksauter:
This pull request makes a set of nanoBragg simulated images deterministic, with respect to scrambling the order in which the simulations are performed.  Previously the add_noise() function unintentionally stored global state in the static variable "iset", which unfortunately had the effect of making the "random" image noise unpredictable.  In the new redesign, the variable "iset" is re-initialized for each image, i.e., whenever the add_noise() function is called.

### DIFF
--- a/simtbx/nanoBragg/nanoBragg.h
+++ b/simtbx/nanoBragg/nanoBragg.h
@@ -120,9 +120,6 @@ double ran1(long *idum);
 /* ln of the gamma function */
 double gammln(double xx);
 
-/* return Gaussian deviate with rms=1 and FWHM = 2/sqrt(log(2)) */
-//take this out of header.  declare and define in cpp.
-//double gaussdev(long *idum);
 /* return Poissonian deviate given expectation value */
 double poidev(double xm, long *idum);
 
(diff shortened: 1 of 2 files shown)


simtbx/nanoBragg/nanoBragg.h
```diff
@@ -121,7 +121,8 @@ double ran1(long *idum);
 double gammln(double xx);
 
 /* return Gaussian deviate with rms=1 and FWHM = 2/sqrt(log(2)) */
-double gaussdev(long *idum);
+//take this out of header.  declare and define in cpp.
+//double gaussdev(long *idum);
```
JBlaschke, line 125: You've deleted the `gaussdev` definition in `nanoBragg.cpp` -- therefore this should be cleaned up (i.e. remove the comment)

JBlaschke:
RE making the RNG deterministic: I recommend using a counter-based RNG -- we can initialize the counters to the image ID. I've got good experiences with this library: https://www.deshawresearch.com/resources_random123.html -- cf https://github.com/hbrunie/conus of an example of how to use Random123. This would prevent us from having to "throw away" random numbers.

JBlaschke:
@nksauter I have maybe a naive question: is the seed reset for every image => every image gets the same random numbers this way?

nksauter:
Closing the branch.  Johannes raises two valid points 1) I reinitialize the gaussdev() state for each function call to poidev(), which is inefficient.  Therefore I'm closing this branch so as to start a new branch fixing this problem.  2) the framework as is will not support parallel execution; this is a larger problem not in the scope of work right now.

PR #514, opened 2020-08-11
Phenix/Schrodinger interface fix

latrocinia:
require PHENIX_SCHRODINGER env variable to activate Schrodinger interface in addition to SCHRODINGER env variable.

### DIFF
--- a/mmtbx/regression/tst_schrodinger_interface.py
+++ b/mmtbx/regression/tst_schrodinger_interface.py
@@ -1,0 +1,29 @@
+"""
+Test Phenix/Schrodinger interface
+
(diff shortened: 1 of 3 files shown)


nwmoriarty:
Great to have the test. Two comments.

1. We don't use "unittest" so not necessary.
2. We have some required import like __future__ etc so some checks failed.

But all fine to pull.

mmtbx/regression/tst_schrodinger_interface.py
```diff
@@ -0,0 +1,29 @@
+"""
```
dcliebschner, line 1: Please add
```suggestion
from __future__ import absolute_import, division, print_function
```

PR #519, opened 2020-08-21
remove completely unnecessary normalisation by number of datasets and…

nicholas-pearce:
Removed a dumb choice in my original implementation. Originally normalised by number of atoms, so relative precision changes with the number of atoms. find_clutter runs cleanly and no unused imports. tests run cleanly.

### DIFF
--- a/mmtbx/tls/optimise_amplitudes.cpp
+++ b/mmtbx/tls/optimise_amplitudes.cpp
@@ -368,14 +368,14 @@ void MultiGroupMultiDatasetUijAmplitudeFunctionalAndGradientCalculator::calculat
 
 void MultiGroupMultiDatasetUijAmplitudeFunctionalAndGradientCalculator::calculateFGLeastSquares()
 {
-  // Normalisation term (number of datasets)
-  double norm_all = 1. / (double)(n_dst * n_atm);
+  // Normalisation term (none)
+  double norm_all = 1.;
 
   // Normalisation term for atomic level
   double norm_res = 0.0; // if unchanged, atomic level will not be optimised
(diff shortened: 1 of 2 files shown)


mmtbx/tls/optimise_amplitudes.cpp
```diff
@@ -368,14 +368,14 @@ void MultiGroupMultiDatasetUijAmplitudeFunctionalAndGradientCalculator::calculat
 
 void MultiGroupMultiDatasetUijAmplitudeFunctionalAndGradientCalculator::calculateFGLeastSquares()
 {
-  // Normalisation term (number of datasets)
-  double norm_all = 1. / (double)(n_dst * n_atm);
+  // Normalisation term (none)
+  double norm_all = 1.;
```
bkpoon, line 372: Only comment would be to remove `norm_all` if you're just multiplying by 1. But if you're planning on trying another normalization procedure, then keeping it is fine.

nicholas-pearce:
Hi,

The code is ready to merge, and yes, future alternate normalisations are indeed the purpose of keeping norm_all.

However, I’m not sure what I’m meant to understand by your comment about a secret file and failing tests?

Nick

bkpoon:
The secret file is an SSH key used for downloading some data for the tests in `XFEL CI`. Because your branch is in a fork, secret files cannot be used in the automated tests because of security issues. For example, you could modify the automated tests in your branch to do something else with the SSH key instead of the intended purpose. We're planning on migrating the repository to use git-lfs anyway (probably on GitLab), so it's not a long term issue.

PR #522, opened 2020-08-28
Add a deprecate_method function.

dagewa:
This makes it easier to emit a `DeprecationWarning` for methods of Boost Python classes. From an idea by @Anthchirp.


boost_adaptbx/tests/tst_deprecation_tools.py
```diff
@@ -0,0 +1,30 @@
...
+def exercise():
+  with warnings.catch_warnings(record=True) as w:
+    warnings.simplefilter("always")
+
+    # Choose a Boost Python class for which to deprecate a method. Here we
+    # use rational.int.
+    from boost_adaptbx.boost import rational
+
+    original_value = rational.int().numerator()
+    deprecate_method(rational.int, "numerator")
```
Anthchirp, line 15:
```suggestion
  # Choose a Boost Python class for which to deprecate a method. Here we
  # use rational.int.
  from boost_adaptbx.boost import rational

  original_value = rational.int().numerator()
  deprecate_method(rational.int, "numerator")
    
  with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
```

to ensure the warning is thrown at call-time, not at setup-time.

Anthchirp (approved):
I would suggest putting the function into `boost_adaptbx/boost/python.py`, basically next to the `inject()` function.

boost_adaptbx/tests/tst_deprecation_tools.py
dagewa: Ok, can do. I was wrapping the first call to numerator too in order to be sure that didn't already raise a warning, but I suppose that isn't really necessary.

dagewa:
Thanks! It looks like the CI is happy, so I'll merge.

PR #523, opened 2020-08-28
Remove unused variables and typedefs

rjgildea:
A large number of compiler warnings for unused variables and typedefs makes it harder to spot warning genuine warnings for potential bugs such as that identified in f9e1677.


bkpoon (approved):
Thanks! I verified tests from Phenix.
@phyy-nx is on vacation, so he probably won't look at this until next week.

phyy-nx (approved):

PR #525, opened 2020-09-01
Fix bug

BobWangRobot:
fix bug in the atomic enviroment vectors. Add include_non_linked = True to generate_ca function.


nwmoriarty:
Adding Billy for some advice on possible conflicts with his changes. æ \chi @

nwmoriarty:
A confusing set of changes but here we go. 🤞

nwmoriarty:
@BobWangRobot 

Perhaps you should avoid root or sudo doing this process. Also make sure you do a "git pull" before you make a pull request.

PR #527, opened 2020-09-16
Remove redundant libtbx Python 3 syntax check test

Anthchirp:
Thanks to @bkpoon this is now covered by a Microsoft Azure test, which covers more code than just libtbx, is platform-agnostic, runs different versions of Python directly on any commit, has its own (up-to-date) exception list, and thus provides much more useful feedback.

We should therefore remove my old implementation of a python 3 syntax checker from libtbx.


Anthchirp:
Syntax checks are only failing due to unrelated mmtbx stuff.

libtbx/run_tests.py
```diff
@@ -54,7 +54,6 @@
   tst_list_base += tst_list_windows_fail
 # failing tests on Windows, Python 3
```
bkpoon, line 55: This whole section can be removed.
```
# failing tests on Windows, Python 3
tst_list_windows_fail_py3 = [
]
if sys.platform == 'win32' and sys.version_info[0] > 2:
  tst_list_fail += tst_list_windows_fail_py3
else:
  tst_list_base += tst_list_windows_fail_py3
```

bkpoon (changes requested):
Please remove some extra cruft. Thanks!

Anthchirp:
Test failures are unrelated (and fixed on master) → merging.

PR #529, opened 2020-09-20
Remove unused HDF5 nexus jiffy

Anthchirp:
As far as I can tell this was a programming jiffy from 2012 to create
minimalistic files that just about passed as HDF5 Nexus files.

In 2020 we have real data files that we can use, so this code is no
longer necessary.

This was flagged as it is the only place in cctbx_project having an
HDF5 library dependency.

@graeme-winter, since [you added this originally](https://github.com/cctbx/cctbx_project/commit/08bff560105c17696e32a361198513f2e29d79e5) - are you happy for this to go away?


Anthchirp:
for clarification: these files are not built and not referenced anywhere else.

graeme-winter (approved):
Yes, please do make this go away 👍

PR #533, opened 2020-09-25
Add fixed width integer types to flex arrays

### LINKED ISSUE #532: Provide more flex array types
For data processing tasks, we have identified a need for coverage of some more C types, for example `flex.short` to mirror `flex.long`, and some short unsigned types. This issue is to discuss which types we need, and to decide on naming and any additional features before we set up a branch to create these. Assigning @graeme-winter as the originator of the request, but I suspect various others may have an interest.

graeme-winter:
Added defined-width integer types to flex - signed and unsigned 8, 16, 32 &
64 bit using the usual fixed naming e.g. flex.(u)intNN. Added tests to ensure
rollover etc. behaves as C++; behaves the same as numpy; returns the right
numpy type as as_numpy_array. Added alias for int32 to int in flex.py for
consistency without duplication.

Fixes #532


graeme-winter:
Looks like I have some fixes to make based on the test failures, will pick up later. 

Not sure what the platform differences are here though...

graeme-winter:
Abandon hope all ye who enter here

Too much type aliasing, boost annoyance to give a complete quite of fixed width integer types so I am trimming this back to rely only on 8, 16 bit integers (which answers my current problems)

graeme-winter:
I think the test failures above are not obviously to do with the code changes (windows one I last checked was complaining that a downloaded zip was not a zip file...)

graeme-winter:
@bkpoon is there some kind of change log or news fragment or something I should add this to?

bkpoon:
I'll take a deeper look next weekend. I think we can avoid copying `flex_int.cpp` for each integer type. And I can try adding back the 32 and 64 bit types. Can I make changes to this branch? Or should I use a separate branch?

graeme-winter:
@bkpoon please do make changes in this branch - however in the interests of education, if you could avoid rewriting the history that would be great, so I can see what I have been doing wrong. 32, 64 bit types would also be great thanks 🙂

bkpoon:
In the first commit message, the comment says there's an alias for `int32` that uses `int`. I think that's the way to work around the registration problem, so I added that back. I wrapped the other types, but I think more tests are needed. I'll add more tests this week.

So the new types should work like before and you can do things like,
```
a = flex.uint64(range(10))
a.as_numpy_array().dtype
b = a - 100  # there should be no negative values in b
```

bkpoon:
I'll have to look at the converter registry to avoid multiple registrations.

bkpoon:
There is an `af::shared<unsigned int>` conversion in `iotbx.pdb` that's conflicting with `uint32_t`.

graeme-winter:
@bkpoon good spot, though what it is going in there I have no idea... at least explains why I could not make this work, thank you

graeme-winter:
@bkpoon scanning over the change set this looks sensible but it is not obvious to me that the work here is done - though the tests all pass. Is this now ready to roll? Good stuff if so, keen to use (strongly suggest squash merge here as I will want this picking over to a release branch)

bkpoon:
@graeme-winter Sure, I think this is ready now and squashing is fine. I just updated some existing tests. I do think more review is needed from @luc-j-bourhis, @pcxod, and @Oeffner since I reorganized some things in `flex` and `shared`.

Here is a summary of the changes that I made.

1. `flex_long.cpp` was removed and the `long` wrapping was rolled into `flex_int.cpp`. The signed integer types are in this file.
2.  The unsigned integer types are in `flex_size_t.cpp`
3. The functions in `flex_int.cpp` and `flex_size_t.cpp` were changed to use templates, and the wrapping code was changed to be a macro.
3. The `unsigned int` type was wrapped in `scitbx/array_family/shared_ext.cpp` and it was imported in Python as `from scitbx.array_family.shared import unsigned` This type is the same as `uint32_t` so it was moved from `shared_ext.cpp` to `flex_size_t.cpp` with the other unsigned integer types.
4. On Visual Studio, a `long` is the same as an `int` so `int64_t` is wrapped. On linux and macOS, a `long` is the same as `int64_t`. This is for modern 64-bit platforms.

On another note, some functions in `flex` may not work on the new integer types. For example, `flex.sort_permutation` would need to updated in `scitbx/array_family/boost_python/flex_sort.cpp` to wrap the new types.

graeme-winter:
@bkpoon sounds good 👍  thank you

Re: sorting &c. - my use case is to use these to hold pixel data so this is unlikely to be an issue however I do wonder if we should at least make an issue to note that these types _don't_ work with certain operations, and should cast to int before using?

Oeffner:
I just build phaser and ran some regression tests using python 2.7 and VS2008 as well as using python 3.6 and VS2015
In both cases phaser regression tests passed fine .

The output from libtbx.scons using VS2008 did have some extra warnings which I dont think I've seen when using the master branch:
```
c:\users\oeffner\oeffnerstuff\work\phaserpy2build\modules\cctbx_project\msvc9.0_include\stdint.h(242) : warning C4005: 'INTMAX_C' : macro redefinition
        c:\users\oeffner\oeffnerstuff\work\phaserpy2build\modules\boost\boost/cstdint.hpp(470) : see previous definition of 'INTMAX_C'
c:\users\oeffner\oeffnerstuff\work\phaserpy2build\modules\cctbx_project\msvc9.0_include\stdint.h(243) : warning C4005: 'UINTMAX_C' : macro redefinition
        c:\users\oeffner\oeffnerstuff\work\phaserpy2build\modules\boost\boost/cstdint.hpp(471) : see previous definition of 'UINTMAX_C'
```

Oeffner (approved):

bkpoon:
@graeme-winter Warning developers is a good idea. We can also wrap more functions as needed. Do you want this in the October release of cctbx in addition to the release branch in the DIALS fork? I'm picking the commit for the October release this week and waiting for more approvals is fine.

@Oeffner It seems okay with VS2015 and VS2017, so it should be fine.

graeme-winter:
Getting this in the October release would be neat, thanks - provided everyone happy with it 🙂

luc-j-bourhis (approved):
Olex2 relies on smtbx, so if all the tests pass, then I think it's good to go for me and Olex. I can't think of anything that relies on integer type length in stmtbx anyway.

bkpoon (approved):

bkpoon:
If no one else has any comments, I think this can be squashed and merged.

PR #541, opened 2020-10-09
Add r_anom to merging_statistics output

rjgildea:
Example output:
```
$ iotbx.merging_statistics $(dials.data get -q x4wide_processed)/AUTOMATIC_DEFAULT_scaled_unmerged.mtz
Format: ccp4_mtz
Miller array info: /Users/rjgildea/software/cctbx_py3/build/dials_data/x4wide_processed/AUTOMATIC_DEFAULT_scaled_unmerged.mtz:I,SIGI
Observation type: xray.intensity
Type of data: double, size=66052
Type of sigmas: double, size=66052
Number of Miller indices: 66052
Anomalous flag: False
Unit cell: (42.3695, 42.3695, 39.6915, 90, 90, 90)
Space group: P 41 21 2 (No. 92)

                     ----------Merging statistics----------                    

Resolution: 29.96 - 1.20
Observations: 66005
Unique reflections: 11773
Redundancy: 5.6
Completeness: 99.75%
Mean intensity: 400.1
Mean I/sigma(I): 6.1
R-merge: 0.100
R-meas:  0.110
R-pim:   0.045

Redundancies (non-anomalous):
  1 : 172
  2 : 619
  3 : 1048
  4 : 1556
  5 : 1955
  6 : 2489
  7 : 1997
  8 : 1200
  9 : 486
  10 : 202
  11 : 43
  12 : 6

  Statistics by resolution bin:
 d_max  d_min   #obs  #uniq   mult.  %comp       <I>  <I/sI>    r_mrg   r_meas    r_pim   r_anom   cc1/2   cc_ano
 29.97   2.58   7074   1288    5.49  99.31    2336.0    17.7    0.052    0.057    0.023    0.080   0.998   0.512
  2.58   2.05   7032   1203    5.85  99.83     725.8    12.7    0.114    0.125    0.050    0.113   0.983  -0.032
  2.05   1.79   6594   1187    5.56  99.83     293.2     8.4    0.167    0.183    0.074    0.161   0.969  -0.087
  1.79   1.63   6876   1171    5.87 100.00     166.1     6.7    0.200    0.220    0.088    0.185   0.959  -0.147
  1.63   1.51   6546   1165    5.62  99.74      92.4     4.4    0.252    0.277    0.114    0.244   0.947  -0.004
  1.51   1.42   6390   1158    5.52  99.74      60.3     3.2    0.314    0.346    0.143    0.295   0.905  -0.166
[...]

PR #542, opened 2020-10-14
Add R_{Friedel}

dagewa:
Following #541 I noted the similarity in calculation between R_{anom} and R_{Friedel}. The latter is used in electron diffraction experiment quality assessments. It is possibly too niche to add to `iotbx.merging_statistics` at this point, but having the function available for calculation seems useful.

### DIFF
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -2953,8 +2953,8 @@ class array(set):
 
   def anomalous_differences(self, enforce_positive_sigmas=False):
     """
-    Returns an array object with DANO (i.e. F(+) - F(-)) as data, and
-    optionally SIGDANO as sigmas.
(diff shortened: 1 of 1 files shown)


dagewa:
@bkpoon I added you as reviewer as you looked at #541.

dagewa:
converted to draft as it was pointed out to me that the equivalence between `r_anom` and `r_friedel` is only in triclinic space groups. I'll look into that.

dagewa:
This isn't ready yet. It requires unmerged Fs, or at least Fs that are only merged in P1. It could be calculated from unmerged intensities by merging in P1 then truncating to get Fs, but it would be better not to force apply the truncate procedure as part of the calculation. As this needs more careful bookkeeping I'll close the PR for now.

PR #546, opened 2020-10-20
Ngl hk lviewer2

Oeffner:
Apart from a few unrelated changes to the HKLviewer this pull request is mainly to provide an optional new constructor for iotbx.cif.builders.miller_array_builder that uses regular expressions for parsing reflection data labels in cif blocks. This makes it  more easy to associate data columns appropriately with one another. The default behavior is still to use the original constructor for backwards compatibility. The new constructor is invoked with the style="new" keyword.
This is useful when parsing reflection files as cif where the data labels are not the generic ones detailed in https://www.iucr.org/__data/iucr/cifdic_html/2/cif_mm.dic/index.html
The regular expression syntax is more compact and easier to extend to parse additional data labeling conventions.
The files concerned are 
cctbx_project\iotbx\cif\__init__.py, 
cctbx_project\iotbx\cif\builders.py

Example on the python commandline with 6c5f_phases_sf.cif (available from the RCSB):
```
>>> from iotbx.reflection_file_reader import any_reflection_file
>>> from iotbx import cif
>>> import os.path
>>> hkl_file = any_reflection_file("6c5f_phases_sf.cif")
>>> cifarrays = hkl_file.as_miller_arrays(merge_equivalents=False)
>>> for arr in cifarrays:
...   print( arr.info().labels)
...
['6c5f_phases', '_refln.DELFWT']
['6c5f_phases', '_refln.FC']
['6c5f_phases', '_refln.FC_ALL']
['6c5f_phases', '_refln.FC_ALL_LS']
['6c5f_phases', '_refln.FOM']
['6c5f_phases', '_refln.FP']
['6c5f_phases', '_refln.FREE']
['6c5f_phases', '_refln.FWT', '_refln.PHWT']
['6c5f_phases', '_refln.PHDELWT']
['6c5f_phases', '_refln.PHIC']
['6c5f_phases', '_refln.PHIC_ALL']
['6c5f_phases', '_refln.PHIC_ALL_LS']
['6c5f_phases', '_refln.SIGFP']

>>> cifarrays = hkl_file.file_content().as_miller_arrays(merge_equivalents=False, style="new")
>>> for arr in cifarrays:
...   print( arr.info().labels)
...
['_refln.FP', '_refln.SIGFP']
['_refln.FWT', '_refln.PHWT']
['_refln.FC', '_refln.PHIC']
['_refln.FC_ALL', '_refln.PHIC_ALL']
['_refln.FC_ALL_LS', '_refln.PHIC_ALL_LS']
['_refln.DELFWT', '_refln.PHDELWT']
['_refln.FREE']
['_refln.FOM']
>>>
```
Using the default constructor only some data columns are associated correctly into one miller array. But the new constructor correctly associates data and sigmas into one miller arrrays. The same goes for map coefficients.


olegsobolev:
@Oeffner , did you check if the new implementation can completely replace the old one? Or maybe with small adjustments. It looks like there should be only one correct way to parse data CIF into miller arrays, isn't it? I'm little worried that we'll have two chunks of ~250 lines of code to do the similar job. Do we have a plan of picking one (probably new one) later and removing the rest?

Oeffner:
> @Oeffner , did you check if the new implementation can completely replace the old one? Or maybe with small adjustments. It looks like there should be only one correct way to parse data CIF into miller arrays, isn't it? I'm little worried that we'll have two chunks of ~250 lines of code to do the similar job. Do we have a plan of picking one (probably new one) later and removing the rest?

@olegsobolev I take your point. I have not done an exhaustive tests of the new implementation on all the different reflection data cif files that may be in use, only a few ones with anomalous data and map coefficients. If the implementation fails to associate more data columns as one miller_array the fall back is always to just emit miller_array objects with only a flex.double array. 
@rjgildea or somebody else, would you know of a set of cif files I could test the new implementation on which represents all the generic reflection data labels specified by the mmCIF dictionary? 
From your paper, https://doi.org/10.1107/S0021889811041161 I gather that the cif parser is used by the Olex2 software as well. So it's important not to break existing functionality in a reimplementation of the old code.

PR #549, opened 2020-10-23
fix commit ffd70478

dwpaley:
If we use an environment in a yaml file, then we call `conda env
create`, which does not have a -y flag. HOWEVER if the environment is
in a .txt file, we call `conda create`, which needs -y or it will wait
forever for confirmation.

@bkpoon Does this look ok? I believe this is the correct implementation of the original fix.

### DIFF
--- a/libtbx/auto_build/install_conda.py
+++ b/libtbx/auto_build/install_conda.py
@@ -715,7 +715,7 @@ builder.""".format(filename=filename, builder=builder))
       command_list.append('--copy')
     if offline and not yaml_format:
       command_list.append('--offline')
-    if builder in ("dials", "dials-old", "xfel") and command != "create":
+    if builder in ("dials", "dials-old", "xfel") and not yaml_format:
       command_list.append("-y")
     if builder in self.env_without_python:
       python_version = tuple(int(i) for i in (python or "36"))


libtbx/auto_build/install_conda.py
```diff
@@ -715,7 +715,7 @@ def create_environment(self, builder='cctbx', filename=None, python=None,
       command_list.append('--copy')
     if offline and not yaml_format:
       command_list.append('--offline')
-    if builder in ("dials", "dials-old", "xfel") and command != "create":
+    if builder in ("dials", "dials-old", "xfel") and command_list[1] != "env":
```
bkpoon, line 718:
```suggestion
    if builder in ("dials", "dials-old", "xfel") and not yaml_format:
```
There is a `yaml_format` variable that is `True` when a `.yaml` file is found. That should be consistent with the other checks for `.yaml` files.

bkpoon:
Also, the tests should pass now (except for the CI/XFEL CI branch pipelines unless you update your branch to include the latest changes to master).

dwpaley:
@bkpoon Thanks. I made your change and rebased on master. I'll merge today if tests pass.

bkpoon:
There was an issue with the Windows CI, which is why they failed. This can be merged. Did you test a build with a `.yaml` file?

bkpoon (approved):

dwpaley:
Build from yml works fine after some unrelated troubleshooting. Thanks @bkpoon and @phyy-nx for the help.

PR #550, opened 2020-10-27
Remove platform.dist() call in installer

Anthchirp:
`platform.dist` has been removed in Python 3.8, meaning that the
`create-installer` script fails. The code was only used to identify
CentOS 5, which has been out of support over three years, so this
should be safe to remove now.


bkpoon (approved):
Thanks! I don't think this needs the full suite of tests to be run so just merge when you're ready.

Anthchirp:
Thank you!

PR #551, opened 2020-10-28
XFEL GUI: shifter fixes

phyy-nx:
This is a set of changes to get the XFEL gui submission running on NERSC using their [shifter](https://docs.nersc.gov/development/shifter/how-to-use/) system.

@JBlaschke this is mostly the shifter-fixes-2 branch, with just a bit of cleanup, including removing commented out code/extra print statements, dropping the hardcoded eV value for events without a wavelength and changing the burst buffer bit to only be used for shifter submission.

When merged, I'll delete the other shifter-fixes branches.


dwpaley:
Looks good. I've been using this branch for the last couple weeks and have a couple observations:

- When writing results on CFS, the staged logs in the burst buffer don't reach their final destination. Proposed fix:
  - Add a phil param `mp.staging = datawarp` 
  - Make `xfel/command_line/cxi_mpi_submit.py` line 362 say `if params.mp.method=='shifter' and params.mp.staging=='datawarp'`
  - Would also have to remove the lines in `sbatch.sh` that mention DW. At present, we can just delete any line containing `'DW'`.

- Sbatch and srun templates:
  - Can we put them under version control?
  - If there is no reservation, the corresponding line should be removed from `sbatch.sh`.

I volunteer to implement these, but please let me know if it's ok to work on this branch.

phyy-nx:
Totally ok to work on it.

For the template files, it makes sense to keep them in xfel/ui/db/cfgs.  Would we then remove the associated text boxes from the Advanced Settings dialog or pre-fill them in with these paths?  I remember we had different versions of these templates for different purposes?

PR #552, opened 2020-10-29
LCLS psana environment, build instructions and test

dwpaley:


dwpaley:
@bkpoon I'm sure there are some unnecessary flags in the two bootstrap runs for the LCLS build. Would appreciate if you can call out anything that is junk.

xfel/conda_envs/README.md
```diff
@@ -0,0 +1,65 @@
...
+There are two environments in this directory. Psana_environment.yml is suitable
+for general use and contains the usual CCTBX dependencies plus psana and its
+dependencies. Psana_lcls_environment.yml should only be used at LCLS; it adds
+a local build of mpi4py that is required for running jobs on multiple psana
+nodes.
+
+The build steps below were tested on Oct 28, 2020. They should be done in a
+clean environment (no cctbx installation has been activated, etc.)
+
+Creating the conda environment (step `base`) may take up to ~20 min with no
```
bkpoon, line 12: We can change `install_conda.py` to show output as it's happening. It is currently using the `check_output` function which stores the output (and the `--verbose` flag for `libtbx.install_conda` will display it at the end), but we can switch it to `call` which will display the output while processing.

xfel/conda_envs/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+$ ssh psexport
+$ cd $INSTALL; mkdir cctbx; cd cctbx
+$ wget https://raw.githubusercontent.com/cctbx/cctbx_project/master/libtbx/auto_build/bootstrap.py
+$ wget https://raw.githubusercontent.com/cctbx/cctbx_project/master/xfel/conda_envs/psana_lcls_environment.yml
+$ python bootstrap.py --builder=dials --use-conda=psana_lcls_environment.yml \
+  --config-flags="--compiler=conda" --config-flags="--use_environment_flags" \
+  --config-flags="enable_cxx11" --config-flags="--no_bin_python"             \
+  --no-boost-src --python=37 hot update base
```
bkpoon, line 44: The `--config-flags` flags are only needed for the `build` step so they can be removed here.

xfel/conda_envs/psana_lcls_environment.yml
```diff
@@ -0,0 +1,72 @@
...
+
+ # testing
+ - pytest
+ - pytest-xdist
+ - pytest-mock
+
+ # documentation
+ - docutils
+ - numpydoc
+ - sphinx
```
bkpoon, line 32: These are only necessary if you're building documentation.

xfel/conda_envs/psana_lcls_environment.yml
```diff
@@ -0,0 +1,72 @@
...
+ - tqdm
+ - eigen
+ - dials-data
+
+ # xia2
+ - tabulate
+
+ # Phenix
+ - send2trash
+ - PyRTF
```
bkpoon, line 68: The Phenix GUI uses these, so they're probably not needed.

bkpoon (changes requested):
The Check pipeline is failing (`xfel/conda_envs/test_psana_lcls.sh: trailing empty lines=4`), but other than that looks good.

xfel/conda_envs/README.md
dwpaley: That would be nice and would catch some problems early. E.g. when the -y flag was missing, I waited about an hour for something to happen...

xfel/conda_envs/psana_lcls_environment.yml
dwpaley: re this comment and the previous one: sphinx, pyrtf, and docutils are in the dials environments (https://github.com/dials/dials/blob/master/.conda-envs/linux.txt et al). My preference (as well as Aaron's suggestion) is to do the same here, to prevent any surprises, maybe at the cost of carrying around some unnecessary stuff. I'll remove send2trash and numpydoc. Sound ok?

xfel/conda_envs/README.md
dwpaley: Done, thanks

dwpaley:
Built and tested again at LCLS with the `config-flags` removed from the `hot update base` step and with `send2trash` and `numpydoc` removed from the environment. Everything looked ok.

bkpoon (approved):
Oh, the XFEL CI won't pass since the fork is in a separate fork. The SSH key for updating `xfel_regression` won't be available.

PR #554, opened 2020-10-29
Allow specifying python flags in run_tests.py

Anthchirp:
If specified these will be applied to all generated python commands within the relevant module, using the existing flag mechanism in `libtbx.run_tests_parallel`.

This allows modules to specify that tests should be run Python Development Mode by adding the line
```python
python_flags = "-X dev" if sys.hexversion >= 0x3070000 else ""
```
to `run_tests.py`

Context: I looked into [running DIALS tests in development mode](https://dev.azure.com/azure-dials/dials/_build/results?buildId=1518), and a lot of the resulting warnings go back to cctbx code (fixed a bunch in #553). It may therefore be of interest to the wider community to run tests of the central cctbx components in development mode, too.

Note: While the Python documentation does say the PDM makes tests run more slowly I only noticed - if anything - a decrease in test run times.

### DIFF
--- a/libtbx/test_utils/__init__.py
+++ b/libtbx/test_utils/__init__.py
@@ -222,7 +222,7 @@ def iter_tests_cmd(co, build_dir, dist_dir, tst_list):
     cmd += tst_path
     cmd += cmd_args
 
-    if hasattr(co,'python_keyword_text') and co.python_keyword_text:
+    if getattr(co, 'python_keyword_text'):
       cmd=cmd.replace("libtbx.python","libtbx.python %s" %(
          co.python_keyword_text))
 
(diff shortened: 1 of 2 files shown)


bkpoon:
The Python development mode sounds interesting. I turned it on in Azure Pipelines (by setting the PYTHONDEVMODE environment variable) and fixed some warnings. #555 and #557 should fix more of warnings. There are some failures, but they are just from the warnings (the Python 2.7 and 3.6 runs pass). I may turn PYTHONDEVMODE off next week until more of the warnings are addressed.

Since Python development mode and the tracemalloc debug tool can be enabled by environment variables (PYTHONDEVMODE and PYTHONTRACEMALLOC, respectively), that seems like a simpler approach than having to modify the testing utilities.

Anthchirp:
I didn't expect there to be enthusiastic uptake to enable this across the board for testing - but if you want to do that then that'd be great and then indeed you don't need these changes :+1:

On the other hand, this would also ensure locally run tests are always run with appropriate settings - and you can check whether there is an adverse runtime impact on some modules and potentially forego PDM testing there.

Anthchirp:
No longer relevant to us

PR #555, opened 2020-10-30
Fix failing libtbx.containers import

Anthchirp:
Was still seeing warnings in DIALS and realised the import line is wrong:

We try to import `OrderedDict` from `collections.abc`, but that will always fail as `OrderedDict` isn't there.
From Python 3.6+ onwards all CPython dictionaries are `OrderedDict`s anyway, so just point to `dict` instead.

(2.7 does fail the import so it bypasses that line, we do not support any Python 3.5 or any non-CPython 3.6)

### DIFF
--- a/libtbx/containers.py
+++ b/libtbx/containers.py
@@ -2,7 +2,8 @@ from __future__ import absolute_import, division, print_function
 
 import collections
 try:
-  from collections.abc import MutableSet, OrderedDict  # special import
+  from collections.abc import MutableSet  # special import
+  OrderedDict = dict
 except ImportError:
   from collections import MutableSet, OrderedDict  # special import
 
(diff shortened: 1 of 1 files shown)


Anthchirp:
I looked at the Python 3.8 implementation of `OrderedDict` and [it is subtly different from `dict`](https://docs.python.org/3/library/collections.html#ordereddict-objects).
It likely makes zero difference in 100% of our use cases, and, as a corollary, we should probably be switching to `dict`s everywhere. However, it would be insincere to offer the name `OrderedDict` when it is not.

(Then again one we probably shouldn't export this symbol in the first place, but that is a different discussion)

bkpoon:
Thanks! I think I put the `OrderedDict` in the wrong place. The `MutableSet` lines do not need the special comment since it is used. Other than that, you can just merge this without all the tests finishing.

PR #557, opened 2020-10-30
Fix more ResourceWarnings

Anthchirp:
observed a set of these in dxtbx tests
```
modules/cctbx_project/libtbx/introspection.py:232: ResourceWarning: unclosed file <_io.TextIOWrapper name='/proc/meminfo' mode='r' encoding='UTF-8'>
```

Again this is just using basic context managers, so will merge once tests pass.


bkpoon:
This can also be merged before all the tests pass. Most of the CI Branch (Python 3.9 is pending) passed and these changes should be fine.

Anthchirp:
Thank you

PR #558, opened 2020-10-30
TB_3DBUTTONS is deprecated

phyy-nx:
TB_3DBUTTONS is gone in wxpython 4.1.0 (dials/dials#1475).  This will affect any py3 wx4 compliant code that uses TB_3DBUTTONS.  Because of this I did a find/replace in cctbx_project to remove the style.

I tested removing TB_3DBUTTONS using cctbx.image_viewer in py2 and it worked with and without this change and without changing the toolbox visually.  So if phenix is using any of the files modified it should be ok.


bkpoon (approved):
Checked some GUIs on all 3 platforms and they look fine.

PR #560, opened 2020-11-03
iotbx.data_plots.table_data: column_formats bug fix if row contains None

rjgildea:
This would previously fail with e.g. `TypeError: %i format: a number is required, not NoneType`

### DIFF
--- a/iotbx/data_plots.py
+++ b/iotbx/data_plots.py
@@ -283,7 +283,9 @@ class table_data(object):
       else :
         if (self.x_is_inverse_d_min):
           row = [ math.sqrt(1/row[0]) ] + row[1:]
-        return [ (f % x) for f, x in zip(self.column_formats, row) ]
+        return [
+          (f % x) if x is not None else "*" for f, x in zip(self.column_formats, row)
+        ]
     else :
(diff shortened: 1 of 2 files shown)


bkpoon (approved):
Thanks! The docstring for the `format` function does show a `*` as a possible output so it's good that it does that now.

rjgildea:
> Thanks! The docstring for the `format` function does show a `*` as a possible output so it's good that it does that now.

Yes - if `custom_formats=None` with the same test input data, then the output does contain `*` - if was only when setting `custom_formats` that it could lead to an `TypeError` if the input data contained `None`.

https://github.com/cctbx/cctbx_project/blob/dff04a17993bbeecd783a965211881c4210e074b/iotbx/regression/tst_data_plots.py#L68-L104
From here:
https://github.com/cctbx/cctbx_project/blob/dff04a17993bbeecd783a965211881c4210e074b/iotbx/data_plots.py#L581-L586
via:
https://github.com/cctbx/cctbx_project/blob/dff04a17993bbeecd783a965211881c4210e074b/iotbx/data_plots.py#L288-L294

PR #566, opened 2020-11-19
On GPU, pre-check the exp() argument for the Gaussian shapetype

nksauter:
In analyzing the GPU performance of the Mullen-Holton CUDA kernel, it was discovered that significant time is spent evaluating the exp() function in double precision, signifying the offset from the Ewald sphere.  This only applies to the GAUSS shapetype for the reciprocal lattice point.  Furthermore, the argument can be checked before evaluation; if the exp() result will be essentially zero, it can be skipped.  Thread warps where all exp() values are insignificant (away from reciprocal lattice points) will coalesce, saving measurable GPU wall clock time.  The new code offers a new shapetype, GAUSS_ARGCHK where the exp() argument is pre-evaluated.

Also modified:  correctly initializing integer arrays with integer literals

### DIFF
--- a/simtbx/nanoBragg/nanoBragg.cpp
+++ b/simtbx/nanoBragg/nanoBragg.cpp
@@ -2459,6 +2459,7 @@ nanoBragg::show_params()
     if(xtal_shape == ROUND)  printf("ellipsoidal");
     if(xtal_shape == SQUARE) printf("parallelpiped");
     if(xtal_shape == GAUSS ) printf("gaussian");
+    if(xtal_shape == GAUSS_ARGCHK ) printf("gaussian_argchk");
     if(xtal_shape == TOPHAT) printf("tophat-spot");
     printf(" xtal: %.0fx%.0fx%.0f cells\n",Na,Nb,Nc);
     printf("Unit Cell: %g %g %g %g %g %g\n", a_A[0],b_A[0],c_A[0],alpha*RTD,beta*RTD,gamma*RTD);
@@ -2745,6 +2746,14 @@ nanoBragg::add_nanoBragg_spots()
                                         /* fudge the radius so that volume and FWHM are similar to square_xtal spots */
                                         F_latt = Na*Nb*Nc*exp(-( hrad_sqr / 0.63 * fudge ));
                                     }
+                                    if (xtal_shape == GAUSS_ARGCHK)
+                                    {
+                                        /* fudge the radius so that volume and FWHM are similar to square_xtal spots */
+                                        double my_arg = hrad_sqr / 0.63 * fudge; // pre-calculate to check for no Bragg signal
+                                        if (my_arg<35.){ F_latt = Na * Nb * Nc * exp(-(my_arg));}
+                                        else { F_latt = 0.; } // not expected to give performance gain on optimized C++, only on GPU
+                                    }
+
                                     if(xtal_shape == TOPHAT)
                                     {
                                         /* make a flat-top spot of same height and volume as square_xtal spots */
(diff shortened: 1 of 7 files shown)


nksauter:
Comment from @pafonine :  Everywhere (refinement, etc) where I need to evaluate exp() a lot I use a precomputed table (available somewhere deep down in CCTBX). This saves a lot of time! I think we also do sin/cos the same way.

biochem-fan:
Do you really need double precision? Does single precision, or the fast math version (`__expf`), lead to numerical instability? Oftentimes, only the accumulator variable needs to be `double`.

nksauter:
Interesting question.  I don't think we have enough data to decide on float vs. double at this stage.  The motivation for double is to allow numerous contributions to a given simulated pixel that might be of widely differing magnitudes.

biochem-fan:
It is worth investigating, because GPU is much faster with single precision numbers.

> The motivation for double is to allow numerous contributions to a given simulated pixel that might be of widely differing magnitudes.

A typical strategy in such a case is to calculate individual contributions in single precision and accumulate them into a double precision variable. For example, molecular dynamics simulation codes are often written in this way.

bkpoon:
During the hackathon, I don't think we had the time to investigate if we could do calculations in single precision and accumulate in double precision, but I also think it would be worthwhile to check since the hardware intrinsics are going to be faster and the single precision performance is generally going to be better (~ 2x from the hardware specifications).

My only other data point is from implementing the structure factor calculation using direct summation with CUDA and there was no significant difference with the single and double precision implementations on the GPU versus the regular CPU direct summation output. My array for accumulating was not double precision for the single precision calculation, though.

bkpoon:
The CI Branch error can be ignored.

JBlaschke:
All of this is interesting conversion -- here are my two cents:
1) I like the fast math/argcheck/single precision approaches -- would it make sense to expose switches at compile time to select between: 1. full double; 2. fast-math; 3. argcheck; 4. single precision? I think the "best" approach might depend on whichever application that kernel is used in (eg. single precision is usually good fine if the result can be refined to double precision)
2) I am somewhat hesitant of look-up tables. I wonder how they are better than reduced precision, or argcheck?

JBlaschke (approved):

bkpoon (approved):

pafonine:
Re tables: they've been in use in the context of crystallographic 
structure refinement since early days (read 70th or so) and that means a 
LOT of structures solved since then! Of course I have zero idea if and 
how this may fit here, so I voiced this option just for the sake of 
completeness! They do save a lot of time when exp() is called repeatedly 
many times in calculations like density sampling of structure factors 
calculation with no need to downgrade for lower precision.

Cheers,
Pavel

biochem-fan:
@pafonine How many entries do you have in the look up table?

russell-taylor:
That depends on the precision needed, the variability of the function, and whether you're feeding it into an iterative solver or using the built-in hardware texture interpolator to do linear interpolation.  (Note: the hardware linear interpolator has a limited interpolation precision.)  Lookup can be used as part of a multi-step process; for example, the division instruction in Intel processors uses a look-up-table based on the high-order bits in the number and then does iterations of a numerical solver to get full-resolution results.  For CUDA, you probably want to make sure that the table is entirely within constant memory for fastest lookup.  You can do the interpolation within your CUDA code, requiring two lookups for Linear and 3+ for higher-order interpolation.

JBlaschke:
Right, thanks all for elaborating.

I know that lookup tables are a common tool when representing expensive functions -- my experience with them though was limited to CPUs (in the context of memoization).

I was curious because I found only limited examples of their use when programming accelerators. My working theory is that linear interp + lookup would consume far too much fast memory (assuming that we're aiming for double precision). There are fancier methods out there (i.e. interp + lookup + additional math), but then we're getting into the domain of specialized math functions.

Anyway, feel free to correct my assessment (which is mainly based on anecdata). I want to keep this discussion in mind though moving forward. We went with the arg check approach because it produces accurate results with reasonable performance -- while being fast to implement ;)

PR #567, opened 2020-11-24
Build reduce whenever we build cctbx, so we'll have it available for …

russell-taylor:
…mmtbx.hydrogenate as we start to integrate it.  Switches the build source from svn to the current git master repository at rlabduke.

### DIFF
--- a/libtbx/auto_build/bootstrap.py
+++ b/libtbx/auto_build/bootstrap.py
@@ -990,7 +990,8 @@ class suitename_module(SourceModule):
 
 class reduce_module(SourceModule):
   module = 'reduce'
-  anonymous = ['svn', 'https://github.com/rlabduke/reduce.git/trunk']
+  anonymous = ['git', '-b master',
+               'https://github.com/rlabduke/reduce.git']
 
 class king_module(SourceModule):
   module = 'king'
@@ -1902,6 +1903,7 @@ class CCIBuilder(Builder):
     'tntbx',
     'clipper',
     'eigen',
+    'reduce',
   ]
   CODEBASES_EXTRA = []
   # Copy these sources from cci.lbl.gov
@@ -1989,6 +1991,7 @@ class PhaserBuilder(CCIBuilder):
     'tntbx',
     'phaser_regression',
     'phaser',
+    'reduce',
   ]
   # Configure for these cctbx packages
   LIBTBX = [


graeme-winter:
I note no license in https://github.com/rlabduke/reduce so not obvious can be distributed e.g. with `dials` (which would be a side-effect of this PR)

russell-taylor:
@graeme-winter Good call.  The individual source files have an open-source license description in them, but there is not a separate license file in the repository.  I'll look into getting a separate file added.

bkpoon:
Did you create a separate license file for `reduce`? I will have to modify the next build of the `conda` package to add that license.

bkpoon:
And there seems to be build failures on Windows now.
```
D:\a\1\modules\reduce\reduce_src\main.cpp(465): error C2065: 'HET_DICTOLD': undeclared identifier
```

https://dev.azure.com/cctbx/cctbx_project/_build/results?buildId=11237&view=logs&j=6920b088-0a9c-537e-50b9-bbc88f01c222&t=46247e75-bb72-5238-6395-27703cbb0595

russell-taylor:
The HET_DICTOLD would be my fault.  I'll take a look at why it is not building.  It works for me on VS 2019 but may not work in other environments.  Sorry about that.

russell-taylor:
There are a variety of licenses within the source files in the various subdirectories.  I did not make a separate license file.

russell-taylor:
Fixing the code now.

Where should the files be stored for the Windows command line?  C:/usr/local/ is where I have been putting them.

russell-taylor:
This happened when I merged the library version of reduce into rlabduke/master.  It was separate from the pull request to build reduce.  The build process re-pulls master each time, so this won't get caught in a pull request.  It may make sense to change to a specific commit in the cctbx project.

It should be fixed now in rlabduke/master.

bkpoon:
Thanks! Things are working now.

Which files have the licenses? I would still need to pull them out for the `conda` package. Or should I not include `reduce` in the `conda` package? I can do that while hydrogenate is finalized.

russell-taylor:
I'll pull out license files and point you at them.  That will be easier for all involved.

russell-taylor:
I added LICENSE.txt files in the subdirectories and in the main directory.  The one in the main directory contains all of the license variations from the subdirectories.

russell-taylor:
I've pointed the cctbx bootstrap.py at a tagged commit when it checks out reduce so that it will not be automatically changed when reduce master is changed.  We can update intentionally whenever we want to and regression test ahead of doing that.

Issue #575, opened 2020-12-23
cbf.h in modules/cbflib/include #undefines ULLONG_MAX, breaking VS2019 compile


russell-taylor:
When building using `python F:\taylorr\Personal\Work\consulting\RichardsonLab\src\cctbx_project\libtbx\auto_build\bootstrap.py --use-conda --python=38 --nproc=1` on Windows 10 from the Visual Studio 2019 Developer Command Prompt, we get a single error compiling.

That is caused by `#include <include/cbf.h>` which in the case of __cplusplus being defined undefines the ULLONG_MAX constant.  This appears to be a side effect of trying to determine what kind of structures to build internally, but it persists for later compilation.

I can work around this by rearranging the order of inclusions in tst_memory.cpp (pull request coming soon) but this seems like it is a trap waiting for the unwary in the future and should probably be fixed in the cbf.h file itself.

PR #576, opened 2020-12-23
Reordering include files to avoid failure due to cbf.h undefining ULL…

russell-taylor:
…ONG_MAX as a side effect.

Workaround for https://github.com/cctbx/cctbx_project/issues/575


bkpoon:
Thanks! We're working on getting a fix into `cbflib`. Feel free to merge once the pipelines finish. I'll update the VS2017 and VS2019 builds to us the `cctbx` builder after this is merged.

PR #577, opened 2020-12-27
Use binary mode in open

hodovani:
Use binary mode in open (current mode is `w`) to avoid encoding-related issues for written file, on Windows or with Python 3.


Anthchirp:
As the failing tests indicate opening the file in text mode is correct here. What error prompted you to open this pull request?

PR #581, opened 2021-01-05
Add as_size_t method

### LINKED ISSUE #580: Method needed to cast flex.int to flex.size_t
If I want to select I need a `size_t` but there appears to be no way to "cast" from `flex.int` to `flex.size_t` without the absurd solution of going through a Python list:

```
>>> from scitbx.array_family import flex
>>> a = flex.int(range(10))
>>> b = flex.size_t(a)
  File "<stdin>", line 1, in <module>
Boost.Python.ArgumentError: Python argument types in
    size_t.__init__(size_t, int)
did not match C++ signature:
    __init__(boost::python::api::object, boost::python::numpy::ndarray)
    __init__(boost::python::api::object, std::__1::vector<unsigned int, std::__1::allocator<unsigned int> >)
    __init__(_object*, scitbx::af::shared_plain<unsigned long>)
    __init__(_object*, unsigned long)
    __init__(_object*, unsigned long, unsigned long)
    __init__(_object*, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> >)
    __init__(_object*, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> >, unsigned long)
    __init__(_object*)
>>> b = flex.size_t(list(a))
```

This would be useful when you want to use a result of another `cctbx` method which returns `int` as a selection (e.g. using `ann`)

graeme-winter:
Copies the integer array into a size_t array such that it can be used for
selections etc.

Fixes #580

Usage:

```
>>> from scitbx.array_family import flex
>>> a = flex.int(range(10))
>>> b = flex.int(reversed(range(10))
... )
>>> a.select(b)
  File "<stdin>", line 1, in <module>
Boost.Python.ArgumentError: Python argument types in
    int.select(int, int)
did not match C++ signature:
    select(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > self, scitbx::af::const_ref<unsigned long, scitbx::af::trivial_accessor> indices, bool reverse=False)
    select(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > self, scitbx::af::const_ref<unsigned int, scitbx::af::trivial_accessor> indices, bool reverse=False)
    select(scitbx::af::versa<int, scitbx::af::flex_grid<scitbx::af::small<long, 10ul> > > self, scitbx::af::const_ref<bool, scitbx::af::trivial_accessor> flags)
>>> a.select(b.as_size_t())
<scitbx_array_family_flex_ext.int object at 0x7fc4ea437710>
```

### DIFF
--- a/scitbx/array_family/boost_python/flex_int.cpp
+++ b/scitbx/array_family/boost_python/flex_int.cpp
@@ -82,6 +82,21 @@ namespace scitbx { namespace af { namespace boost_python {
     return result;
   }
 
+  template <typename intType>
+  af::versa<size_t, af::flex_grid<> >
+  as_size_t(
+    af::const_ref<intType, af::flex_grid<> > const& O)
+  {
+    af::versa<size_t, af::flex_grid<> > result(
+      O.accessor(), af::init_functor_null<size_t>());
+    std::size_t n = O.accessor().size_1d();
+    size_t* r = result.begin();
+    for(std::size_t i=0; i<n; i++) {
+      r[i] = static_cast<size_t>(O[i]);
+    }
+    return result;
+  }
+
   /* For allowed syntax for the optional format_string argument see:
        http://www.boost.org/libs/format/doc/format.html#syntax
    */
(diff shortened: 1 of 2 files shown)


bkpoon (changes requested):
Looks fine, but please add a test. Since this is like the `as_long` function, you can add something similar to the `as_long` tests in `scitbx/array_family/boost_python/tst_flex.py` but without the negative numbers.

graeme-winter:
> Looks fine, but please add a test. Since this is like the `as_long` function, you can add something similar to the `as_long` tests in `scitbx/array_family/boost_python/tst_flex.py` but without the negative numbers.

I had a quick look before but did not completely get what that test was - have now added this just after in the same spirit

bkpoon (approved):
That works, it should test that the conversion from the other integer types matches what a `size_t` array would be when created from scratch.

The XFEL CI tests should pass once this is merged.

PR #584, opened 2021-01-25
uctbx.unit_cell: Fix f-string/__format__ behaviour with empty string

ndevenish:
Output from:
```python
from cctbx.uctbx import unit_cell
cell = unit_cell((80, 80, 80, 90, 90, 129.0))
print(" str:", str(cell))
print("fstr:", f"{cell}")
```
is:
```
 str: (80, 80, 80, 90, 90, 129)
fstr: 
```
Reason: When converting with `f"{cell}"` it's not `str(cell)` that is called but `format(cell, "")`, so the `unit_cell.__format__` gets called with an empty format specification (which would otherwise be passed in the `{expr:format_spec}` part of the format value). This then calls `format_spec.format(...)` - since `format_spec=""` this outputs an empty string.

This isn't a comprehensive treatment - I haven't tried to define the formatting semantics when using f-strings, but avoids this problem in the simple case of no specifier at all:
```
 str: (80, 80, 80, 90, 90, 129)
fstr: (80, 80, 80, 90, 90, 129)
```

### DIFF
--- a/cctbx/uctbx/__init__.py
+++ b/cctbx/uctbx/__init__.py
@@ -125,7 +125,9 @@ class _():
   def __repr__(self):
     return format(self, "uctbx.unit_cell(({}, {}, {}, {}, {}, {}))")
 
-  def __format__(self, format_spec="({:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g})"):
+  def __format__(self, format_spec=""):
+    if not format_spec:
+        format_spec = "({:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g})"
     return format_spec.format(*self.parameters())
 
   def show_parameters(self, f=None, prefix="Unit cell: "):


bkpoon:
Looks good, thanks!

Should the `__str__` and `__repr__` functions be updated so that the formatting is consistent? For example, should this 
https://github.com/cctbx/cctbx_project/blob/f6f58efb1b19a6681f50137deb010716313ff2f4/cctbx/uctbx/__init__.py#L122-L123
be changed to
```python
def __str__(self):
  return format(self, "")
```
and this
https://github.com/cctbx/cctbx_project/blob/f6f58efb1b19a6681f50137deb010716313ff2f4/cctbx/uctbx/__init__.py#L125-L126
changed to
```python
def __repr__(self):
  return format(self, "uctbx.unit_cell(({:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g}, {:.6g}))")
```

PR #587, opened 2021-02-04
Correct problem with Modulated Structures dict.

jongablop:
In some cases the reference obtained is not present in the dictionary. 

For example, ``_geom_bond_atom_site_label_1`` tag's reference is ``_geom_bond_atom_site_label_`` but that is not present in the Modulated Structures dictionary and it raises a KeyError: ``_geom_bond_atom_site_label_``.

This change tries to correct that error.

I'm attaching a CIF file that can serve as an example.
[bp5098Isup1.cif.txt](https://github.com/cctbx/cctbx_project/files/5924572/bp5098Isup1.cif.txt)

### DIFF
--- a/iotbx/cif/validation.py
+++ b/iotbx/cif/validation.py
@@ -369,7 +369,10 @@ class dictionary(model.cif):
         if isinstance(references, string_types):
           references = [references]
         for reference in references:
-          ref_data = self.get_definition(reference)
+          try:
+            ref_data = self.get_definition(reference)
+          except:
+            ref_data = self.get_definition(key)
           ref_names = ref_data['_name']
           if isinstance(ref_names, string_types):
             ref_names = [ref_names]
--- a/iotbx/cif/validation.py
+++ b/iotbx/cif/validation.py
@@ -371,7 +371,7 @@ class dictionary(model.cif):
         for reference in references:
           try:
             ref_data = self.get_definition(reference)
-          except:
+          except KeyError:
             ref_data = self.get_definition(key)
           ref_names = ref_data['_name']
           if isinstance(ref_names, string_types):


iotbx/cif/validation.py
```diff
@@ -369,7 +369,10 @@ def validate_loop(self, loop, block):
         if isinstance(references, string_types):
           references = [references]
         for reference in references:
-          ref_data = self.get_definition(reference)
+          try:
+            ref_data = self.get_definition(reference)
+          except:
```
bkpoon, line 374:
```suggestion
          except KeyError:
```
Based on the `get_defintion` and `find_definition` functions, it looks like the expected exception is a `KeyError`. Trapping the specific exception is helpful in case another part of the com fails. @olegsobolev and @nwmoriarty can double-check. This change should fix the failing Checks tests.

olegsobolev:
@bkpoon comments are definitely valid and need to be taken into account before merging. I would like to see more context. The Modulated Structure cif (msCIF) dictionary is a supplement to Core cif (https://www.iucr.org/resources/cif/dictionaries/cif_ms) and the latest version is from 2005. The example file does not pass validation against the Core cif:
```
~/Documents/phenix/testing/5 $ iotbx.cif.validate bp5098Isup1.cif.txt --dic=cif_core.dic 
Both data item '_diffrn_radiation_source' and exclusive alternate '_diffrn_source' present in data block
Invalid loop: value '?' present in child '_geom_hbond_atom_site_label_D' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_hbond_atom_site_label_H' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_hbond_atom_site_label_A' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_torsion_atom_site_label_1' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_torsion_atom_site_label_2' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_torsion_atom_site_label_3' but not found in parent '_atom_site_label'
Invalid loop: value '?' present in child '_geom_torsion_atom_site_label_4' but not found in parent '_atom_site_label'
```
It is not clear if msCIF is up to modern standards. The example file validates fine against e.g. mmcif_pdbx_v50.dic. Either way, this does not seem like a big deal and if @jongablop can explain why this is a problem and fix @bkpoon 's concerns, I think we can merge this.

jongablop:
Hi, thank you for your soon response.

I'm attaching another CIF file which does pass the core CIF validation. The msCIF is the current standard for modulated structures and the issue was detected by Luis Elcoro and Gotzon Madariaga, both developers of the Bilbao Crystallographic Server. In fact, G. Madariaga collaborated in the definition of the msCIF.

Regarding the previous example, there is another issue: the character '?' shouldn't be marked as an error, as it is used to express that the value of that field is unknown (https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax, point 57).
[bp5054sup1-edited.cif.txt](https://github.com/cctbx/cctbx_project/files/5963226/bp5054sup1-edited.cif.txt)
[bp5081sup1-edited.cif.txt](https://github.com/cctbx/cctbx_project/files/5963227/bp5081sup1-edited.cif.txt)

I'll apply @bkpoon 's corrections and see if all tests are passed.

jongablop:
Hi,

The only checks that haven't passed are the ones related with XFEL. I've seen that you changed that test's configuration recetly, could you re-send the test?

phyy-nx:
Rebasing onto master would fix the XFEL tests, but that shouldn't be a stopper for this PR. No worries from the XFEL side.

olegsobolev:
I'd delay the merge until after the Phenix point release (matter of days) just to be on the safe side. Otherwise, good to go in my opinion.

bkpoon:
It's not a problem. The Phenix release uses a branch in a separate fork.

PR #588, opened 2021-02-04
Make d_max and d_min the right way round

rjgildea:
The current code means that anyone calling `miller.array.setup_binner_d_star_sq_step` and wishing to override the parameters `d_min` and `d_max` must pass `d_min=d_max` and `d_max=d_min`. The `binning` constructor takes arguments in the order `d_max`, `d_min`, i.e. the opposite order to order of the variable names when the constructor is called from the existing Python code.

https://github.com/cctbx/cctbx_project/blob/08b8e83bb3f9ebe47188ecea6eba7db1f8243942/cctbx/miller/bins.h#L44-L49

### DIFF
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -1729,16 +1729,17 @@ class set(crystal.symmetry):
       assert d_star_sq_step > 0 or (d_star_sq_step is None)
     if auto_binning:
       d_spacings = self.d_spacings().data()
-      d_max=flex.min(d_spacings)
-      d_min=flex.max(d_spacings)
+      d_max=flex.max(d_spacings)
+      d_min=flex.min(d_spacings)
       del d_spacings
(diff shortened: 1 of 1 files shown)


bkpoon (approved):
Interesting, just to confirm, this does not change any behavior. It's just that the definitions for `d_min` and `d_max` were swapped and this change makes the definitions consistent?

Looks good, feel free to merge when you're ready. Thanks!

rjgildea:
@bkpoon yes - previously if you called `miller.array.setup_binner_d_star_sq_step ` setting `d_min` and `d_max` as expected, it would leave to the incorrect behaviour. With this change it will sort them internally such that `d_min < d_max` before passing them to the `binning` constructor, ensuring that we won't break any code that worked around the previous behaviour, whilst allowing new code to correctly set `d_min` and `d_max`.

PR #589, opened 2021-02-09
Performance optimizations for cctbx::miller::match_indices

dwpaley:
Add a `match_indices` constructor that only takes one miller array.
Then use `mi.match_cached(h1)` as a replacement for `match_indices(h0, h1)`,
but called as follows:

```
mi = match_indices(h0)
mi.match_cached(h1)
```

before using `mi.pairs()`, `mi.singles(i)`, etc. This is faster when a single
array `h0` will be used to match several arrays `h1`, `h2`, ... because a lookup map
is constructed from `h0` and cached.

Another new member function `match_cached_fast` has the same call signature
as `match_cached`, but subsequently only `mi.pairs()` is available. This is
even faster when you only want the pairs.

### DIFF
--- a/cctbx/miller/match_indices.cpp
+++ b/cctbx/miller/match_indices.cpp
@@ -69,6 +69,8 @@ namespace cctbx { namespace miller {
 
     miller_indices_[1] = miller_indices_1;
     pairs_.clear();
+    singles_[0].clear();
+    singles_[1].clear();
     af::shared<long> temp_pairs(lookup_map_.size(), -1);
 
     if (miller_indices_[0].id() == miller_indices_[1].id()) {
@@ -79,6 +81,13 @@ namespace cctbx { namespace miller {
       }
       return;
     }
+
+    singles_[0].reserve(miller_indices_[0].size());
+    singles_[1].reserve(miller_indices_[1].size());
+    if (miller_indices_[0].size() < miller_indices_[1].size())
+      pairs_.reserve(miller_indices_[0].size());
+    else
+      pairs_.reserve(miller_indices_[1].size());
 
     for(std::size_t i=0; i<miller_indices_[1].size(); i++) {
       lookup_map_type::const_iterator
@@ -111,6 +120,10 @@ namespace cctbx { namespace miller {
     pairs_are_valid_ = true;
 
     pairs_.clear();
+    if (miller_indices_[0].size() < miller_indices_[1].size())
+      pairs_.reserve(miller_indices_[0].size());
+    else
+      pairs_.reserve(miller_indices_[1].size());
 
     if (miller_indices_[0].id() == miller_indices_1.id()) {
       // short-cut if same array
--- a/cctbx/miller/match_indices.cpp
+++ b/cctbx/miller/match_indices.cpp
@@ -44,7 +44,6 @@ namespace cctbx { namespace miller {
         l = lookup_map_.find(miller_indices_[0][i]);
       if (l == lookup_map_.end()) {
         singles_[0].push_back(i);
-        ;
       }
       else {
         pairs_.push_back(af::tiny<std::size_t, 2>(i, l->second));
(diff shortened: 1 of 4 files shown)


nksauter:
My comment is that run-time Python help() can be made available for the new functions if you insert appropriate string comments into the boost.python wrapper in cctbx/miller/boost_python/match_indices.cpp. Admittedly this has not been common practice among cctbx developers, but I believe it is a good practice to adopt now for new work.  Just insert a quoted block of text here:
 .def("match_cached", &w_t::match_cached, "detailed help message for this function, can include newline")

dwpaley:
> My comment is that run-time Python help() can be made available for the new functions if you insert appropriate string comments into the boost.python wrapper in cctbx/miller/boost_python/match_indices.cpp. Admittedly this has not been common practice among cctbx developers, but I believe it is a good practice to adopt now for new work. Just insert a quoted block of text here:
> .def("match_cached", &w_t::match_cached, "detailed help message for this function, can include newline")

Done, thanks!

dwpaley:
@bkpoon Not sure if you'd like to suggest anyone else to review. `match_indices` is mostly untouched since written by @rwgk in 2002.

nksauter (approved):

cctbx/miller/match_indices.cpp
```diff
@@ -19,17 +35,16 @@ namespace cctbx { namespace miller {
...
     }
     std::vector<bool> miller_indices_1_flags(miller_indices_[1].size(), false);
     for(std::size_t i=0;i<miller_indices_[0].size();i++) {
       lookup_map_type::const_iterator
-        l = lookup_map.find(miller_indices_[0][i]);
-      if (l == lookup_map.end()) {
+        l = lookup_map_.find(miller_indices_[0][i]);
+      if (l == lookup_map_.end()) {
         singles_[0].push_back(i);
+        ;
```
bkpoon, line 47: This looks like an extra semicolon.

bkpoon:
I would suggest @nksauter, but he's already a reviewer. :smiley:

One other comment would be if reserving the capacity of `pairs_` to the shorter array size would further improve performance of `match_cached_fast`. That would avoid any reallocations with the `push_back` call. Also, is it possible to call `pairs_.m_set_size(0)` (or setting the `size` property) instead of `pairs_.clear()`? I think that would avoid deallocating memory after the first call.

I'm assuming this is for sorting out all the crystals in an XFEL experiment and figuring which images have which indices? And that this will be called thousands of times? Profiling with "normal" workloads would be needed if this helps.

Other than that, looks fine to me.

bkpoon (approved):

dwpaley:
> I would suggest @nksauter, but he's already a reviewer. 😃
> 
> One other comment would be if reserving the capacity of `pairs_` to the shorter array size would further improve the performance of `match_cached_fast`. That would avoid any reallocations with the `push_back` call. Also, is it possible to call `pairs_.m_set_size(0)` (or setting the `size` property) instead of `pairs_.clear()`? I think that would avoid deallocating memory after the first call.
> 
> I'm assuming this is for sorting out all the crystals in an XFEL experiment and figuring which images have which indices? And that this will be called thousands of times? Profiling with "normal" workloads would be needed if this helps.
> 
> Other than that, looks fine to me.

Thanks, this was super helpful!

We're calling it millions of times. It's part of an implementation of this algorithm: https://journals.iucr.org/d/issues/2014/01/00/wd5226/index.html, originally implemented in DIALS by @rjgildea and soon to be a new worker for cctbx.xfel.merge. To resolve a m-fold indexing ambiguity for n shots you need n**2 m**2 /2 correlation coefficients. Matching the indices was by far the slow step, caching the lookup map is a big improvement.

I did reserve the space for pairs and (more importantly) singles as you suggested. The non-fast version is almost as fast as the fast version now. Also found a bug while doing this.

Will merge tomorrow. Thanks for your help!

PR #590, opened 2021-02-13
Update XFEL build instructions

graeme-winter:
Couple of missing instructions, useful for the non-expert.


phyy-nx:
I'm going to run through the LCLS XFEL instructions today to make sure they are still working.  I imagine they need the same updates as were put in in cefad8803fcfded5368f20f52c737e636d8c681b.

PR #592, opened 2021-02-23
Adapt the cctbx.xfel GUI for Diamond light source

TiankunZhou:
Adapt the cctbx.xfel GUI for Diamond light source. The mysql will ask for root password when re-starts to raise the connections. job submission and tracking system are working fine now for data processing


xfel/ui/command_line/ui_server.py
```diff
@@ -90,6 +91,11 @@ def run(args):
 
     assert easy_run.call("mysqld --defaults-file=%s --initialize-insecure"%(cnf_path)) == 0
 
+  else:
```
phyy-nx, line 94: Raising the number of connections without mysql root access does fail for some versions of mysql but not others and I haven't tracked down why.  In the meantime, could you instead use a new parameter to control whether the root password is used?  That will allow us to submit mysql server jobs to a cluster without needing to enter the root password on systems where that is possible.

phyy-nx (changes requested):

phyy-nx (approved):

PR #594, opened 2021-02-26
A Python 3 fix for gltbx string rendering

dagewa:
I came across this function that appeared to be broken on Python 3, but is fixed by replacing a string literal with the equivalent bytes object.


Anthchirp:
You're overthinking this. On Python 2.7 `b"" == ""`, so you only need to add a single `b`.

dagewa:
Thanks, more like not thinking enough

PR #595, opened 2021-03-02
Add modify_cosym worker to cctbx.xfel.merge

dwpaley:
@phyy-nx @nksauter This has incorporated all the changes we discussed, and the tests are passing locally. 

Merging will be a tiny bit tricky because there's a corresponding branch in xfel_regression. I think the steps will be:

- Merge xfel_regression branch test_merge_polar
- Manually run pipeline "Update data cache"
- Revert commit 2a86c573 "temporarily test against xfel_regression branch test_modify_cosym"
- Merge this branch

This will cause XFEL regression test failures on other branches until they pull in these changes, but I think the failures should be self explanatory.

Any other comments?


dwpaley:
Just noting that all tests passed before the commit from this morning, which is just a README update.

nksauter (approved):

phyy-nx (approved):
The two comments above are optional :)

PR #597, opened 2021-03-04
Diamond tkz

TiankunZhou:
Changed mp.py and job.py. Now the merging function should work at Diamond. The submission command use only one node and will be fix in the future, probably.
Co-authored-by: Aaron Brewster <asbrewster@lbl.gov>


phyy-nx:
I cherry-picked the relevant commit and pushed it manually rather than merging this PR.  This PR has many older commits which made me a bit nervous, even though the diff looks right.  It's likely squash and merge would have been fine but I thought cherry-pick would be clearer.  Thanks @TiankunZhou!

PR #598, opened 2021-03-15
Mask

nksauter:
Special purpose kernel--just simulate the nanoBragg pixels over pre-determined Bragg spot shoeboxes.


nksauter:
Will merge in 24 hours if no issues come up.

JBlaschke (approved):
Discussed this with @nksauter in the group meeting. This makes from my perspective.

Issue #606, opened 2021-04-14
cctbx.xfel mpi_logger doesn't support debug logging


rjgildea:
The `mpi_logger` in cctbx.xfel doesn't support the full logging interface:
* it doesn't support [`logger.debug`](https://docs.python.org/3/library/logging.html#logging.Logger.debug)
* it doesn't support `logger.info(msg, *args, **kwargs)`, i.e. valid used such as `logger.info("string_%s_%s", foo, bar)`

https://github.com/cctbx/cctbx_project/blob/f8b5de70cca3ff76c2253b5501835655735c8d7b/xfel/merging/application/mpi_logger.py#L60-L61

This means that `cctbx.xfel` is liable to break if upstream code introduces changes that use otherwise perfectly valid logging statements, e.g.:
https://github.com/dials/dials/pull/1647/files#diff-4eb0e90752ec0c5de32be735a2b55c490308091767055a97f52eb6315a7fac07R204-R206

dwpaley:
Fixed by f7ce08b

PR #607, opened 2021-04-14
Compatibility with dials cosym clustering changes

rjgildea:
This removes the overrides of some private methods. It wasn't clear to me what they were doing that was different to the original dials.cosym versions, and after removing them and falling back on the original dials implementations, the tests still pass.

Caveat: the cctbx.xfel cosym test currently fails as a result of https://github.com/cctbx/cctbx_project/issues/606, however after commenting out the `logger.debug` statement in https://github.com/dials/dials/pull/1647 the test passes.

See https://github.com/dials/dials/pull/1647

### DIFF
--- a/xfel/merging/application/modify/aux_cosym.py
+++ b/xfel/merging/application/modify/aux_cosym.py
@@ -24,6 +24,10 @@ class CosymAnalysis(BaseClass):
     self.output_dir = kwargs.pop('output_dir', None)
     super(CosymAnalysis, self).__init__(*args, **kwargs)
 
+  def run(self):
+    super(CosymAnalysis, self).run()
+    if self.do_plot: self.plot_after_cluster_analysis()
+
   def plot_after_optimize(self):
           print ("optimized coordinates", self.coords.shape)
           xx = []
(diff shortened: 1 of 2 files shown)


dwpaley:
Richard, I made some small further changes here: https://github.com/cctbx/cctbx_project/tree/xfel-cosym-upstream

Could we close this PR and open a new one from my branch? I can take care of it tomorrow as long as it's ok with you.

Thanks
Dan

Anthchirp:
Both branches are in the same repository. You could just push your changes to this branch.

Anthchirp:
Hah, Github mobile does not in fact not show that the branch comes from a fork, so I take the above comment back.

(Still, you can push to this PR branch even though it is inside Richard's fork)

dwpaley:
> Hah, Github mobile does not show that the branch actually comes from a fork, so I take the above comment back.
> 
> (Still, you can push to this PR branch even though it is inside Richard's fork)

Thanks! I did not know that.

xfel/merging/application/modify/aux_cosym.py
```diff
@@ -24,6 +22,10 @@ def __init__(self, *args, **kwargs):
     self.output_dir = kwargs.pop('output_dir', None)
     super(CosymAnalysis, self).__init__(*args, **kwargs)
 
+  def run(self):
+    super(CosymAnalysis, self).run()
```
rjgildea, line 26: For Python 3 you don't need to provide any arguments to `super()`, i.e. the following is equivalent (and similarly for the `super().__init__` call above which isn't currently touched by this PR):
```suggestion
    super().run()
```

phyy-nx: I think our syntax checker would complain about that.  We are still running a py2 syntax checker on cctbx_project.

Anthchirp: (Note: syntactically this is fine. It just doesn't work on Python 2. However, Python 2 will already have crashed out way before it reaches this line, since it won't be able to import anything from dials.)

dwpaley (approved):

PR #612, opened 2021-05-10
safety checks after we write to array whitelist_kernel_model

dwpaley:


dwpaley:
> seems OK assuming the performance hit is negligible.

Yes, we only test these once per call to `analyze3` so the cost is definitely negligible.

PR #616, opened 2021-05-27
Add ABI checks for `scitbx.flex`

Anthchirp:
We see these tests sometimes fail when the ABI shifts slightly, eg.
dials/dials#1001. Adding the tests here to catch these situations early.

So far, no clear idea what causes this.

### DIFF
--- a/scitbx/array_family/boost_python/tst_flex.py
+++ b/scitbx/array_family/boost_python/tst_flex.py
@@ -1498,6 +1498,12 @@ mean:   2.00
   assert approx_equal(flex.double([3,7]).sample_standard_deviation(), 8**0.5)
 
 def exercise_complex_functions():
+  assert (flex.complex_double() == None) is False
+  try:
+    cd_none = flex.complex_double([None])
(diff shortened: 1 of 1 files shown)


ndevenish:
This also causes test fails in `cctbx/xray/targets/tst_r1.py`: https://github.com/cctbx/cctbx_project/blob/20d41ffe0d4b8dea04f416829b1f1aa35cc72efe/cctbx/xray/targets/r1.py#L36

and `cctbx/xray/targets/tst_shelxl_wght_ls.py`: https://github.com/cctbx/cctbx_project/blob/20d41ffe0d4b8dea04f416829b1f1aa35cc72efe/cctbx/xray/targets/__init__.py#L5

bkpoon (approved):
Looks good. I will be adding nightly tests of the `conda-forge` package and the next release will skip certain tests since some files are missing.

I'm testing different builds to sort this out.

Anthchirp:
Thank you and thanks for also looking into the underlying issue

PR #617, opened 2021-06-02
Fix bug in merging_statistics.as_cif_block function

jbeilstenedmands:
Upon iterating through the name dict, the test for finding the string 'percent' should be on the keys, not the values.

Added a line to a test to verify. This issue means that when the merging stats were converted to a cif block, the `_reflns.percent_possible_obs` item has been in the range 0 -> 1 rather than 0 -> 100 as intended.


bkpoon:
Looks good to me, thanks!

Can be merged once @rjgildea approves.

dagewa:
Can this be merged?

PR #618, opened 2021-06-05
Allow subclassing of sharing_handle

ndevenish:
I want to make scitbx arrays whose memory is managed other than via malloc/delete. This isn't currently possible because  `sharing_handle` isn't polymorphic - it doesn't have a virtual destructor (or anything), so it's impossible to reliably override the implementation.

This patch changes that by:
- Making destructor and functions virtual
- Protect against SEGV or double-freeing of data from subclasses

### DIFF
--- a/scitbx/array_family/shared_plain.h
+++ b/scitbx/array_family/shared_plain.h
@@ -72,7 +72,7 @@ namespace scitbx { namespace af {
       {}
 
       virtual ~sharing_handle() {
-        if (data != nullptr) {
+        if (data != NULL) {
 #ifdef SCITBX_AF_HAS_ALIGNED_MALLOC
             aligned_free(data);
 #else
--- a/scitbx/array_family/shared_plain.h
+++ b/scitbx/array_family/shared_plain.h
@@ -82,7 +82,7 @@ namespace scitbx { namespace af {
       }
 
       virtual void deallocate() {
-        if (data != nullptr) {
+        if (data != NULL) {
 #ifdef SCITBX_AF_HAS_ALIGNED_MALLOC
             aligned_free(data);
 #else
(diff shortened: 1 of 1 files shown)


graeme-winter:
Build errors appear unrelated to this change e.g. 

```
Installing h5py 2.10.0...
Read and write HDF5 files from Python
  log file is /__w/1/base_tmp/h5py_install_log
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 1682, in <module>
    installer(args=sys.argv, log=sys.stdout)
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 256, in __init__
    self.build_dependencies(packages=packages)
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 782, in build_dependencies
    getattr(self, 'build_%s'%i)()
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 1220, in build_h5py
    confirm_import_module='h5py', extra_options=["--no-binary=h5py"])
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 694, in build_python_module_pip
    log=log)
  File "modules/cctbx_project/libtbx/auto_build/install_base_packages.py", line 359, in call
    return call(args, log=log, verbose=self.verbose, **kwargs)
  File "/__w/1/modules/cctbx_project/libtbx/auto_build/installer_utils.py", line 55, in call
    raise RuntimeError("Call to '%s' failed with exit code %d" % (args, rc))
RuntimeError: Call to '/__w/1/base/bin/python -m pip install  "h5py==2.10.0" --no-binary=h5py' failed with exit code 1
```

bkpoon:
The Base tests have been failing because of some `SyntaxError` error with installing `h5py`. You can ignore those errors and I have disabled those tests.

graeme-winter:
@bkpoon since you've looked, suggested we ignore the errors and the other tests are fine I guess you're happy that we proceed with this very small change?

bkpoon (approved):
Alternatively, you can still use `nullptr`, but define that to be `NULL` when not using C++11 as the minimum.

PR #619, opened 2021-06-07
Uctbx lrl

dwpaley:
Add python bindings for parts of LatticeRepLib

This builds conditionally only if LatticeRepLib is installed in modules

### DIFF
--- a/cctbx/uctbx/lrl/match_lattices.py
+++ b/cctbx/uctbx/lrl/match_lattices.py
@@ -1,0 +1,51 @@
+from __future__ import division
+import cctbx_uctbx_lrl_ext as ext
+
+def get_input():
+  import sys
+  if len(sys.argv) > 1:
+    with open(sys.argv[1]) as f:
+      result = [l.strip() for l in f.readlines()]
+  else:
+    inp = sys.stdin.readline()
+    result = []
+    while inp.lower() != 'end':
+      result.append(inp.strip())
+      inp = sys.stdin.readline()
+  return result
+
+def selling_reduce(reader, mat=None):
+  lc = ext.LatticeConverter
+  if mat is None: mat = ext.MatS6()
+  lat, cell = reader.GetLattice(), reader.GetCell()
+  result = lc.SellingReduceCell(lat, cell, mat)
+  return result
+
+
+inputs = get_input()
+input_readers = []
+for inp in inputs:
+  reader = ext.LRL_ReadLatticeData()
+  reader.CellReader(inp)
+  input_readers.append(reader)
+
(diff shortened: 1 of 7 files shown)


bkpoon:
You can ignore the Base tests. If you rebase on top of `master`, those tests might not run.

dwpaley:
> You can ignore the Base tests. If you rebase on top of `master`, those tests might not run.

Thanks!

cctbx/uctbx/lrl/match_lattices.py
```diff
@@ -0,0 +1,51 @@
...
+  result = lc.SellingReduceCell(lat, cell, mat)
+  return result
+
+
+inputs = get_input()
+input_readers = []
+for inp in inputs:
+    reader = ext.LRL_ReadLatticeData()
+    reader.CellReader(inp)
+    input_readers.append(reader)
```
bkpoon, line 30: Inconsistent indentation

cctbx/uctbx/lrl/test_match_lattices.py
```diff
@@ -0,0 +1,41 @@
...
+
+  inputs = [
+      "r 57.98 57.98 57.98 92.02 92.02 92.02",
+      "r 57.98 57.98 57.98 92.02 92.02 92.02",
+      "h 80.36 80.36 99.44 90 90 120",
+      "c 80.95 80.57 57.1 90 90.35 90",
+      "h 80.36 80.36 99.44 90 90 120",
+      "r 57.1 57.1 57.1 89.75 89.75 89.75",
+      ]
+  inp_file = tempfile.NamedTemporaryFile(mode='w')
```
bkpoon, line 21: `inp_file` is not closed and may trigger a warning

bkpoon (commented):
Maybe @nksauter and @phyy-nx should also review this since would know more about this code. And are we adding it to Azure Pipelines? It can be run daily if it is not a priority.

Also, please rename `test_match_lattices.py` to `tst_match_lattices.py` so that it is consistent with the test names in `cctbx/regression`.

dwpaley:
Thanks @bkpoon. Requesting Aaron's review. I already discussed with @nksauter offline.

I don't think it's necessary to add this test to Azure. This isn't really functional right now, it requires some dev work on the LatticeRepLib side before we can use it for the intended purpose, which is to resolve pseudomerohedral indexing ambiguity. This PR is essentially a checkpoint so the work doesn't get lost.

phyy-nx (approved):
I believe the failing test is because `uctbx_lrl` is behind master, so it's not relevant.  So LGTM.

PR #620, opened 2021-06-08
Use bytes as IO for detector_image_as_png

phyy-nx:
With this fix, labelit.png (dispatcher for iotbx/command_line/detector_image_as_png.py) works in py3 (tested on dials_regression/image_examples/ALS_831/q315r_lyso_001.img)


bkpoon (approved):
👍

PR #621, opened 2021-06-14
optionally skip making a default crystal when instantiating NBcrystal

dwpaley:
The NBcrystal constructor takes about ~~0.5 s~~ 50 ms to make a dummy crystal. We want to instantiate thousands of these, so we need the option to skip the dummy crystal.

Edit: accidentally timed this on a debug build. The real time is 50 ms, but that is still a big fraction of our total runtime so this PR is still needed.


dwpaley:
Closing this to pull a few nanoBragg performance changes into a single PR

PR #623, opened 2021-06-16
Refactor euxfel agipd2nexus

kirienko:
Here we allow to build a NeXus file directly from the NeXus definitions (NXmx at the moment, but it principle — any).
We started from the @phyy-nx 's script for AGIPD and write another version which essentially does the same, but without definitions hard coded. This means that when definitions are changed, the output NeXus file might be easily regenerated without changing a single line of code. 

We introduce the notion of `Rules` here: by default, a dummy NeXus file is created, but that file is compliant with the definitions by design. But one may define a so-called `Ruleset` that defines the information should be written into particular places of the NeXus hierarchy, for example:
```python
        self.field_rules = {
            '/entry/definition': np.string_("NXmx"),
            '/entry/file_name': np.string_(self.output_file_name),
            '/entry/start_time': np.string_('2000-10-10T00:00:00.000Z'), 
            '/entry/data/data': LazyFunc(cxi.copy, "entry_1/data_1/data",  "entry/data"),
            '/entry/instrument/name':
                NexusElement(full_path='/entry/instrument/name', value=self.params.nexus_details.instrument_name,
                             nxtype=NxType.field, dtype=h5py_str(),
                             attrs={'short_name': self.params.nexus_details.instrument_short_name}),
            '/entry/instrument/AGIPD/group_names': np.array([np.string_('AGIPD'), np.string_('ELE_D0')], dtype='S12'),
            # and so on
            }
```

There are also some other implemented: for example, we could add XGM data for each pulse.

The initial script was also refactored a bit (mostly due to the fixes in the `read_geom()` function), but the output remains more or less the same.

The code is really specific for the European XFEL, therefore I think it might be merged _as is_ (no one else uses it). But the idea might be extended much further.


phyy-nx:
Hi, @kirienko, very excited to see this addition and I'm supportive of getting this PR merged.  I've skimmed it and I have a few requests up front:

1. Can you fast forward to the latest master commit and force push the branch?  That way I can test these changes using the latest sources.
2. The python2 syntax checks are failing.  cctbx still supports python2 due to it being a dependency for other packages.  Can you either remove the f-strings and type hints in `agipd_from_defs.py` and `read_geom.py` or add these files to `cctbx_project/.azure-pipelines/py2_syntax_exceptions.txt`?
3. The clutter checks are failing as well.  Can you run libtbx.find_clutter on euxfel/*.py and follow its instructions?

I can also do these three things if you'd prefer, but I'd need to push them to a new branch or do a PR into your branch.  Either would be fine with me.

I will then test this branch on SwissFEL files we've generated previously, and compare that our code interprets them the same way as it used to.  Also, since read_geom is imported into the PAL and SwissFEL modules which also create NeXus files, I'll test that those work the same too.  We don't have good nightly regression tests for this code at present, due to needing to run it on an actual dataset.

Finally, this PR adds HDRMX/definitions as a git submodule.  Two points here.  1) I don't think we've done that before in cctbx.  @bkpoon do you have an opinion on this?  An alternative could be to use curl or wget to get NXmx.nxdl.xml and use it directly, but I appreciate the niceness of using a git submodule instead. I see the submodule isn't cloned unless you [ask for it explicitly](https://github.com/cctbx/cctbx_project/pull/623/files#diff-0f9cabe74e760b5d25c535837023ae8a2b3ed390b2ddd82d2598de28599ed7d2R34), so it wouldn't affect our redistributed cctbx code.  2) Even if it's ok, are you sure you want to point to [HDRMX/definitions](https://github.com/HDRMX/definitions/tree/master) and not [nexusformat/definitions](https://github.com/nexusformat/definitions)?  I believe most of the work in HDRMX/definitions has been folded into nexusformat/definitions, and if there are things missing we should address them.

Thanks!

kirienko:
Hi @phyy-nx, many thanks for reviewing. I did those three things you had asked, and now all checks are okay.

Regarding the source of definitions: for me that's not that important, and if `nexusformat/definitions` now contain all ingredients then we can switch to that. But it might be done in a separate PR. I'll check if it works.

Also keep in mind that here we also added the data from XGM. And without this data the script will fail, since `total_flux` is a mandatory parameter, and we don't want to put one dummy value for all pulses while we have the real values in XGM (yes, we do unit conversion there).

xfel/euxfel/agipd_cxigeom2nexus.py
```diff
@@ -26,176 +28,315 @@
...
-  import sys
-  nexus_helper = agipd_cxigeom2nexus(sys.argv[1:])
-  nexus_helper.create_nexus_master_file()
+    """
+        output_file_name = os.path.splitext(self.params.cxi_file)[0] + "_master.h5"
+        f = h5py.File(output_file_name, "w")
+        entry = f.create_group("entry")
+        entry.attrs["NX_class"] = "NXentry"
+        # --> definition
+        definition_string = f"NXmx:{get_git_revision_hash()}"
```
phyy-nx, line 103: Hi, I ran the code on a .cxi file, and this line adds the NeXus definitions repo revision hash to the NXmx statement.  I haven't seen that before, and it does cause dxtbx.print_header to fail.  Is the hash needed?  The definition group schema is [here](https://manual.nexusformat.org/classes/base_classes/NXentry.html#nxentry-definition-field), and it has a couple attributes that could be used, such as the version attribute.  There is an NXDL_VERSION file in the definitions folder that could be picked for this.

phyy-nx:
Hi @kirienko, I've tested `agipd_cxigeom2nexus.py` from this PR on different cxi and vds datasets from EuXFEL using the AGIPD detector. For each I ran `dxtbx.print_header` to see if the headers are read the same and it seems they are (excepting the comment above on get_git_revision_hash which will need to be resolved).  So that's good.  I need to run it on SwissFEL data to test the changes to read_geom.py.  I'll do that next.

I'd like to also test your new code `agipd_from_defs.py` so I can see if the files are still readable by dxtbx, but I don't have the [extra_data module](https://github.com/kirienko/cctbx_project/blob/refactor-euxfel-agipd2nexus/xfel/euxfel/agipd_from_defs.py#L3).  Is this an EuXFEL module?  Can you include directions on how to run agipd_from_defs.py?

Thanks!

xfel/euxfel/agipd_cxigeom2nexus.py
kirienko: This is a very good question, and I'm glad that you've asked!
There are two answers:
1. No, it's not needed directly, and indeed it causes some other `cctbx` pieces of software to fail. But it's just because of the way `cctbx` is written, there's nothing technically wrong to store there anything beyond those four letters "NXmx".
2. But on the other hand, yes, it is needed if we want to be consistent with definitions! Because as soon as we stick to the submodule and to the creation of NeXus from definitions automatically, at that particular moment we need to know the version of definitions (referring to a particular commit, i.e. hash), not only a name to the repo (like `NXmx`). Storing this information in a separate file makes the whole idea of NeXus as a self-contained format useless.

That's why I think it's the best place to store the version, but it's not in the NeXus standard yet. So I'd like to ask you, as a member of the NeXus committee, to think about adding this to the format itself :smiley:

kirienko: I rolled it back for a time being.

kirienko:
> I'd like to also test your new code `agipd_from_defs.py` so I can see if the files are still readable by dxtbx, but I don't have the [extra_data module](https://github.com/kirienko/cctbx_project/blob/refactor-euxfel-agipd2nexus/xfel/euxfel/agipd_from_defs.py#L3). Is this an EuXFEL module? Can you include directions on how to run agipd_from_defs.py?

Yes, that's [our development](https://github.com/European-XFEL/EXtra-data), but it's available from `pip`, so just type `pip install extra-data` to get it in your environment.

phyy-nx:
Hi @kirienko, can you look at [this possible revision](https://github.com/nexusformat/definitions/compare/nxentry_gitcommit) to NXentry and NXsubentry that adds the git commit id as an attribute to the definition field?  Not quite what you had but I like it better than adding the commit id to the definition field itself as it avoids text parsing and breaking existing code that queries the definition field.  Do you have any suggestions?

I discussed this with the NIAC today and they agreed to consider the proposal.  This kind of change needs a NIAC vote, which will take a bit to set up.  After you and I have agreed on the set of changes to NXentry/subentry, I'll make a PR for them which will then get approved ASAP.  Reasonable?

Smaller changes typically get through without votes, but changing NXentry/subentry is a big change and needs wider consideration across the NeXus community.

None of this is a blocker to this PR though, right?

phyy-nx (approved):
Hi, I tested `swissfel/jf16m_cxigeom2nexus.py` since it imports read_geom and it gives the same NeXus files as it did before, so that's good to go.

Approving the PR at this point.  We can either merge it as is or wait for the definitions/revision stuff to go through, your choice.  Just let me know.

I'd like to squash merge the 39 commits, if that's ok.

Thanks!

kirienko:
Hi, sorry for the delay, I was on vacation. I'd prefer this to be merged (well, squashed). Any other changes could be done on top. of this. Thanks a lot for the review!

phyy-nx:
Sounds good.  Merging.

phyy-nx:
@kirienko see nexusformat/definitions#937

PR #629, opened 2021-07-03
start of type_energies for all atoms

nwmoriarty:
A way to get the type_energy out of the interpretation.

### DIFF
--- a/mmtbx/monomer_library/pdb_interpretation.py
+++ b/mmtbx/monomer_library/pdb_interpretation.py
@@ -2457,6 +2457,7 @@ class build_chain_proxies(object):
     if restraints_loading_flags is None: restraints_loading_flags={}
     self._cif = cif_output_holder()
     self.pdb_link_records = {}
+    self.type_energies = []
     self.conformation_dependent_restraints_list = \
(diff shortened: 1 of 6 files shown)


mmtbx/monomer_library/pdb_interpretation.py
```diff
@@ -2549,6 +2550,21 @@ def _get_next_residue():
...
+          name = atom.name.strip()
+          if mm.mon_lib_names is not None:
+            name = mm.mon_lib_names[i]
+          al = atom_dict.get(name.strip(), None)
+          if al:
+            self.type_energies.append(al.type_energy)
+          else:
+            self.type_energies.append(None)
+            print(atom.quote())
+            assert 0
```
olegsobolev, line 2566: Maybe put meaningful message in the assertion, like atom name so that when a user gets a crash here we'll get some more info straight in the bug-report.

olegsobolev (approved):
Otherwise looks fine.

mmtbx/monomer_library/pdb_interpretation.py
nwmoriarty: OK will change

PR #632, opened 2021-07-17
Fix a compiler warning, conversion-null

nksauter:
Return type must be void* to squash the conversion of NULL to
non-pointer type 'int'.


bkpoon:
Let's see if rebuilding on Windows from scratch will work.

PR #633, opened 2021-07-22
Ensure anomalous_probability_plot is picklable

rjgildea:
The previously locally-defined namedtuple result meant that the result of `iotbx.merging_statistics.dataset_statistics(..., anomalous=True, ...)` was not picklable.

Fixes xia2/xia2#611


bkpoon (approved):
Looks good! I guess this is a general issue with `namedtuple` and nesting classes during pickling.

The CI Branch failures can be ignored since all the jobs in Full are passing.

PR #635, opened 2021-07-26
Major update to the modify_cosym merging worker.

nksauter:
To determine symmetry alignment among crystals, the updated code can run jobs with a fixed MPI communicator size (and thus the same command line, e.g. mpirun -n32 cctbx.xfel.merge) for automatic data processing in a high throughput environment. Internally the program will use fewer ranks, so as to increase the tranch size to a pre-determined acceptable number of lattices (600 for photosystem I).

Larger tranch size increases the wall time O(N**2) but also the accuracy and completeness of lattice alignment to ground truth. Use compare_results.py to compare trial results to maximal-accuracy results obtained with mpirun -n1.  99.5% agreement is probably acceptable, 90% is not.

Requires a new phil parameter:
modify.cosym.tranch_size=600

Improvements to algorithm: 1) tranches are pre-planned with knowledge of the number of experiments input to each rank. Ranks are accessed in random order so that tranches are maximally balanced, even e.g., if the first-half of ranks have more experiments. 2) Rank==0 is reserved for the future plan of computing the anchor alignment in parallel with the tranch alignment.  3) Mutual tranch alignment requires a new algorithm, in fact it is the dimensional embedding algorithm of Brehm & Diederichs; in other words B&D is used for both lattice alignment and tranch alignment.

Known issues.  1) Worker is only tested for the use case of one merohedral operator; and will likely require additional code for the general case.  However there is no sanity-check to verify one operator.  2) A unit test is planned for df_cosym.py tranch alignment. 3) A multi-rank MPI regression test is planned for token_passing_left_right.py.


nksauter:
Of interest for the phil file:
modify.cosym.d_min=2.7 # going from 1.9 to 2.7 saves some wall time on ANCHOR.run()
merging.d_min=2.7 # saves time going from 1.9 to 2.7, as the model is computed for cosym anchor.  However, this also reduces the resolution of final merging output to 2.7 Angstrom.

dwpaley (approved):
discussed offline with @nksauter

PR #644, opened 2021-08-23
Suitename

russell-taylor:
This implements the MolProbity SuiteName program as a native CCTBX Python script under mmtbx rather than as a module pulled from a separate repository.  It has been tested against the original SuiteName running on all of the files in the PDB that have RNA scores.  For those files that mp_geo can run on, the scores are the same as the original except for "Banker's rounding" of the least significant digit in Python 3.6.

The new version has been tested for all the command-line arguments used by the MolProbity web site.

The new version has been tested with Python 2.7 and 3.6 on Linux.

The new version has a unit and regression test included in mmtbx/run_tests.py.

The new version is also able to directly load PDB and CIF files, removing the need to run mp_geo as a preprocess.


russell-taylor:
@bkpoon I need to fix a bug involving running the programs under Python 2.7 that does not show up in the unit tests (time for a new unit test!).  Please hold off on the merge until I fix that.

Thanks,
Russ

russell-taylor:
@bkpoon Okay, I've satisfied the lint checkers and also fixed the problem with the code not running in Python 2.7, so I think we're good to go if it looks good to you.

bkpoon (approved):
We should squash this one since there are merge commits. Is there anything else to be discussed next week before merging? I'll update the test in phenix_regression once this is merged.

russell-taylor:
Squashing and merging!  Thank you!

PR #652, opened 2021-09-06
Adapt libtbx/Sconscript for perlmutter

Trzs:
The sconscript requires a few changes to build on perlmutter


bkpoon (approved):

JBlaschke:
I don't know about the XFEL CI fail -- @bkpoon  this seems unrelated to this PR though, right?

bkpoon:
Yeah, the error is not related to this change. The failure seems to be intermittent, but mostly fails.

PR #666, opened 2021-10-06
Update CONTRIBUTING.md

clacri:
just corrected the originally word


Anthchirp:
I don't think this trivial change warrants 100 Azure builds :)

clacri:
Thank you, and sorry about that I did not realise that and went for the 'safe' option instead of committing to master. Thanks again!

PR #668, opened 2021-10-08
Trying to compile on gcc without warnings.

nksauter:
Three warnings can be safety suppressed, but it would be better if
the underlying code could be fixed so no gcc warnings are generated.


nksauter:
Would like to be able to compile all code with libtbx.scons without generating gcc warnings.  These three warning suppressions are currently needed (-Wno-deprecated-declarations -Wno-maybe-uninitialized -Wno-terminate").  Initiating a conversation in the hope that the underlying code can be fixed, so that these warning suppressions aren't needed.  Asking for developer review.

PR #677, opened 2021-10-21
Hkl info

Oeffner:
This branch adds the command line program cctbx.hklinfo which summarises properties of the miller_array objects in a reflection file and prints it out on the screen in a table format. 

Properties of interest can be displayed/omitted by setting PHIL parameter values as additional input.
The main part of the code is stored in iotbx\gui_tools\reflections.py which is also used by HKLviewer Together with the PHIL parameters this lets the user display or omit columns in the table of miller arrays of the loaded reflection file.

A regression test has been added which took some time to develop due to me fighting unicode characters in the output. Perhaps this could be done in a more robust way. Currently the test is only testing snippets of sca, hkl and cif reflection data embedded as strings in crys3d\tst_hklinfo.py. Test of mtz file format will be added when I work out how to access files in phenix_regression\reflection_files if present.

Suggestions/improvements are welcome.


olegsobolev:
I don't have any issues with this PR, but I'm not an expert on the data handling either.

pafonine:
Hi All,

looks good to me too. What's the purpose of " New commandline program 
HKLinfo .."? How it is different from phenix.reflection_statistics ?

To me HKLinfo seems very similar (purpose wise) to 
phenix.reflection_statistics : both show information about reflection 
file. If this is the case, I'd suggest adding unique functionality of 
HKLinfo to phenix.reflection_statistics (rather than creating new CL 
program).

Pavel

Oeffner:
> Hi All, looks good to me too. What's the purpose of " New commandline program HKLinfo .."? How it is different from phenix.reflection_statistics ? To me HKLinfo seems very similar (purpose wise) to phenix.reflection_statistics : both show information about reflection file. If this is the case, I'd suggest adding unique functionality of HKLinfo to phenix.reflection_statistics (rather than creating new CL program). Pavel
> […](#)

Point taken. I should have been aware of that program. HKLinfo is meant to give a summary of the contents of a reflection file in a tabular format where columns can be selected or deselected from phil input. The tabular format is used by the HKLviewer and I thought it would be nice to have it as an external CL program. So the code base of HKLinfo is shared with that. This is similar to some of the output from phenix.reflection_statistics. If I understand it correctly phenix.reflection_statistics prints out features such as twinning and tncs statistics as well as correlations between miller_array objects in addition to most of the features output by HKLinfo. 

I am unsure if it would be worthwhile merging HKLinfo into phenix.reflection_statistics. I suppose the tabular HKLinfo format could be an optional output format the user may choose. But the resulting program would be a two-headed beast emitting output either as already done for phenix.reflection_statistics or as tabular HKLinfo format if supplied with a command line option. In which case we might as well have it as a separate program. On the other hand it would be good to have common code base to avoid code duplication.

A separate issue is that the phenix.reflection_statistics program is not yet using the program template so that would also need to be addressed. 
Maybe we can talk about it at the office hours tomorrow.

pafonine:
phenix.reflection_statistics is old (written by Ralf) and so far AFAIK 
no-one adopted it to maintain. So no wonder it does not use modern 
utilities such as program template (I'm not even sure if it uses Phil!).

So if you are into this now, this may be an opportunity to do something 
about it.. For example, you can add to HKLinfo whatever statistics that 
phenix.reflection_statistics computes and then make 
phenix.reflection_statistics essentially be an alias for HKLinfo 
(running phenix.reflection_statistics would in fact run HKLinfo).

Having two programs that do essentially the same but slightly different 
may be confusing for users, complicates maintenance and likely increases 
code duplication.

Pavel

Oeffner:
Sounds good. Statistical properties not present in HKLinfo I'll migrate 
from phenix.reflection_statistics to HKLinfo. Hope to be done soon but 
until that has been completed the two programs will coexist. One thing I 
don't think should be migrated is the twinning and tncs analysis. These 
are already done more comprehensively by Xtriage. Likewise I don't see 
the importance of the calculation of pairwise  CC values between 
miller_array objects.

Rob

pafonine:
Hi Rob,

> Statistical properties not present in HKLinfo I'll migrate
> from phenix.reflection_statistics to HKLinfo.

Sounds great! Thanks!

> Hope to be done soon but
> until that has been completed the two programs will coexist. 

I think this is fine as long as they don't co-exist permanently -;)

> One thing I
> don't think should be migrated is the twinning and tncs analysis. These
> are already done more comprehensively by Xtriage. Likewise I don't see
> the importance of the calculation of pairwise  CC values between
> miller_array objects.

I fully agree!

Pavel

PR #686, opened 2021-11-12
Don't divide by zero if rotating by 0°

benjaminhwilliams:
The quaternion of a 0° rotation is, straightforwardly, 1.  Sometimes, since the axis direction for a 0° rotation is arbitrary, `axis_and_angle_as_unit_quaternion` may be called with an angle of 0 and a placeholder axis of (0, 0, 0).  In this case, return early to guard against a `ZeroDivisionError` when trying to normalise the axis.

### DIFF
--- a/scitbx/matrix/__init__.py
+++ b/scitbx/matrix/__init__.py
@@ -1693,7 +1693,7 @@ def exercise_1():
   assert approx_equal(x.angle(y, deg=True), 25.6685689758)
   assert approx_equal(y, (0.2739222799, -0.5364841936, -0.6868857244))
 
-  # Check that a rotation of 0° around a dummy axis of (0, 0, 0) is tolerated.
+  # Check that a rotation of 0 degrees around a dummy axis of (0, 0, 0) is tolerated.
   zero_rotation = col((0, 0, 0)).axis_and_angle_as_unit_quaternion(angle=0)
   assert approx_equal(zero_rotation, (1, 0, 0, 0))
   # Check that we can therefore do a round trip from R3 identity matrix to axis and
--- a/scitbx/matrix/__init__.py
+++ b/scitbx/matrix/__init__.py
@@ -452,6 +452,10 @@ class rec(object):
   def axis_and_angle_as_unit_quaternion(self, angle, deg=False):
     assert self.n in ((3,1), (1,3))
     if (deg): angle *= math.pi/180
+
+    if not angle % (2 * math.pi):
+      return col((1, 0, 0, 0))
+
     h = angle * 0.5
     c, s = math.cos(h), math.sin(h)
     u,v,w = self.normalize().elems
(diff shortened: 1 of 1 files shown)


benjaminhwilliams:
The test failure seems to be in `tst_bcr.py`, which seems unrelated to this PR and I notice that @bkpoon has since marked it as unstable on master, so I've merged master into this branch and hopefully the CI will now pass.

benjaminhwilliams:
I've also added a test that highlights how I discovered this problem.  I found some old code in DIALS that was effectively trying to do a round trip from rotation matrix to quaternion, to axis and angle, to quaternion, to rotation matrix.  When you try this, using the ℝ³ identity matrix as your starting point and without the changes in this PR, you get
```python
>>> from scitbx import matrix
>>> id = matrix.identity(3)
>>> q = id.r3_rotation_matrix_as_unit_quaternion()
>>> angle, axis = q.unit_quaternion_as_axis_and_angle()
>>> axis.axis_and_angle_as_r3_rotation_matrix(angle)
  File "/home/ekm22040/DIALS/conda_base/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3444, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-7cb59b38bfe8>", line 1, in <module>
    axis.axis_and_angle_as_r3_rotation_matrix(angle)
  File "/home/ekm22040/DIALS/modules/cctbx_project/scitbx/matrix/__init__.py", line 461, in axis_and_angle_as_r3_rotation_matrix
    uq = self.axis_and_angle_as_unit_quaternion(angle=angle, deg=deg)
  File "/home/ekm22040/DIALS/modules/cctbx_project/scitbx/matrix/__init__.py", line 457, in axis_and_angle_as_unit_quaternion
    u,v,w = self.normalize().elems
  File "/home/ekm22040/DIALS/modules/cctbx_project/scitbx/matrix/__init__.py", line 266, in normalize
    return self / abs(self)
  File "/home/ekm22040/DIALS/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in __truediv__
    return rec([e/other for e in self.elems], self.n)
  File "/home/ekm22040/DIALS/modules/cctbx_project/scitbx/matrix/__init__.py", line 154, in <listcomp>
    return rec([e/other for e in self.elems], self.n)
ZeroDivisionError: float division by zero
```

benjaminhwilliams:
I see there are now a couple of different failing tests, `scitbx.linalg.tests.tst_svd.exercise_inverse` and fails in one pipeline and `cma_es.tst_cma_es.test_cma_es_file` fails in another.  There doesn't seem to be any consistency in the failures between different pipelines.  I can't see how these failures are related to this PR and I can't reproduce them locally.  I don't really have much experience of the CCTBX CI set-up.  Are these just idiosyncrasies that can be ignored?

bkpoon:
Yep, those failures can be ignored. They fail less than 1% of the time, so I wouldn't consider them unstable. I will also generally re-run the failed jobs once the pipeline completes so that they are cleared.

Also, generally, you can ignore the results from the "CI branch" and "XFEL CI branch" pipelines since they run on the current state of your branch. The "Full" and "XFEL CI" pipelines are the ones that run on the merged code.

benjaminhwilliams:
It looks like I've got all the CI checks to pass now.  I'm not sure who's best placed to review this.  @bkpoon, any suggestions?

bkpoon:
I chatted with @nksauter and I can merge this.

benjaminhwilliams:
Thanks.

PR #687, opened 2021-11-19
Add documentation for debug builds and heap analysis

dwpaley:
* Instructions to build cctbx with debug symbols
* Steps for heap analysis in gdb-heap with example data


dwpaley:
@bkpoon you originally suggested dox/rst as a location for this document, but I'm still open to suggestions if you have any other idea.

bkpoon:
Are you running Docker so that you can use `gdb` on macOS? If you're on linux, `gdb` should just work, right?

dwpaley:
No, there were a few other considerations. Gdb-heap was a Fedora package
(although not available anymore) so that made the setup easier. Also wanted
root access for installing debuginfo packages. I’m sure this is all
achievable in a non-Docker environment, but not sure when I’ll have time to
go on another dependency adventure, this setup was a multi day project.

bkpoon:
Was going to say that the `Dockerfile` would be useful! Feel free to squash and merge.

dwpaley:
Thanks!

PR #688, opened 2021-11-23
Molprobity2 improvements

russell-taylor:
Features, bug fixes, and unit test for mmtbx/probe and mmtbx/reduce.


russell-taylor:
The only failing test is cctbx/maptbx/tst_bcr.py and it is finding a bad rpeak value.  None of the commits in this request could impact that, so I'm going to go ahead and merge under the assumption that the bug will be fixed separately.

PR #694, opened 2021-12-01
try_all_readers: skip cif reader unless extension is .cif

dwpaley:
As discussed on phenixbb: "iotbx.merging_statistics on Python 3: too memory consuming"

### DIFF
--- a/iotbx/reflection_file_reader.py
+++ b/iotbx/reflection_file_reader.py
@@ -126,6 +126,10 @@ def try_all_readers(file_name):
   except Exception: pass
   else: return ("shelx_hklf", content)
   try:
+    # The cif parser uses a lot of memory when reading a file with millions
+    # of words (like an xds_ascii file). Thus we only try the cif reader if
+    # the filename looks like a cif.
+    assert os.path.splitext(file_name)[1].lower() == '.cif'
     content = cif_reader(file_path=file_name)
     looks_like_a_reflection_file = False
     for block in content.model().values():


bkpoon:
We (@olegsobolev @nwmoriarty) chatted about this yesterday and we think this is a good idea.

pafonine:
Can it be a little smarter (at virtually little coding effort) and not 
only check the extension (.cif) but also look if the file looks like a 
cif file by, for example, reading a first few lines and looking for one 
or a few hints (like loop cards)? I mean... what if someone gives you an 
.mmcif file (I guess I've seen that before)?

Pavel

bkpoon:
That was part of the discussion as well. Do all CIF files have "data_" as the first few characters? For example, for PDB code 1yjp, the model file starts with "data_1YJP" and the data file starts with "data_r1yjpsf".

pafonine:
If we assume that we only supposed to deal with model, reflections and 
restraints CIF files, then I think they always have "loop_" cards. The 
handful of cifs I can at hand all have 'data_' as well, but I don't know 
if having 'data_' is mandatory in general. Perhaps CIF expecrts can 
comment more!

Pavel

dwpaley:
I think that is true although I'm not a cif expert. This page seems to
agree:
https://mmcif.wwpdb.org/docs/tutorials/mechanics/pdbx-mmcif-syntax.html

How about accepting any of the following:
-extension cif or mmcif
-first non-blank, non-comment line starts with data_
-any line starts with loop_

olegsobolev:
I would add a link to more formal description:
https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax

and suggest to check the following:
- extension is cif, mmcif, dic and all of them with possible addition of
.gz;
- first non-empty, non-comment line starts with NON case sensitive "data_"
allowing preceding whitespaces;

Looking for the "loop_" seems excessive since absence of data_ makes even
core CIF invalid.

pafonine:
Perhaps not only .gz, but also tgz, tar, tar.gz, zip, bzip, rar, bzip2, 
etc etc. I had an impression this is handled separately in cctbx.

Pavel

dwpaley:
> Perhaps not only .gz, but also tgz, tar, tar.gz, zip, bzip, rar, bzip2, etc etc. I had an impression this is handled separately in cctbx. Pavel
> […](#)

Not sure if I understand. On the master branch, if I call `any_reflection_file` on a `cif.gz`, `cif.tar.gz`, or `cif.bz2`, it fails to open with "invalid start byte". `.cif.tar` does open, understandably since it's just the same data plus a header. Do you just want to make sure that current behavior is preserved after this PR, or something else?

rowlesmr:
> I would add a link to more formal description: https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax and suggest to check the following: - extension is cif, mmcif, dic and all of them with possible addition of .gz; - first non-empty, non-comment line starts with NON case sensitive "data_" allowing preceding whitespaces.

To add to this, if the CIF file conforms to the [v2.0 specification](https://www.iucr.org/resources/cif/cif2), the very first token in the file must be `#/#CIF_2.0`, in addition to the first non-comment line starting with case-insensitive `data_`

PR #696, opened 2021-12-04
Kokkos incoming

nksauter:


nksauter:
Kokkos kernels now implement the logic in the original nanBragg-Mullen kernel

PR #698, opened 2021-12-15
Molprobity2 improvements2

russell-taylor:
Adds Probe2 vs. Probe regression test program along with command-line options for dumping atom type and radius information to be compared.

Marks Nitrogens in HET groups that have no bound Hydrogens as acceptors, bringing the Probe and Probe2 scores very close together for 4fenH.

Other features and bug fixes for Probe2 and Reduce2.

### DIFF
--- a/mmtbx/probe/CompareOriginalAndNewProbeAtomInfo.py
+++ b/mmtbx/probe/CompareOriginalAndNewProbeAtomInfo.py
@@ -17,7 +17,7 @@ from __future__ import print_function, nested_scopes, generators, division
 from __future__ import absolute_import
 
 import subprocess
-import os
+import os,sys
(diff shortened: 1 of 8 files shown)


russell-taylor:
So the syntax checks are failing in the main-branch code I'm trying to merge with.  I can't fix them in my pull request because they are not in my pull request to begin with -- running the checks on the pull request files shows no errors.

russell-taylor:
Ah, good -- it looks like someone else pulled a fix and it is re-running the checks.

bkpoon:
The PR will run the pipelines on the merged code so any recent changes will be present. However, the CI Branch and XFEL Branch pipelines are only for your branch (but those are passing anyway).

PR #699, opened 2021-12-15
Small mol updates

dermen:
Includes updates made to diffBragg while getting it to work on tethrene small molecule XFEL data

### DIFF
--- a/simtbx/command_line/hopper.py
+++ b/simtbx/command_line/hopper.py
@@ -222,7 +222,7 @@ def save_up(Modeler, x, exp, i_exp, input_refls):
     LOGGER = logging.getLogger("refine")
     Modeler.best_model, _ = hopper_utils.model(x, Modeler.SIM, Modeler.pan_fast_slow, compute_grad=False)
     LOGGER.info("Optimized values for i_exp %d:" % i_exp)
-    hopper_utils.look_at_x(x,Modeler.SIM)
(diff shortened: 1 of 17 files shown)


mewall:
Is this comment still true?


simtbx/diffBragg/refiners/stage_two_refiner.py:60

# how many parameters per shot, currently just scale and B-factor


And I'm not entirely sure what's going on with refine_Icell in the diffBragg kernel. When this flag is set, are the I(hkl) values being refined elsewhere, and the intensity values returned from the kernel not being scaled by the I(hkl) values? It might be good to add a comment explaining how this works.



From: Derek Mendez ***@***.***>
Sent: Wednesday, December 15, 2021 1:53:14 PM
To: cctbx/cctbx_project
Cc: Wall, Michael E; Review requested
Subject: [EXTERNAL] Re: [cctbx/cctbx_project] Small mol updates (PR #699)


@dermen<https://github.com/dermen> requested your review on: #699<https://github.com/cctbx/cctbx_project/pull/699> Small mol updates.

dermen:
@mewall , thanks, the comment is no longer true, I will update it.. 

As for refining I_cell, I was wanting to explore the scenario where I_cell is allowed to be negative (and then just use Wilson statistics to convert to F_cell , after running stage 2). In such a case diffBragg could potentially return a negative value for a pixel model .

mewall (approved):
Thanks for the clarifications

PR #702, opened 2021-12-16
Nasty workaround for DYLD_FALLBACK_LIBRARY_PATH

### LINKED ISSUE #701: DYLD_LIBRARY does not work as expected

graeme-winter:
Since thanks to SIP this is clobbered in any fork / exec scenario the SHELL
default value of this is lost in the dispatcher script which is usually
harmless but in some cases - as I found - harmful.

This restores /usr/lib to the cctbx default value of that path so ensures
libraries which are depended on are found. This is not a problem in the usual
Python / conda ecosystem so this nasty workaround is probably valid and
justified.

Fixes #701

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -1186,6 +1186,8 @@ Wait for the command to finish, then try again.""" % vars())
       p = op.join(sys.prefix, d)
       if (op.isdir(p)):
         result.append(self.as_relocatable_path(p))
+    if (sys.platform.startswith("darwin")):
+      result.append(absolute_path("/usr/lib"))
     return result
 
   def write_conda_dispatcher(self, source_file, target_file,


bkpoon:
A more detailed explanation is here

http://phenix-online.org/pipermail/cctbxbb/2021-December/002331.html

But the general solution is to define `DYLD_FALLBACK_LIBRARY_PATH` in the dispatchers by using the `dispatcher_include_template.sh` file in the `build` directory and running `libtbx.refresh` to regenerate all the dispatchers with the new setting.

The dispatcher template enables adding any paths and reordering paths, in necessary. This mechanism is used in HPC builds for using optimized libraries at runtime.

graeme-winter:
Worked around with the suggestion from @bkpoon  _but_ will leave issue open until I find a way which does not depend on end user rebuilding.

Issue #704, opened 2021-12-17
Eigen throwing errors during compilation


Trzs:
original [issue](https://github.com/JBlaschke/alcc-recipes/issues/7) reported by @dwpaley:

Im finding when I build cctbx, I encounter compilation errors due to CUDA + Eigen. I then have to replace the modules/eigen with an appropriate version from the gitlab repo, then re-run the build script (e.g. setup_perlmutter.sh). I usually comment out the mamba install steps prior to the build script re-run, as well as the hot update flags.

On a second note, it seems the most recent eigen commits in the gitlab repo also wont compile, however the commits from few months back do.

Trzs:
Fixed with 02cb04a5

PR #705, opened 2021-12-17
Bump Eigen to 3.4.0

Trzs:
Fix for #704 and [alcc-recipes #7](https://github.com/JBlaschke/alcc-recipes/issues/7)


bkpoon:
Sounds good, but let me add a check so that the older compiler uses 3.3.9. I think that's why I haven't bumped the version yet.

PR #713, opened 2022-01-06
Decouple cuda and kokkos build for Spock

Trzs:
Spock uses AMD MI100 cards, which require HIP. This PR changes the build script so kokkos can be build with the HIP backend independent of cuda.


nksauter (approved):

bkpoon (approved):
Once backwards compatibility is added to around line 3067 of `libtbx/env_config.py`.

PR #718, opened 2022-01-12
Modifications for Kapton Plugin: Part 2

voklejas2:
Adjust tolerance values in Kapton plugin These tolerances are associated with logic used to determine whether or not lines demarcating max/edge Kapton absorption intersect with detector panel(s). Tolerance values were increased by a factor of 3. This increase was chosen based on desired performance and ad hoc testing.

NOTE: This change will not work without pulling dials branch 'fix_max_edge_pt2'.

### DIFF
--- a/xfel/util/kapton_2019_correction_frame_frame_plugin.py
+++ b/xfel/util/kapton_2019_correction_frame_frame_plugin.py
@@ -535,8 +535,8 @@ class KaptonSettingsPanel(wx.Panel):
       """Takes in  x1,y1,x2,y2 and returns the panel where these points lie. If outside any panel then returns None"""
       x1,y1,x2,y2=elem
       # FIXME tolerance levels, a bit hacky
-      d1=1.03
-      d2=-0.03
+      d1=3.09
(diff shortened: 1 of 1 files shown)


voklejas2:
Examples of behavior for various values of Kapton tape rotation angles:

Angle = 40.2 degrees with single panel (desired behavior)
Screen Shot 2022-01-10 at 3 12 18 PM

Angle = 60.2 degrees with single panel (OK behavior, but the lines are drawn well past the boundary of the detector image)
Screen Shot 2022-01-10 at 3 11 07 PM

Angle = -60.2 degrees with single panel (OK behavior, but the lines are drawn well past the boundary of the detector image)
Screen Shot 2022-01-10 at 3 09 11 PM

Angle = -60.2 degrees with multi-panel (desired behavior):
Screen Shot 2022-01-10 at 3 07 02 PM

Angle = 60.2 degrees with multi-panel (desired behavior):
Screen Shot 2022-01-10 at 3 02 08 PM

phyy-nx (approved):

PR #719, opened 2022-01-14
Updated interface for vec2_double.

nksauter:
Add functions for each_normalize() and rotate_around_origin(),
especially useful for processing pixelwise image data.

### DIFF
--- a/scitbx/array_family/boost_python/flex_vec2_double.cpp
+++ b/scitbx/array_family/boost_python/flex_vec2_double.cpp
@@ -10,6 +10,7 @@
 #include <boost/python/make_constructor.hpp>
 #include <boost/python/args.hpp>
 #include <boost/python/return_arg.hpp>
+#include <boost/format.hpp>
 #include "flex_helpers.h"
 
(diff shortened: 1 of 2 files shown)


scitbx/array_family/boost_python/flex_vec2_double.cpp
```diff
@@ -256,6 +287,28 @@ namespace {
     return std::sqrt(sum_sq_(self));
   }
 
+  af::shared<vec2<double> >
+  each_normalize(
```
graeme-winter, line 291: I looked at this and thought `each_normalize` is an odd (non-obvious) name to use - straight `normalize` would probably be the name and behaviour I would expect. However I see this was what Ralf chose to use in 2009 for `flex.vec3_double`

I wonder though whether the method name here _should_ be `normalize` and we add an alias for `vec3` & deprecate the old name.

graeme-winter: Best guess is such a change may not be too egregious, as it is a little used function 

```
silver-surfer cctbx_project :) [master] $ git grep each_normalize .|wc -l
      14
```

PR #721, opened 2022-01-18
Adding intentional failures in probe and reduce tests to ensure these…

russell-taylor:
… failures are caught by the testing system during a pull request

DO NOT MERGE.  This is a test to ensure that we're catching failures.


bkpoon:
Tests in the `mmtbx` module are not currently run on Azure, but it something I want to add this month or next.

russell-taylor:
Okay, that explains why bugs are sneaking past my tests.

Very much looking forward to seeing chem_data and geostd in CCTBX!

PR #723, opened 2022-01-21
allows pdb input to diffBragg for Fcalcs

dermen:


dermen:
I modified `tst_nanoBragg_multipanel` so I could simply import the `dxtbx` crystal, beam, and detector models for use in other tests (as done in `tst_diffBragg_structure_factors.py`)

PR #725, opened 2022-01-22
Kokkos/CUDA equivalent interfaces for exascale_api

nksauter:
Each execution context now has two overloaded versions of exascale_api::add_energy_channel_mask_allpanel


Trzs (commented):
-- placeholder -

PR #738, opened 2022-02-25
Add the Planck constant and neutron mass to scitbx/constants.h

toastisme:
This just adds neutron mass and Planck's constant to `constants.h`.
Naming and values based on `scipy.constants`.


phyy-nx:
👍

PR #746, opened 2022-03-21
Diff bragg updates

dermen:
mostly updates revolving around per-spot refinement


mewall (approved):
>         use_exascale_api=params.predictions.method == "exascale",
>         use_db=params.predictions.method == "diffbragg",
>         show_timings=params.predictions.verbose,
>         quiet=not params.predictions.verbose,

for readability, recommend 

quiet=(not params.predictions.verbose)

PR #751, opened 2022-04-05
Add debranched version of nanoBraggSpotsKernel

Trzs:
This adds an optimized, debranched version of the nanoBraggSpotsKernel to simtbx/gpu. This allows better comparison between CUDA and KOKKOS now that both use a debranched Kernel.

Previously, CUDA was using the 'old' nanoBraggSpotsKernel from simtbx/nanoBragg.


Trzs:
Closed after discussion with @nksauter.

Trzs:
Closed after discussion with @nksauter

PR #753, opened 2022-04-10
xfel SCons: Don't overwrite dials helper vars if already set

ndevenish:
At the moment, these variables are defined in both XFEL and DIALS, because dials might be configured after XFEL in the module load order. This change continues the current behaviour, but in cases where DIALS has already set the variables, they will not be overwritten.


phyy-nx:
👍

ndevenish:
Errors appear to be unrelated python 2.7 errors (and are present on main).

PR #761, opened 2022-05-08
Add new constraint scalar_scaled_adp

dwpaley:
This takes a group of atoms and allows all their ADP components to
scale by a consistent factor. When applied to the whole structure,
the result is a single-parameter ADP scale for all atoms. This was
used to model radiation damage in:

https://doi.org/10.1002/anie.201806426


pcxod:
Sorry - no memory trace on this... Btw I have added similar scaled and rotated version in my fork...

PR #766, opened 2022-06-17
Rework prep_dataframe()

Trzs:
This PR replaces the randomized load distribution in prep_dataframe() with a more optimized one. The updated version cuts waiting times between ranks by 2 to 3.

### DIFF
--- a/simtbx/diffBragg/ensemble_refine_launcher.py
+++ b/simtbx/diffBragg/ensemble_refine_launcher.py
@@ -29,9 +29,9 @@ def global_refiner_from_parameters(params):
     pandas_table = pandas.read_pickle(params.pandas_table)
     LOGGER.info("EVENT: BEGIN prep dataframe")
     if params.prep_time > 0:
-        pandas_table = prep_dataframe(pandas_table, params.prep_time)
+        work_distribution = prep_dataframe(pandas_table)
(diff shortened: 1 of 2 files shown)


dermen:
This looks good! I had one question: this bit of code which seems to prep the weights array :

```
        print("Found %d shots" % nshots)
        high_shot = max([i_shot for i_shot,n in refls_per_shot])
        assert nshots == (high_shot+1)

        weights = [0] * nshots
        for i_shot,n in refls_per_shot:
            assert weights[i_shot] == 0
            weights[i_shot] = n
```

I think it could be trimmed down a bit if you want, but not necessary:

```
refls_per_shot = [(0,120), (2,240), (1,120)]  # e.g.


# new code to get weights list:
refls_per_shot = sorted(refls_per_shot)
indices, weights = zip(*refls_per_shot)
assert list(indices) == list(range(nshots))
```

dermen:
Also, the string formatting is not backwards compatible.. 

```
f"ERROR! Rank {COMM.rank} has no work_distribution!"
```
so that will need to be fixed before checking-in I suppose...

Trzs:
I never had the asserts fail, I mostly put it in as a sanity check in case MPI does anything funky. I'll trim it down 👍

PR #772, opened 2022-07-20
New ability to associate nanoBragg sources with

nksauter:
gpu structure factor amplitude channels. Extends the exascale API, only
available in Kokkos now.


bkpoon (approved):
The test failures are from downloading issues from anaconda.org.

Trzs (commented):
approved

PR #774, opened 2022-07-29
Lower hard-coded unit cell limit from 70 A^3 to 40 A^3

toastisme:
Indexing NaCl neutron data with DIALS gives a P1 volume lower than the current hard-coded threshold in rstbx. The highest symmetry solution is an almost perfect fit with the experimental result (5.64, see below), so fairly confident the P1 cell is valid.
img


nksauter:
Looks good.

PR #775, opened 2022-07-29
Updated parse_pae_file function

hlasimpk:
I've updated the parser to work with pkl files made in alphafold monomer_ptm/multimer runs and to work with the newly updated PAE format in the alphafold database v3.

### DIFF
--- a/mmtbx/domains_from_pae.py
+++ b/mmtbx/domains_from_pae.py
@@ -58,23 +58,31 @@ def parse_pae_file(pae_file):
 
         try:
             with open(pae_file, 'rt') as f:
-                data = json.load(f)[0]
+                data = json.load(f)
         except Exception:
             raise Sorry(f"Unable to read the json file: {pae_file}")
 
-        if 'residue1' in data.keys() and 'distance' in data.keys():
-            r1, d = data['residue1'], data['distance']
-            size = max(r1)
-            matrix = numpy.empty((size, size))
-            matrix.ravel()[:] = d
-        elif 'pae' in data.keys():
+        if isinstance(data, dict) and 'pae' in data:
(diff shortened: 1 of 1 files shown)


tristanic:
Ah... should have checked before I started on essentially the same thing (#776)! Two things you might want to cherry-pick from mine:

- ColabFold provides a PAE file with each individual model which is in a slightly different format again (simple dict, not list, containing {'pae': 2d pae array}). Added support for those.
- The new-style AlphaFold PAE matrices are rounded to the nearest integer, so now contain zeros where the previous minimum value was 0.2. Either the zero values need to be replaced with a small non-zero value (easiest), or the edge-weighting function in `domains_from_pae_matrix_networkx()` needs to be changed to avoid divide-by-zero.

hlasimpk:
@tristanic Thanks for the suggestions, i've added those in now

mmtbx/domains_from_pae.py
```diff
@@ -38,29 +38,62 @@
...
+    import numpy
 
-    matrix.ravel()[:] = d
+    ext = Path(pae_file).suffix
+    if ext == '.json':
+        import json
+
+        try:
+            with open(pae_file, 'rt') as f:
+                data = json.load(f)[0]
```
tristanic, line 61: To clarify, the ColabFold PAE files will fail with an `IndexError` here since they only contain a dict, whereas an AlphaFold PAE file contains a list with the dict as the first entry.

hlasimpk: I've fixed this and tested it against every PAE variant so ti should all be working now :)

tristanic: Looks good!

terwill: Thanks very much Adam and Tristan!

PR #776, opened 2022-07-29
Update to support new PAE JSON formats

tristanic:
You've undoubtedly seen yesterday's announcement by the AlphaFold team. Along with their 200-odd million (!) new structures they've made changes to the PAE JSON file format to make it a bit more compact:

- doing away with the mostly-superfluous "residue1" and "residue2" arrays;
- storing the values in the file as a 2D array;
- rounding all values to the nearest integer (we're in discussions with the EBI team on this one - it may be partially reversed so only values above 4 are rounded).

The proposed changes should allow loading of the PAE matrix from all three formats currently in circulation:

- the original AlphaFold format
- the format of the per-model PAE files provided by ColabFold (essentially the same as the new AlphaFold format, without the rounding)
- the new AlphaFold format

I'm afraid I don't have a cctbx  build environment currently set up for proper testing and I'm a bit rushed considering my upcoming move... @terwill would you mind giving it a quick test run? I'm afraid tst_domains_from_pae.py will probably need an update as well since (unless they go ahead and keep <4 values to existing precision) the results will undoubtedly be slightly different for the new PAE matrices - although surprisingly much **less** different than I originally feared, from the examples I've tried!

### DIFF
--- a/mmtbx/domains_from_pae.py
+++ b/mmtbx/domains_from_pae.py
@@ -43,17 +43,30 @@ def parse_pae_file(pae_json_file):
 
     try:
       with open(pae_json_file, 'rt') as f:
-        data = json.load(f)[0]
+        data = json.load(f)
     except Exception as e:
       raise Sorry("Unable to read the json file %s" %(pae_json_file))
 
-    r1, d = data['residue1'],data['distance']
+    if isinstance(data, dict) and 'pae' in data:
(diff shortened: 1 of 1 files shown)


tristanic:
Closing this, since it mostly duplicates #775 (and the parts that aren't duplicates are now merged in there).

PR #782, opened 2022-08-25
Molprobity2 debug5

russell-taylor:
Improvements to Probe2 and Reduce2.


russell-taylor:
So the syntax checks all worked, then the job restarted without me pushing any new commits.  They worked the second time, then the job restarted again without any pushed commits and reported a failure on files that I was not editing.  Guess I'll try rebasing it on master and see if that kicks it into working properly.

russell-taylor:
The unused import detector is finding errors in a file that I did not touch, so I commented it out and re-committed it to try and avoid the errors.  On my machine, find_unused_imports_crude is crashing on a UnicodeDecodeError but I hope that isn't happening on the build-check machine and it happens whether I leave the commented-out line in place or delete the whole line so it seems to be unrelated.

PR #791, opened 2022-09-15
Work around for ev11 crash

phyy-nx:
Work around for when ev11 gets a negative global_sum_of_delta_squared_in_bin for a given bin.

The test case I have shows that the refinement takes a bad step giving a high target function value, even if I throw out a bad bin.  It then recovers the minimization procedure:
```
5669.41
5669.41
2638.52
2571.27
2291.39
1115.96
2253525.53  <-- this is the bad step with the single bin that needs to be thrown out to avoid a math domain error
1115.78
2289.79
943.97
936.21
927.23
918.19
868.21
803.22
799.45
```

### DIFF
--- a/xfel/merging/application/errors/error_modifier_ev11.py
+++ b/xfel/merging/application/errors/error_modifier_ev11.py
@@ -124,7 +124,7 @@ class error_modifier_ev11(worker):
       global_sum_of_der_wrt_sadd_in_bin     = comm.reduce(flex.sum(sum_of_der_wrt_sadd_in_bin),   MPI.SUM, root=0)
 
       if self.mpi_helper.rank == 0:
-        if global_number_of_reflections_in_bin > 0:
+        if global_number_of_reflections_in_bin > 0 and global_sum_of_delta_squared_in_bin > 0:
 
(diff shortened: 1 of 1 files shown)


nksauter:
I have nothing useful to contribute here.  For my current project (p20231 data) the ev11 algorithm fails intermittently on Perlmutter regardless of whether this code diff is used.  In other words this pull request does not fix the issue I am investigating, and I have no data on whether the code diff improves anything.  For the record my symptom is an error message in the slurm log:

```
libfabric:2583:1667067712:cxi:core:cxip_ux_onload_cb():2259<warn> nid004075: RXC (0xe872:3:0): PtlTE 128LE resources not recovered during flow control. FI_CXI_RX_MATCH_MODE=[hybrid|software] is required.
srun: error: nid004075: task 67: Aborted
srun: launch/slurm: _step_signal: Terminating StepId=3520235.0
```

irisdyoung:
This diff prevents jobs failing for me at LCLS -- I vote to merge it.

phyy-nx:
Ok, i've seen it help in a bunch of cases, so we'll put it in.

PR #795, opened 2022-10-01
Reconcile SMV and CBF writers in nanoBragg.

nksauter:
Use an output scale producing visible Bragg spots. Output scale is applied as a function argument to SMV writer But with CBF it is a multiplier to the actual pixels Verify both files (CBF, double) and (SMV, int) have same pixel values to within rounding error.

Run the test with libtbx.python tst_nanoBragg_cbf_write.py
View the two result images with dials.image_viewer test_full_100.*


phyy-nx (approved):
> Run the test with libtbx.python tst_nanoBragg_cbf_write.py
> View the two result images with dials.image_viewer test_full_100.*

Done, looks good to me

PR #797, opened 2022-10-04
Stop importing pip, use subprocesses instead. Will close #710

### LINKED ISSUE #710: configure.py crashes connected to distutils and/or setuptools
Recent commits are hitting test failures for `xfel_psana_centos_7 python3.9` on step "Configure and build" with tracebacks that look like this:

<details> <summary> Traceback </summary>

```
  File "../modules/cctbx_project/libtbx/configure.py", line 34, in <module>
    if not run():
  File "../modules/cctbx_project/libtbx/configure.py", line 29, in run
    libtbx.env_config.cold_start(sys.argv)
  File "/__w/1/modules/cctbx_project/libtbx/env_config.py", line 3017, in cold_start
    env.refresh()
  File "/__w/1/modules/cctbx_project/libtbx/env_config.py", line 2251, in refresh
    module.process_libtbx_refresh_py()
  File "/__w/1/modules/cctbx_project/libtbx/env_config.py", line 2568, in process_libtbx_refresh_py
    exec(to_str(fh.read()), global_vars)
  File "<string>", line 24, in <module>
  File "/__w/1/modules/cctbx_project/libtbx/pkg_utils.py", line 54, in <module>
    import setuptools
  File "/__w/1/miniforge/envs/psana_env/lib/python3.7/site-packages/setuptools/__init__.py", line 8, in <module>
    import _distutils_hack.override  # noqa: F401
  File "/__w/1/miniforge/envs/psana_env/lib/python3.7/site-packages/_distutils_hack/override.py", line 1, in <module>
    __import__('_distutils_hack').do_override()
  File "/__w/1/miniforge/envs/psana_env/lib/python3.7/site-packages/_distutils_hack/__init__.py", line 73, in do_override
    ensure_local_distutils()
  File "/__w/1/miniforge/envs/psana_env/lib/python3.7/site-packages/_distutils_hack/__init__.py", line 61, in ensure_local_distutils
    assert '_distutils' in core.__file__, core.__file__
AssertionError: /__w/1/miniforge/envs/psana_env/lib/python3.7/distutils/core.py
```
</details>

It looks like the same issue discussed in https://github.com/pypa/pip/issues/10742 and https://github.com/pypa/setuptools/issues/2993. I don't have any real idea what's going on, so just flagging this for now.

dwpaley:
As discussed in: https://github.com/pypa/setuptools/issues/3297 it seems that importing pip is just not supported. Our use case was not really compelling, we were (1) checking the version and (2) using a `pip._internal.main` api which is a thin (and to be deprecated) wrapper over `pip._internal.cli.main.main` There's a clear discussion here: https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program of why using pip via `subprocess` is preferred. 

This is a draft because there's a second place in `libtbx/auto_build/install_base_packages.py` where `pip._internal.main` is used in a similar way.

### DIFF
--- a/libtbx/auto_build/conda_envs/cctbx.devenv.yml
+++ b/libtbx/auto_build/conda_envs/cctbx.devenv.yml
@@ -21,3 +21,6 @@ dependencies:
   - dials-data
   - pytest-mock
   - pytest-xdist
+
+  # extra
+  - packaging
--- a/libtbx/auto_build/conda_envs/cctbx.devenv.yml
+++ b/libtbx/auto_build/conda_envs/cctbx.devenv.yml
@@ -21,6 +21,3 @@ dependencies:
   - dials-data
   - pytest-mock
   - pytest-xdist
-
-  # extra
-  - packaging
--- a/libtbx/auto_build/conda_envs/cctbxlite.devenv.yml
+++ b/libtbx/auto_build/conda_envs/cctbxlite.devenv.yml
@@ -37,6 +37,7 @@ dependencies:
   # extra
   - libsvm_py
   - pytest
+  - packaging
 
   # docs
   - docutils
--- a/libtbx/auto_build/install_base_packages.py
+++ b/libtbx/auto_build/install_base_packages.py
@@ -15,6 +15,7 @@ import platform
 import sys
 import time
 import zipfile
+import subprocess
 from optparse import OptionParser
 
 if __name__ == '__main__' and __package__ is None:
@@ -644,9 +645,9 @@ Installation of Python packages may fail.
     if extra_options:
       assert isinstance(extra_options, list), 'extra pip options must be passed as a list'
     if download_only:
-      try:
-        import pip
-      except ImportError:
+      pip_version_cmd = ['python', '-c', 'import pip; print(pip.__version__)']
+      pip_version_result = subprocess.run(pip_version_cmd, stdout=subprocess.PIPE)
+      if pip_version_result.returncode != 0:
         print("Skipping download of python package %s %s" % \
               (pkg_info['name'], pkg_info['version']))
         print("Your current python environment does not include 'pip',")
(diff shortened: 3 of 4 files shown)


dwpaley:
In principle this is done, but I don't know how to exercise the `build_python_module_pip` method in `install_base_packages.py`...

bkpoon (approved):
The `install_base_packages.py` code is only really for building Python 2.7 dependencies and I doubt that it will work now since we do not specify the versions of every dependency. Probably, one dependency will be installed via `pip` which pulls in an incompatible dependency that we did not version.

But you can try to build a minimal set of Python 2.7 dependencies with
```
python bootstrap.py base --builder=cctbxlite
```

JBlaschke:
Hi Folks, I am still waiting to test this on Perlmutter (longer than expected maintenance), but I can say that it works on Cori.

JBlaschke (approved):
This looks like a run-of-the-mill replacement of the pip module with a sub-process call. The only concern RE HPC is if you tried that with MPI then you might get into trouble because MPI (and libfabric) are not considered fork safe, while subprocess will call `fork`. I don't think that this will be an issue for an installer however -- so I think it's safe to proceed.

Another thing to consider is that this might be called withing a workflow management tool. In that case we need to make sure that the workflow tool exposes the correct environment to all children. Most workflow tools use `fork` or `execve` so we should be safe there, and I would counsel vigilance in case we see the "wrong" `pip` being picked up.

JBlaschke:
@dwpaley for cleanliness have you or @bkpoon considered: https://github.com/amoffat/sh

JBlaschke:
@bkpoon I have a question about the most elegant way to tell `bootstrap.py` to use a specific branch: Right now I edit `boostrap.py` manually to add the `-b <branch name>` to the module. For someone who uses `boostrap.py` deploy and build in one single command, is there an argument that I can use to do straight to a specific module's branch?

PR #799, opened 2022-10-05
XFEL gui: cache database connections

dwpaley:
This was tested live for 1-2 days at SwissFEL with no apparent problems. Subjectively the GUI was muuuuuch more responsive. @phyy-nx I suspect our test coverage is on the light side here, how do you think we should proceed?

### DIFF
--- a/xfel/ui/db/__init__.py
+++ b/xfel/ui/db/__init__.py
@@ -18,7 +18,7 @@ def get_run_path(rootpath, trial, rungroup, run, task=None):
     p = os.path.join(p, "task%03d"%task.id)
   return p
 
-def get_db_connection(params, block=True):
+def get_db_connection(params, block=True, autocommit=True):
   if params.db.password is None:
     password = ""
   else:
@@ -29,7 +29,14 @@ def get_db_connection(params, block=True):
   sleep_time = 0.1
   while retry_count < retry_max:
     try:
-      dbobj=MySQLdb.connect(passwd=password,user=params.db.user,host=params.db.host,db=params.db.name,port=params.db.port)
+      dbobj=MySQLdb.connect(
+          passwd=password,
+          user=params.db.user,
+          host=params.db.host,
+          db=params.db.name,
+          port=params.db.port,
+          autocommit=autocommit
+      )
       return dbobj
     except Exception as e:
       retry_count += 1
--- a/xfel/ui/db/xfel_db.py
+++ b/xfel/ui/db/xfel_db.py
@@ -303,7 +303,7 @@ class db_application(object):
       assert val in ['execute', 'cache_commits']
     return super(db_application, self).__setattr__(prop, val)
 
-  def execute_query(self, query, commit = False):
+  def execute_query(self, query, commit=True):
     from MySQLdb import OperationalError
 
     if self.mode == 'cache_commits' and commit:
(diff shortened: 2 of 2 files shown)


dwpaley:
(see more discussion in commit messages)

phyy-nx:
@dwpaley from the commit, "There were cases where a query would modify the database but not commit the changes. If the connection was not reused, those changes were lost, as intended."  Can you provide an example?

dwpaley:
That's weird, nowhere in the code base do I see the hypothetical bug-triggering situation where a query is executed with `commit=False` but the connection is left in a modified state. During the last experiment there was a very reproducible bug that was fixed by adding the call to `rollback()`. I will have to go figure out the details.

dwpaley:
Ok, found it. 

- Start a GUI, turn on the run watcher and the job submitter. 
- Go to the jobs tab. Select a completed run and restart it.
- Without `rollback()`, the job is deleted but not resubmitted. When the job submitter is restarted, then the job is submitted.
- With `rollback()`, the job is deleted and resubmitted right away.

It also seemed that restarted jobs might not be appearing in the job monitor until the thread is restarted, but I'm less sure about that one.

dwpaley:
From discussion with @phyy-nx and research, the problem was not uncommitted changes, but instead unclosed transactions:

> The reason for this behavior is that inside a single transaction the reads need to be consistent. 

from: https://stackoverflow.com/questions/1617637/pythons-mysqldb-not-getting-updated-row

So we enable autocommit when opening the connection, resulting in 1 query = 1 transaction as we probably expected in the first place. An exception is made if the caller to `execute_query` specifies `commit=False`; in that case we always open a new connection with `autocommit=False`. We don't worry about `dbobj.rollback()`, which was a red herring; we also don't have to manually call `dbobj.commit()`.

phyy-nx:
@dwpaley is the dummy cursor still needed if the problem was autocommit?

dwpaley:
@phyy-nx Not really and if the calling code goes on and does something with the live cursor, things might even be less potentially confusing now with autocommit enabled. 

However to me it seems inconsistent if we say "database queries are handled through `db_application`, but actually, here's a cursor, do whatever you want". So with the dummy cursor I tried to resolve that inconsistency by emulating the cursor but without giving access to the database.

I don't really feel strongly about it and it's not essential for this PR, I'm happy to revert it, let me know what you prefer.

phyy-nx:
I think I'd prefer the dummy cursor taken out.  Cursors haver several [properties](https://www.tutorialspoint.com/What-is-MySQL-Cursor-What-are-its-main-properties) but the most important one is that they are read only.  So they don't give access to the database, instead they give access to query results, and can be more memory efficient in the case of large queries (fetchone vs. fetchall).  Not that we are using that capability at the moment, but I don't think we need to disable the rest of the cursor API here.

dwpaley:
Hmm, if MySQLdb cursors are read-only then I'm really missing something because the cursor executes the query. On the current master branch it looks like this:
```
cursor = dbobj.cursor()
cursor.execute(query)
if commit:
  dbobj.commit()
```

In principle the memory footprint issue makes sense, but apparently with a "regular" `MySQLdb.cursors.Cursor` the whole result is already in memory once `execute` has been called: https://askcodes.net/coding/cursor-fetchall---vs-list-cursor--in-python. If we want to save memory we can specify the `cursorclass` when opening the connection.

Still no objection to killing the dummy cursor, just want to be sure we're on the same page.

phyy-nx:
Ah ok!  Interesting!  Ok then, let's keep it!  Thanks!  Also, this really did improve the GUI performance.  Yay!

phyy-nx (approved):

PR #801, opened 2022-10-07
Generalize tst_exafel_api to both cuda and kokkos

nksauter:
Accordingly, retire tst_kokkos_lib in run_tests.py. Switch contexts with a command-line flag.


nksauter:
Created a pull request. Will also commit an additional unit test to exercise the full exascale api, adding it to the pull request before merging to master.

nksauter:
Will merge the pull request later today, if there are no further comments.

PR #802, opened 2022-10-12
libtbx_refresh runs with setuptools in "legacy_editable" mode

dwpaley:
In https://github.com/pypa/setuptools/pull/3265 (merged in setuptools 64), the default behavior of `pip install --editable` changed. This broke the libtbx.refresh scripts that previously placed an entrypoints file in `modules`, e.g. `modules/dxtbx/src/dxtbx.egg-info/entry_points.txt`.

We restore the legacy behavior by setting an environment variable `SETUPTOOLS_ENABLE_FEATURES=legacy_editable` before running the libtbx_refresh scripts. This allows using the current setuptools, otherwise we would have to pin it again to <=63.

### DIFF
--- a/libtbx/env_config.py
+++ b/libtbx/env_config.py
@@ -2260,8 +2260,21 @@ selfx:
 
       for path in self.pythonpath:
         sys.path.insert(0, abs(path))
+
+      # Some libtbx_refresh scripts do a `pip install --editable`. The default
+      # behavior was changed in setuptools 64 and currently we expect the old
+      # behavior. See: https://github.com/pypa/setuptools/pull/3265
+      sef_previous = os.environ.get('SETUPTOOLS_ENABLE_FEATURES')
+      os.environ['SETUPTOOLS_ENABLE_FEATURES'] = 'legacy_editable'
+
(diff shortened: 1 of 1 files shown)


ndevenish:
What has physically changed with the layout? It tries to put the .egg-info in a different place now?

The documentation changes seem to be pushing people towards using virtual environments. I know cctbx has resisted this for some time (hence the whole bespoke engineering around read-only installs), but is it going to end up being unavoidable if they drop the legacy setting?

dwpaley:
Hi Nick, the new editable install doesn't create egg-info directories:
```
dwpaley@dials:/dev/shm/dwpaley/20221012
$ libtbx.refresh
[...]
$ find . -name dxtbx.egg-info
[nothing]
$ find . -name entry_points.txt |grep dxtbx
./build/dxtbx/lib/python3.9/site-packages/dxtbx-3.12.dev0.dist-info/entry_points.txt
```
I guess a more maintainable fix would be to fix the python environment so the entrypoints files are discovered in their new location? @bkpoon mentioned possibly installing in `build` instead of `build/dxtbx` etc.

PR #805, opened 2022-10-18
Implement kokkos_gpu version of gpu test.

nksauter:
Within the exascale_api, most API functions are now subject to identical tests in both the cuda and kokkos_gpu contexts.  This commit adds a test for the add_background() function, tested with a multi-source beam model.

### DIFF
--- a/simtbx/kokkos/kokkos_ext.cpp
+++ b/simtbx/kokkos/kokkos_ext.cpp
@@ -120,6 +120,7 @@ namespace simtbx { namespace Kokkos {
              "to use for structure factor lookup.  If -1, skip this source wavelength."
              )
         .def("add_background", &simtbx::Kokkos::exascale_api::add_background,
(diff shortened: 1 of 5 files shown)


nksauter:
@Trzs, this test was intended to create a kokkos_gpu context for the add_background() API with an override_sources function argument.  Previously, this argument was not implemented in the kokkos kernel.  Currently, the value of override_sources is ignored, and the unit test fails because the image simulation has the wrong scale factor.  It would be quite easy to fix this with a few lines of code, but it is worth having an in-depth discussion about whether those lines should be in the kokkos kernel, or the surrounding C++ wrapper.

Trzs (approved):
I added the scaling to the kokkos kernels and with this change it passes the unit tests.

nksauter:
Tests also pass in my hands.

PR #808, opened 2022-10-21
Add button to start mysql server

voklejas2:
The desired behavior of this PR is to add the required UI features to enable a user to start a database server from the UI. The main changes are:

1.  Logic added to create new dialog boxes to guide user to add required info to start DB server in /ui/components/xfel_gui_dialogs.py
2.  Small changes  in /ui/command_line/ui_server.py to ensure backwards compatibility of ui_server.py
3. Add in more phil parameters in /ui/__init.py__ necessary to support ability to launch server from GUI
4.  Add args in `do_submit` function in command_line/cxi_mpi_submit.py to pass in log and err file names- helpful to diagnose potential db issues
5. Added new method to class QueueInterrogator in /components/submission_tracker.py to retrieve name of host running server on NERSC and write to phil parameters
6. Added args for log and err files to `do_submit` function call in /ui/db/job.py due to 4 
7. Add function to process extra args added to phil parameters that was previously absent in `get_local_submit_command` class in /util/mp.py
  
Desired behavior:
To start the DB server from GUI.

1. Open the xfel GUI. For testing on NERSC, I found that it was necessary to specify the location of environment setup script as well as the give extra arguments that specify job time, account, CPU constraint argument and qos argument

Screen Shot 2022-11-14 at 11 31 10 AM

2. Access the new dialog boxes via the `DB Credentials` button that pops up in the very first dialog box that appears upon start-up of the GUI.

3. Press the `Start DB` button on the `DB Credentials` dialog box
Screen Shot 2022-11-14 at 9 54 00 AM

4. Enter  password for MySql server root and location of MySql DB and press OK

Screen Shot 2022-11-14 at 11 39 52 AM

6. Job to launch the server should be submitted to the queue. The root password will be deleted from phil parameters and the script used to launch the start server job will be removed for password security. 

The `ui_server.py` script should function just as it had previously.

### DIFF
--- a/xfel/command_line/cxi_mpi_submit.py
+++ b/xfel/command_line/cxi_mpi_submit.py
@@ -283,8 +283,15 @@ def get_submission_id(result, method):
     print('Submission', 'ID', 'is', submission_id)
     return submission_id
 
-def do_submit(command, submit_path, stdoutdir, mp_params, job_name, dry_run=False):
-  submit_command = get_submit_command_chooser(command, submit_path, stdoutdir, mp_params, job_name=job_name)
+def do_submit(command, submit_path, stdoutdir, mp_params, log_name="log.out", err_name="log.err", job_name=None, dry_run=False):
+  submit_command = get_submit_command_chooser(command,
+                                              submit_path,
+                                              stdoutdir,
+                                              mp_params,
+                                              log_name=log_name,
+                                              err_name=err_name,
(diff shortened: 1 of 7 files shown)


voklejas2:
I could use some help testing this because I don't have access to the real time queue and the gui just crashes if the gui isn't launched immediately

phyy-nx (approved):

PR #810, opened 2022-10-25
Diffbragg fhkl and save

dermen:
Lots of changes to the hopper_utils Python API, and extension of the C++/CUDA API. Original stage2 script pertaining to exascale work should not be affected by this. 

It was my intention that current codes are unaffected by this merge... However, if your code uses the output of the `hopper_utils.refine` method, then maybe slight changes will be needed. 

I plan to squash these commits before merging.  I think the most effective way to get this merged is if everyone can test their codes, in particular exaFEL. I'm happy and eager to help!


nksauter:
Let's make sure this is rebased!

dermen:
Is anyone using `hopper_process`? This PR might cause issues with hopper_process (not too hard to fix) as I don't routinely test it ..

nksauter (approved):
I've run the diffs through my current developmental project (SPREAD processing of LY99 data).   My project heavily uses the diffBragg.cpp code.  I see no immediate issues that would preclude merging the pull request.

simtbx/diffBragg/stage_two_utils.py
```diff
@@ -110,25 +111,43 @@ def PAR_from_params(params, experiment, best=None):
...
                                             sigma=params.sigmas.RotXYZ[i],
                                             center=0, beta=params.betas.RotXYZ)
 
-        # TODO: diffuse scattering terms, eta terms
+        PAR.eta[i] = ParameterType(init=init_eta[i], minval=eta_min[i],
+                                       maxval=params.maxs.eta_abc[i], fix=params.fix.eta_abc,
+                                       sigma=params.sigmas.eta_abc[i],
+                                       center=params.betas.eta_abc[i], beta=params.betas.eta_abc[i])
+
+        # TODO: diffuse scattering terms
```
mewall, line 150: What is needed here to support diffuse scattering?

mewall (approved):
I confirmed that the code still builds and runs simtbx.sim_view including diffuse scattering.

simtbx/diffBragg/stage_two_utils.py
dermen: actually, there is a new method in `hopper_utils.DataModeler` called `set_parameters_for_experiment` that handles diffuse models. moving forward, the hope is we can just use that method for both stage1 and stage2  and move away from this method.

PR #818, opened 2022-11-03
More correct XFEL plot_uc_histogram

Baharis:
Slightly modified the `ucell_tmp.png` file displayed in CCTBX.XFEL GUI, tab "Unit Cell". Discovered problems and ways in which they were addressed has been discussed below:

1. X axes on the top row 2D histograms now inherit X ticks and range from X axes on middle row 1D histograms to increase consistency/readability.
2. X axes tick labels on the middle row histograms used to be truncated to one decimal place. This decreased overlap between labels, but sometimes resulted in incorrect values (eg. 9.25 displayed as 9.2). Instead, number of X tick labels was limited to 4.
3. Y axis on the bottom row 1D angle histograms was previously missing "Number of images" label. This label was added.
4. 1D histograms used to feature edge-value data, which was correct only for the last displayed tag and usually overlapped with the edge of the screen or other histograms. Since lack of relations between tags does not allow to calculate edge-values for joined data, this information is now not displayed under histograms if more then one tag is being displayed.
5. Information on angle distribution was missing/misleading whenever more than 1 tag was visualized. To fix it, the three keys for a, b, & c were merged and additionally supplied with angle distribution data. Common information is displayed on the bottom of the key only once to save space.

This PR should not affect any functionality, but I would like some of the users to try out the modified style in order to assert whether reorganizing this information does not decrease readability.


Baharis:
Fixed a missing parameter and re-added the `extra_title` argument to `plot_uc_histogram`, this time functional.

phyy-nx (approved):

PR #825, opened 2022-12-05
update rayonix trusted values to match convention in cctbx/dxtbx#536

dwpaley:
After merging this, the min_trusted_value set here:

https://github.com/cctbx/dxtbx/blob/450277d94596c980762ae30f58fa357a37a1d642/src/dxtbx/format/FormatXTCRayonix.py#L71-L74

will be correct and the saturated_value can be corrected to the new `rayonix_max_trusted_value`.


dagewa (approved):
Looks correct to me :+1:

phyy-nx (approved):

PR #828, opened 2022-12-09
Unpin everything

dwpaley:
The psana=4.0.47 and h5py<3.2 pins are mutually incompatible, apparently related to the packages zlib/libzlib. The h5py issue in h5py/h5py#1880 and conda-forge/h5py-feedstock#92 is no longer reproducible. The problem with C++ compiler standard for boost 1.78 was resolved by 47c364ac, meaning that psana can now be unpinned. Therefore it appears we can unpin everything and monitor for regressions when using h5py.

Co-authored-by: Aaron Brewster <asbrewster@lbl.gov>


dwpaley:
This is not ready yet. Needs some corresponding changes in the xfel bootstrap builder.

PR #830, opened 2022-12-13
Add nxmx to conda dependencies

rjgildea:
See cctbx/dxtbx#597


phyy-nx (approved):

phyy-nx:
(Note, XFEL CI builder is still borked, so it passing isn't a blocker here)

phyy-nx:
(Note, XFEL CI builder is still borked, so it passing isn't a blocker here)

PR #832, opened 2022-12-15
implement one-item slices for miller sets and arrays

dwpaley:
The Miller set/array slicing syntax `thing[i]` is currently forbidden, but I don't see why.

### DIFF
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -40,10 +40,17 @@ fp_eps_double = scitbx.math.floating_point_epsilon_double_get()
 
 generate_r_free_params_str = r_free_utils.generate_r_free_params_str
 
+def _safe_slice(slice_object):
+  if type(slice_object) == int:
+    return slice(slice_object, slice_object+1)
+  elif type(slice_object) == slice:
+    return slice_object
+  else:
+    raise TypeError('bad slice: {}'.format(slice_object))
+
 def _slice_or_none(array, slice_object):
-  assert type(slice_object) == slice
   if (array is None): return None
-  return array.__getitem__(slice_object)
+  return array.__getitem__(_safe_slice(slice_object))
 
 class binner(ext.binner):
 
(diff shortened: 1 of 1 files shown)


dwpaley:
@rwgk added this assertion in b939b1d39cb and might have some insight but I assume this is just a longstanding NotImplementedError...

rwgk:
> @rwgk added this assertion in [b939b1d](https://github.com/cctbx/cctbx_project/commit/b939b1d39cbc78ef6e0d791d2ac5bb79cf581a97) and might have some insight but I assume this is just a longstanding NotImplementedError...

I don't remember TBH, but that's a fair assumption.
The only thing I kind-of remember: I always wanted to make flex array slicing more complete, but never got to it.

dwpaley:
I don't know if this is really right. If `arr` is a Miller array, this makes `arr[i]` an alias for `arr[i:i+1]`. It's not really consistent with slicing other iterables where `thing[i]` is a single item and `thing[i:i+1]` is an iterable of 1 item. I don't know if there is a logical way to maintain that distinction for a Miller array/set which is more like a container of several iterables.

Happy to leave this PR here to ponder, but I can appreciate why the current way has been here for 20 years :)

rwgk:
> I don't know if this is really right. If `arr` is a Miller array, this makes `arr[i]` an alias for `arr[i:i+1]`. It's not really consistent with slicing other iterables where `thing[i]` is a single item and `thing[i:i+1]` is an iterable of 1 item. I don't know if there is a logical way to maintain that distinction for a Miller array/set which is more like a container of several iterables.

Whatever my line of thinking was back then, time has completely erased it.

Maybe relevant:

`miller.array` is not like a (e.g.) Python `list` in that there is no type for `arr[i]`.

That in turn goes back to the decision not create `flex` arrays for e.g. `struct { tiny<int, 3> index; double datum; double sigma; }`, which would be nicest, but also extremely expensive in terms of compile times and Python extension sizes (the `flex` array wrappers accounted for a large fraction of those when I stopped working on cctbx in 2012; I don't know if that's still true).

My nearly context-free feeling: it's probably best to not add `miller_array_instance[i]`, unless there is an important use case. I'd almost certainly not want to make it a shortcut for `arr[i:i+1]`, because I believe that's probably more confusing (to readers of the Python code) than helpful. The alternative would be to make it a shortcut for `(arr.index()[i], arr.data()[i], arr.sigmas()[i])`, maybe as a `namedtuple`, but will that make client code simpler or clearer?

dwpaley:
> probably best to not add miller_array_instance[i], unless there is an important use case.

There certainly isn't. I was just trying to inspect elements while debugging and :
```
arr.indices()[i]
arr.data()[i]
arr.sigmas()[i]
```
was a lot of typing. A convenience method `miller_array_instance.inspect_element(i)` returning a namedtuple would be a good alternative.

PR #834, opened 2022-12-22
Sort jobs correctly in cctbx.xfel jobs tab

Baharis:
The aim of this PR is to introduce a correct numeric sort for columns "Job" and "Task" in cctbx.xfel GUI. These columns display (usually) integers as strings, which according to default sort result in `- < 12 < 123 < 2 < 23`. This is clearly expected, but unintended.

With my limited experience I couldn't find build-in natural column sorter, but I wrote a custom one which handles the issue using try-except statements. This sorter assigns unconvertable strings such as ` ` or `-` value of -1073741820 (slightly shy off half the minimum 32-bit integer, -1e99 caused SegFault). This new sorter acts on columns "Job" and "Task", and does not appear to be slower than the built-in one, although the largest table I could test on had barely 1000 rows.

Additionally, I replaced some mutable default parameters in `cctbx_project/xfel/ui/components/xfel_gui_controls.py` with their immutable and fixed some typos. [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) and some very confusing errors. They have their niche use, but I highly doubt anyone wanted the utilize it in the GUI.

A word of warning: when testing this code today on Perlmutter for ~an hour, at some point my GUI crashed with `BusError` and removed my `.cctbx.xfel/settings.phil` in the process. I do not think this was caused by my changes, but if anything like that happens in near future, then the devil might be somewhere here?

### DIFF
--- a/xfel/ui/components/xfel_gui_controls.py
+++ b/xfel/ui/components/xfel_gui_controls.py
@@ -8,7 +8,6 @@ Last Changed: 06/03/2016
 Description : XFEL UI Custom Widgets and Controls
 '''
 
-import math
 import os
 import wx
 import wx.lib.agw.floatspin as fs
@@ -649,7 +648,7 @@ import wx.lib.mixins.listctrl as listmix
 
 class SortableListCtrl(wx.ListCtrl, listmix.ColumnSorterMixin):
   def __init__(self, parent, style=wx.LC_ICON):
-    self.numeric_columns = set()
+    self.integer_columns = set()
     self.parent = parent
     self.sortable_mixin = listmix
     wx.ListCtrl.__init__(self, parent, style=style)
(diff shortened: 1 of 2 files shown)


Baharis:
A little question: is there any scenario where `Subm ID` in Jobs tab is not an integer (formatted as string)? For example can it assume value 'j123456' or '123.456' ? If not, I'd suggest sorting `Subm ID` using the new integer sorter as well, as it puts `None`s at the end. As a result, sorting by `Subm ID` would give newest jobs (instead of `None`s) at very top.

irisdyoung:
For the foreseeable future, submission ID will always be an integer -- that's what SLURM uses and that seems to be industry standard right now. You could also use separate logic for `None`s if you want. Note that we get `None` when the GUI has attempted job submission but SLURM hasn't responded with info about the queued job yet, so we will always have to be prepared for `None` values. 

While I'm theoretically in agreement re: python mutables, practically speaking some cases are fraught and some are relatively benign. I'm not worried about most of the places we use [] where () would better reflect the intended logic (although I also don't mind you bringing them into better compliance). There are some much more devious situations elsewhere in the ui code, which I highly endorse fixing! I haven't looked over your individual fixes yet.

Baharis:
I've been thinking about mutable variables in the GUI and one bug that might be caused by it. When you shift-click multiple runs in the list to select them, sometimes significantly more runs are selected then expected. If the selecting function uses mutable variables, then it's going to remember previous selections and select more and more lines each time. Something along this lines happened to me a few times, and it is benign - until you accidentally delete some of the wanted runs. I do not think this PR should fix this issue, neither am I sure my prediction is correct - just saying sometimes these benign bugs end up being very unfortunate, and difficult to trace origin later on.

Baharis:
Also, added `Sumb ID` to `self.job_list.integer_columns`.

irisdyoung:
I think the job selection bug is different -- there is dependence on the indices of the rows, and it's very easily corrupted when new rows are added (constantly!). I had thought the pause feature provided a workaround here, although it doesn't address the underlying issue. In any case I think that's deeper than using mutable objects. I would definitely not consider that a benign bug.

phyy-nx (approved):

Baharis:
@irisdyoung Regarding the subject of submission ID: after running some more jobs I have learned, that while Submission ID is an integer, sometimes the xfel GUI attributes two IDs to the same job, which is then formatted and presented in the table as "123456,123457". Using the new sorter, this will be placed at the end of list instead of being placed more intuitively between "123455" and "123458". As a result, I will modify this fragment back to use the string-sorter for the "Subm ID" column.

PR #837, opened 2023-01-03
Use std::shared_ptr instead of boost equivalent

rjgildea:
Required for dials/dials#2310

### DIFF
--- a/.azure-pipelines/xfel/download-source.yml
+++ b/.azure-pipelines/xfel/download-source.yml
@@ -25,10 +25,6 @@ jobs:
       git clone https://github.com/nksauter/LS49.git
       git clone https://gitlab.com/cctbx/ls49_big_data.git
       git clone https://github.com/lanl/lunus.git
-      cd dials
-      git checkout boost-c++11
-      cd ../dxtbx
-      git checkout boost-c++11
     displayName: Download sources
 
   - script: |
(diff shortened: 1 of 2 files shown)


phyy-nx:
Thanks @rjgildea.   Not sure why XFEL CI is failing in the compile step there.  Will investigate.

rjgildea:
> Not sure why XFEL CI is failing in the compile step there.

It's because this change isn't valid in isolation without the [corresponding change](https://github.com/dials/dials/pull/2310/files#diff-cd5b097f967b686571bab3ebd6c93a2f496bc42523e62414d31047eb028351dcR90) in dials/dials#2310, i.e. the type of [this variable](https://github.com/cctbx/cctbx_project/pull/837/files#diff-58c1fa64e341cdd366cddf42b1a638383ccfe13c484c6e5b5ee8daa017a99e40R112) must match the [return type](https://github.com/dials/dials/pull/2310/files#diff-cd5b097f967b686571bab3ebd6c93a2f496bc42523e62414d31047eb028351dcR90) of ``reflection_table.experiment_identifiers()``.

phyy-nx:
oh got it!  i'll test with pulling in the same branches as those prs

phyy-nx:
Ok, it compiled on my local sources, so I've added a commit to this `boost-c++11` branch for XFEL CI to pull the dials and dxtbx `boost-c++11` branches.  If XFEL CI passes, I'll revert that commit and this should be good to go.

rjgildea:
> I've added a commit to this `boost-c++11` branch for XFEL CI to pull the dials and dxtbx `boost-c++11` branches

Thanks for the idea! I've added similar to dials/dials#2310 to [checkout the relevant cctbx_project branch](https://github.com/dials/dials/pull/2310/commits/18bb429478a90760c1c21a526058b30c7f4f6cfe).

PR #838, opened 2023-01-03
HKLviewer regression tests

Oeffner:
These code changes comprises regression tests of HKLviewer developed to run remotely on azure pipelines. The code mitigates that the websocket protocol can be unstable (and fail to establish connection between webbrowser and python) by running tests repeatedly up to 4 times until they pass.
One test consists of loading the file iotbx/regression/data/phaser_1.mtz, displaying amplitudes expanded to P1 with Friedel pairs, slicing reflections for K=-9 and comparing the displayed reflections with a list of expected reflections.
Second test consists of loading iotbx/regression/data/phaser_1.mtz, calculating F/SigF values, sorting reflections into bins according to F/SigF values, hiding reflections with F/SigF > 1, expanding to P1 with Friedel pairs, slicing reflections for  L=9 values and again comparing visible reflections with what is expected.
These tests are run both with the HKLviewer Qt5 gui as well as directly from python using an installed webbrowser on the system, say firefox.

### DIFF
--- a/.azure-pipelines/azure-pipelines-build-cache.yml
+++ b/.azure-pipelines/azure-pipelines-build-cache.yml
@@ -27,7 +27,6 @@ stages:
   displayName: Download
   jobs:
     - template: ./download-source.yml
-    - template: ./xfel/download-source.yml
     - template: ./download-conda.yml
 
 - stage: Build
@@ -38,8 +37,3 @@ stages:
       parameters:
         modules: modules
 
-- stage: Build_xfel
-  displayName: Build and test XFEL
-  dependsOn: Preparation
-  jobs:
-    - template: ./xfel/xfel-ci-build.yml
--- a/.azure-pipelines/azure-pipelines-build-cache.yml
+++ b/.azure-pipelines/azure-pipelines-build-cache.yml
@@ -27,6 +27,7 @@ stages:
   displayName: Download
   jobs:
     - template: ./download-source.yml
+    - template: ./xfel/download-source.yml
     - template: ./download-conda.yml
 
 - stage: Build
@@ -37,3 +38,8 @@ stages:
       parameters:
         modules: modules
 
+- stage: Build_xfel
+  displayName: Build and test XFEL
+  dependsOn: Preparation
+  jobs:
+    - template: ./xfel/xfel-ci-build.yml
--- a/crys3d/run_tests.py
+++ b/crys3d/run_tests.py
@@ -5,7 +5,7 @@ import libtbx.load_env
 
 tst_list = [
 "$D/regression/tst_hklinfo.py",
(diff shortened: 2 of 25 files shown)


.azure-pipelines/azure-pipelines-build-cache.yml
```diff
@@ -27,7 +27,6 @@ stages:
   displayName: Download
   jobs:
     - template: ./download-source.yml
-    - template: ./xfel/download-source.yml
```
bkpoon, line 30: Changes to this file should be reverted

crys3d/run_tests.py
```diff
@@ -3,16 +3,55 @@
 from libtbx import test_utils
 import libtbx.load_env
 
-tst_list = [
-"$D/regression/tst_hklinfo.py",
-#"$D/regression/tst_HKLviewer.py",
-]
+
+if "darwin" in sys.platform:
```
bkpoon, line 7: You can also build up the test lists instead of copying
```python
tst_list = [ "$D/regression/tst_hklinfo.py" ]
tst_list_expected_unstable = [
   # fails sometimes due to websocket connection problem to webbrowser
   "$D/regression/tst_websocket.py",
   "$D/regression/tst_HKLviewerOSbrowserSliceK-9.py",
   "$D/regression/tst_HKLviewerOSbrowserBinFSigF.py",
]
other_tests = [
  "$D/regression/tst_HKLviewerQtGuiSliceK-9.py",
  "$D/regression/tst_HKLviewerQtGuiBinFSigF.py"
]
if sys.platform == "darwin" or sys.platform == "win32":
  tst_list.extend(other_tests)
else:
  tst_list_expected_unstable.extend(other_tests)
```

.azure-pipelines/azure-pipelines-build-cache.yml
Oeffner: should be fixed in latest commit

crys3d/run_tests.py
Oeffner: done

bkpoon:
Do you think something like `geckodriver` (https://github.com/mozilla/geckodriver) would work? It seems to enable some sort of web browser testing on Azure, but I have never used it. The default images have it and something similar for the other browsers. We don't have to sort it out for this PR, though.

Feel fee to "Squash and merge" when you're ready. Double check the commit message, though.

bkpoon:
Or "Rebase and merge". It looks like you've squashed a bunch of commits already.

Oeffner:
Done. I'll take a look at geckodriver and see what it can do.

Oeffner:
I could comment out testing hklviewer on linux for now until we have a working setup to reduce noise in the log file during regression testing.

Rob

PR #844, opened 2023-01-28
Xfel striping fix

Baharis:
Introduce some changes to striping to improve current behavior. In particular, currently `cctbx.xfel.stripe_experiment` stripes/chunks expt and refl files instead of actual experiments and reflections. However, since I have prepared the implementation in hurry, "striping" does not stripe anymore, but rather distributes randomly. Also, some improvements to the default parameters have been suggested.


Baharis:
Disclaimer: I might be unable to find time in the nearest future to fully fix the striping (i.e. read data and stripe it instead of files), but I believe some of the changes suggested here could be useful.

Baharis:
For anyone wandering about implementing real striping experiment-wise: since the striping/chunking script accepts and returns expt/refl *paths* and not individual experiments + associated reflections, one cannot patch-in real experiment-wise striping without significantly changing the workflow of routines which use it to handle actual objects instead of paths.

PR #845, opened 2023-01-30
Add a decorator libtbx.mpi4py.mpi_abort_on_exception

dwpaley:
This PR addresses the situation discussed here: https://github.com/dials/dials/pull/2311#discussion_r1087226198 where a single rank crash (or exit) can cause the MPI job to hang indefinitely. Decorated functions will catch Exception or SystemExit, print the traceback or error message, and call MPI.COMM_WORLD.Abort. The new code will be tested in xfel_regression: https://gitlab.com/cctbx/xfel_regression/-/blob/0dd7f6b44b289d3d01a22cc92c91209a66a42ec2/test/util/tst_mpi_abort.py and this branch is checking out the corresponding test.


phyy-nx (approved):
Looks like it handles all of the use cases, and I like this explicit mode requiring a decorator more than the implicit mode of changing sys directly.  Thanks.

PR #785, opened 2022-08-30
Laue diffuse

dermen:
Mike's and David's PR for better diffuse scattering models in diffBragg

PR #846, opened 2023-02-10
Diffbragg2kokkos

Trzs:
Add KOKKOS backend to diffBragg. 
The CUDA backend will be deprecated and removed in the future. 
This PR includes the changes from #810 and #785.

#### Current issues
Kokkos requires all kokkos objects to be out-of-scope/deleted before the end of the program. Due to the way diffbragg interacts with python, some kokkos objects are not deleted before the end. This leads to error messages when the program ends: 
  - `malloc_consolidate(): invalid chunk size`
  - `corrupted size vs. prev_size in fastbins`, etc.
  
This doesn't affect the normal program execution, but it causes the unit tests to "fail", because the last output is not "OK", but some error message.


bkpoon:
Can you try rebasing the `master` branch? There are some conflicts that need to be resolved. Also, can you add a check to `libtbx/SConscript` so that `enable_kokkos_if_possible` is not run on Python 2?

Trzs:
Thanks for the advice, I will rebase it on master. 
Do you know if some script already has checks for python 2, which I could just copy?

bkpoon:
Wait, why is the building of `kokkos` related things being moved into `libtbx/SConscript`? Should we have a `kokkos_adaptbx` instead of a `kokkotbx`? And the `SConscript` would be in there? If we do that, then the `enable_kokkos_if_possible` function should set `env_etc.enable_kokkos` to `True` for Python 3 only and for any other restrictions (e.g. linux). We should consolidate the checks upfront. Something like
```python
def enable_kokkos_if_possible():
  env_etc.enable_kokkos = False
  if (libtbx.env.build_options.enable_kokkos):
    if sys.version_info.major == 2:
      raise RuntimeError('Kokkos is not available on Python 2')
    if not sys.platform.startswith('linux'):
      raise RuntimeError('Kokkos is only available on linux platforms')
    env_etc.enable_kokkos = True
```
Then the `SConscript` for `kokkostbx` would only check `env_etc.enable_kokkos` before building (or anything Kokkos related). Then the check in `simtbx/SConscript` can also be simplified.
```python
# only build kokkos on linux and if kokkos is enabled
if sys.platform.startswith('linux') and env_etc.enable_kokkos:
  env_simtbx.SConscript("kokkos/SConscript",exports={ 'env' : env_simtbx })
```

Trzs:
Because Kokkos is now used in two different places (`simtbx/diffbragg` and `simtbx/kokkos`), I needed to move the `kokkos` compilation to a higher level. I put it into `libtbx`, because I was mainly copying how cuda is implemented and because I wanted to prepare for kokkos being used in other places.
I don't know what it would mean to convert `kokkostbx` to `kokkos_adaptbx`, so I don't know what the drawbacks of this move would be. 
Ideally, I think having one SConscript in kokkostbx to build the kokkos library would be the best option. Then other kokkos related files such as simtbx/kokkos/kokkos_instance can also move there.

bkpoon:
I don't have a strong preference about the naming of `kokkostbx` or `kokkos_adaptbx`. @nksauter do you have a preference? If not, let's just keep `kokkostbx`. The '_adaptbx` suffix is generally for special modules for building other code.

But yes, let's move the building of base `kokkos` things to a `SConscript` in `kokkostbx`. And we should consolidate the checks for kokkos in `libtbx/SConscript`. Then downstream packages can just check `env_etc.enable_kokkos`. I think originally, we just put checks wherever we were making changes, but that is something that can be streamlined now.

The `enable_cuda_if_possible` is a little more complicated because it sets up how to build `.cu` files and sets up other default CUDA flags. But nothing is actually built.

One other thing to consider is how does this affect the Arcticus builds? Should we merge that code first?

nksauter:
I have no preference on naming.
I would not want the Arcticus port to block this pull request.
No strong opinions on changes.  Would like to have a rebase against master, then a squash merge.  Eventually a removal of the Eigen dependency.  At some point a bug fix for the kokkos deallocation.

nksauter (approved):

mewall (commented):
Overall it looks reasonable to me, thanks. Has the diffuse code been tested? Do you need help with that?

Trzs:
Mike, I run all the unit tests and they passed successfully. Do you have other tests I should run?

mewall:
> 

@bkpoon I just ran into an issue with building on a machine where I need to use the compilers in the environment rather than the conda compilers. Any advice?

export CC=$(which gcc)
export CXX=$(which g++)

python bootstrap.py build --builder=xfel --config-flags="--enable_openmp_if_possible=True" --config-flags="--enable_kokkos" --config-flags="--enable_cuda" --use-conda --python=39 --build-dir="build_kokkos" --config-flags="--use_environment_flags"

@Trzs, @bkpoon I think we don't need to follow the submodules for kokkos-kernels (we just need the headers for that right now, and it slows things down). Would it be possible to clone that repo without them?

mewall:
@bkpoon OK, I fixed the following issue by ensuring that no conda env was activated before starting the build.

> > 
> 
> @bkpoon I just ran into an issue with building on a machine where I need to use the compilers in the environment rather than the conda compilers. Any advice? It looks like the env compilers are used, but linking is occurring using the conda compilers.
> 
> export CC=$(which gcc) export CXX=$(which g++)
> 
> python bootstrap.py build --builder=xfel --config-flags="--enable_openmp_if_possible=True" --config-flags="--enable_kokkos" --config-flags="--enable_cuda" --use-conda --python=39 --build-dir="build_kokkos" --config-flags="--use_environment_flags"
> 
>

bkpoon:
Yes, that should work. By default, I think that builder uses conda environments from the `dials` repository, so it will have the conda compilers in `conda_base`. And on linux, the compilers and headers are separated, so you should not be using the headers from the conda compilers. If you do need to activate the environment, the compiler packages will set a bunch of environment variables (type `printenv` in an active environment to see them). If you want to unset all of them, you will need to remove the compiler packages. It is easiest to delete `conda_base` and recreate it. To do so,

1. delete the existing `conda_base`
2. edit `modules/dials/.conda-envs/linux.txt` and remove the `cxx-compiler` and `c-compiler` entries
3. Run `conda create -p ./conda_base --file modules/dials/.conda-envs/linux.txt python=3.9`

That should recreate `conda_base` (you can pick whichever python version). Then you should delete the "build" directory and rebuild everything with your compiler.

mewall:
@Trzs I was able to build with a Cuda back end targeting V100 using KOKKOS_DEVICES=Cuda KOKKOS_ARCH=Volta70

It looks like there are two cycles of building kokkos via cmake when running bootstrap build. One targets CUDA for device, OpenMP for host parallel, and SERIAL for host serial. The other targets CUDA for device and SERIAL for host serial, without a host parallel back end. Is that what you're seeing? Can it be modified so that there's only one build?

Also do you know how to exercise the kokkos code path, e.g., via the diffBragg tests? I can see a difference in time when I use 

python tst_diffBragg_diffuse_properties.py --cuda --stencil 1

vs

python tst_diffBragg_diffuse_properties.py --stencil 1

but I used --enable_cuda for the build, so not sure if I'm just exercising the cuda code path and the non-kokkos host code path in this case, respectively.

Trzs:
@mewall glad to hear you could built it. The two cycles are a bit annoying and I have to look into where this is coming from. So far, it seems that CCTBX itself has loops twice through the built process?
@bkpoon do you know if it's better to add a flag so kokkos is only build during the first run or should I try to remove the second build run altogether?

Concerning the tests, your interpretation is correct. Without "--cuda" you are using the cpu path and with "--cuda" it's using cuda (or now kokkos). If you want to trigger this with your own script, diffBragg is looking for the `DIFFBRAGG_USE_CUDA` environment variable to decide which routine to use.

dermen:
Nice work! The week before last I added performance changes to the `diffBraggCUDA.cu` file (just to be more conservative when doing host-to-device copies). I'm wondering if that is relevant for kokkos?

Anyhow, I think the best way for me to really test this branch is to process the SwissFEL data with it and see what happens... However I cant do that until early next week (probably Monday).

Trzs:
@dermen so far I have mostly replicated your Cuda implementation in Kokkos (including the use of UVM), so any improvement to the Cuda version should also benefit Kokkos. Long-term I want to try to use GPU-aware MPI, which would limit host-device communication to set-up at the beginning and then copying the results at the end.

Trzs:
@dermen I just noticed a difference between _diffBragg_cpu_kernel_ and _diffBragg_gpu_kernel_ (the cuda version doesn't use the occupancy). Do you know which is correct? With or without `occ`?

Screen Shot 2023-02-17 at 9 32 19 AM

Screen Shot 2023-02-17 at 9 32 46 AM

bkpoon:
@Trzs There are 2 builds of kokkos because there is one for kokkos-kernels. Do we need to compile kokkos-kernels for any future use? If not, I think we can comment out the build for kokkos-kernels and the second kokkos build. Or does the kokkos-kernels build generate the headers for the architecture? Also, I remember the branch for the Arcticus build was supposed to have consolidated the 2 kokkos builds, so we can sort it out after this is merged.

I have a patch for adding `Turing75` as a supported architecture (basically adding `Turing75` as a supported architecture and adding `-DKokkos_ARCH_TURING75` flag), but I can commit it after this PR is merged. This is for GPU support in Google Colab notebooks for @irisdyoung . I have a successfully built the conda package with CUDA support. Is there a simple test I can run in the notebook? @mewall I did have to disable `lunus` for that package to build. There was some issue with `nvcc` not defined when building `lunus` as well as errors with duplicate definitions. I can debug that after seeing if the GPU is being used correctly in Google Colab.

Trzs:
@bkpoon to test kokkos, I often use "%run modules/cctbx_project/simtbx/kokkos/tst_kokkos_lib.py. This should write some files

bkpoon:
@Tzrs Does this output look right? Kokkos seems to be used and I'm getting a warning about CUDA, so I guess the GPU was initialized? This is running on the Google GPU backend.
```
!python /usr/local/lib/python3.8/site-packages/simtbx/kokkos/tst_kokkos_lib.py

Make a beam
Make a dxtbx detector
Make a randomly oriented xtal

# Use case 2.  Three-wavelength polychromatic source

assume three energy channels
CPUnanoBragg_API+++++++++++++ Wavelength 0=1.375602, Flux 1.666667e+10, Fluence 1.666667e+18
CPUnanoBragg_API+++++++++++++ Wavelength 1=1.377602, Flux 5.000000e+10, Fluence 5.000000e+18
CPUnanoBragg_API+++++++++++++ Wavelength 2=1.379602, Flux 3.333333e+10, Fluence 3.333333e+18

# Use case: modularized api argchk=False, cuda_background=False
 - Skipped, no cuda module simtbx_gpu_ext found.

# Use case: modularized api argchk=False, cuda_background=True
 - Skipped, no cuda module simtbx_gpu_ext found.
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false

Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
                                  without setting CUDA_MANAGED_FORCE_DEVICE_ALLOC=1 or
                                  setting CUDA_VISIBLE_DEVICES.
                                  This could on multi GPU systems lead to severe performance"
                                  penalties.

Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
                                  without setting CUDA_MANAGED_FORCE_DEVICE_ALLOC=1 or
                                  setting CUDA_VISIBLE_DEVICES.
                                  This could on multi GPU systems lead to severe performance"
                                  penalties.

# Use case: modularized api argchk=False, cuda_background=True

polychromatic KOKKOS no argchk
USE_EXASCALE_API+++++++++++++ Wavelength 0=1.375602, Flux 1.666667e+10, Fluence 1.666667e+18
USE_EXASCALE_API+++++++++++++ Wavelength 1=1.377602, Flux 5.000000e+10, Fluence 5.000000e+18
USE_EXASCALE_API+++++++++++++ Wavelength 2=1.379602, Flux 3.333333e+10, Fluence 3.333333e+18
Pixel differences between CPU and KOKKOS, minimum=-0.0156 mean=0.0000 maximum=0.0156

# Use case: modularized api argchk=True, cuda_background=True

polychromatic KOKKOS argchk
USE_EXASCALE_API+++++++++++++ Wavelength 0=1.375602, Flux 1.666667e+10, Fluence 1.666667e+18
USE_EXASCALE_API+++++++++++++ Wavelength 1=1.377602, Flux 5.000000e+10, Fluence 5.000000e+18
USE_EXASCALE_API+++++++++++++ Wavelength 2=1.379602, Flux 3.333333e+10, Fluence 3.333333e+18
Pixel differences between CPU and KOKKOS, minimum=-0.0156 mean=0.0000 maximum=0.0156
OK
```
And the output to `!nvidia-smi` is
```
Fri Feb 17 21:46:13 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 510.47.03    Driver Version: 510.47.03    CUDA Version: 11.6     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            Off  | 00000000:00:04.0 Off |                    0 |
| N/A   52C    P0    25W /  70W |      0MiB / 15360MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

Trzs:
@bkpoon 👍 looks good

mewall:
From what I can see in my cctbx build there are actually three builds of kokkos


The first is the non-cmake build


The second is a cmake build


The third is a repeat of a cmake build, after all of cctbx appears to have been built already

Let me know if you'd like to look at lunus together in the coming week. It would probably be good if that can coexist with the Google Colab notebooks.
From: Billy K. Poon ***@***.***>
Sent: Friday, February 17, 2023 11:54:13 AM
To: cctbx/cctbx_project
Cc: Wall, Michael E; Mention
Subject: [EXTERNAL] Re: [cctbx/cctbx_project] Diffbragg2kokkos (PR #846)


@Trzs<https://urldefense.com/v3/__https://github.com/Trzs__;!!Bt8fGhp8LhKGRg!Bknto391aGMsO9tcRJejUd0vtT6u-e7N2u4CNSCemszai94_e1xiIL_s-F6nGfX4VifaFQrT2WuC2DxI_S7fBp1g$> There are 2 builds of kokkos because there is one for kokkos-kernels. Do we need to compile kokkos-kernels for any future use? If not, I think we can comment out the build for kokkos-kernels and the second kokkos build. Or does the kokkos-kernels build generate the headers for the architecture? Also, I remember the branch for the Arcticus build was supposed to have consolidated the 2 kokkos builds, so we can sort it out after this is merged.

I have a patch for adding Turing75 as a supported architecture (basically adding Turing75 as a supported architecture and adding -DKokkos_ARCH_TURING75 flag), but I can commit it after this PR is merged. This is for GPU support in Google Colab notebooks for @irisdyoung<https://urldefense.com/v3/__https://github.com/irisdyoung__;!!Bt8fGhp8LhKGRg!Bknto391aGMsO9tcRJejUd0vtT6u-e7N2u4CNSCemszai94_e1xiIL_s-F6nGfX4VifaFQrT2WuC2DxI_T28M3iP$> . I have a successfully built the conda package with CUDA support. Is there a simple test I can run in the notebook? @mewall<https://urldefense.com/v3/__https://github.com/mewall__;!!Bt8fGhp8LhKGRg!Bknto391aGMsO9tcRJejUd0vtT6u-e7N2u4CNSCemszai94_e1xiIL_s-F6nGfX4VifaFQrT2WuC2DxI_eEhhBbj$> I did have to disable lunus for that package to build. There was some issue with nvcc not defined when building lunus as well as errors with duplicate definitions. I can debug that after seeing if the GPU is being used correctly in Google Colab.

dermen:
@Trzs , with occupancy is correct! I think occupancy was always 1, hence never noticed

Trzs:
> @Trzs , with occupancy is correct (Cpp version is correct)! I think occupancy was always 1, hence never noticed

Thanks for clarifying, I'll add the occupancy to the cuda and kokkos version.

simtbx/kokkos/SConscript
```diff
@@ -204,10 +204,14 @@ print('-'*79)
...
-kokkos_cxxflags = kokkos_cxxflags.split(b'\n')
-kokkos_cxxflags = kokkos_cxxflags[1].decode('utf8').split()
-if kokkos_cxxflags and kokkos_cxxflags[0] == 'echo':
-  kokkos_cxxflags = [f.strip('"') for f in kokkos_cxxflags[1:]]
+#print('5'*79)
+#print(os.environ['KOKKOS_CXXFLAGS'])
+#print('5'*79)
+#kokkos_cxxflags = kokkos_cxxflags.split(b'\n')
+#kokkos_cxxflags = kokkos_cxxflags[1].decode('utf8').split()
+kokkos_cxxflags = os.environ['KOKKOS_CXXFLAGS'].split()
```
bkpoon, line 212: @Trzs I think this part removes the "echo" that is showing up. I remember we saw this earlier and this is probably why it's here.

bkpoon:
It looks like the failure now is that Kokkos is not enabled for the psana build. Are we turning that on for psana? For macOS, the build will fail because the `--enable_kokkos` flag now fails on non-linux platforms. In `.azure-pipelines/xfel/unix-conda-build.yml` there would need to be a way to turn off that flag for macOS builds. I think you can set an environment variable with the `--config-flags` and add the kokkos flag if the Azure platform is linux. Something like
```
CONFIG_FLAGS='      --config-flags="--enable_openmp_if_possible=True" \
      --config-flags="--compiler=conda" \
      --config-flags="--use_environment_flags" \
      --config-flags="--no_bin_python" \'
if [ "$(OS)" = "linux" ]; then
  CONFIG_FLAGS = ${CONFIG_FLAGS}'      --config-flags="--enable_kokkos" \'
fi
```
and then later on in the `bootstrap.py` call, you replace all the flags with the environment variable
```
    python bootstrap.py \
      --builder=xfel \
      --use-conda $CONDA_PREFIX \
      --nproc=4 \
      ${CONFIG_FLAGS}
      --no-boost-src \
      build
```

mewall:
Felix and I met and confirmed that the kokkos diffBragg code with diffuse scattering can be used with simtbx.sim_view. Also reviewed the replacement of Eigen matrices with Kokkos matrices and convinced ourselves that this was done correctly. OK to merge.

mewall (approved):

dermen:
With the kokkos kernel, can GPU vs CPU usage be determined at runtime ? Any particular reason why diffBragg_sum_over_steps (from diffBragg_cpu_kernel.cpp) still seems to play a role here (unless Im missing something)

Trzs:
@dermen I've basically just replaced all CUDA parts with Kokkos equivalents. By default diffBragg uses the cpu version. Only when DIFFBRAGG_USE_CUDA is set, will it use the kokkos backend (I will change this to DIFFBRAGG_USE_KOKKOS to avoid confusion). If you say the pure CPU version is no longer needed, I can remove it. My motivation to keep it for now was as a comparison and as a backup when kokkos is not available (e.g. on Windows)

dermen:
@Trzs , thanks for clarifying.

dermen:
@Trzs , for what its worth,  I could not --enable_kokkos using cmake and cuda version 12+: I had to downgrade to cuda 11.8. Seems this is fixed in recent kokkos versions (see [here](https://github.com/kokkos/kokkos/commit/f64e5a66db145ffb8c6dd61b649ab914ddbc058a)) . But the downgrade was necessary for me.. 

Also, I havent begun testing yet, but the only way I could get this branch to build on my machine (CentOS7 with V100) was to add this diff: 

```diff
diff --git a/simtbx/kokkos/SConscript b/simtbx/kokkos/SConscript
index 254e98c..d6a2bc2 100644
--- a/simtbx/kokkos/SConscript
+++ b/simtbx/kokkos/SConscript
@@ -31,7 +31,7 @@ os.environ['CXXFLAGS'] = '-O3 -fPIC -DCUDAREAL=double'
 
 library_flags = "-Llib"
 if use_cuda:
-  library_flags += " -L$(CUDA_HOME)/lib64"
+  library_flags += " -L$(CUDA_HOME)/lib64 -L$(CUDA_HOME)/lib64/stubs"
   if os.getenv('CUDA_COMPATIBILITY'):
     library_flags += " -L$(CUDA_HOME)/compat"
 elif use_sycl:
@@ -282,6 +282,7 @@ if not env_etc.no_boost_python:
   if 'Cuda' in os.getenv('KOKKOS_DEVICES'):
     kokkos_ext_env.Append(LIBPATH=[os.path.join(os.environ['CUDA_HOME'], 'lib64')])
     kokkos_ext_env.Append(LIBPATH=[os.path.join(os.environ['CUDA_HOME'], 'compat')])
+    kokkos_ext_env.Append(LIBPATH=[os.path.join(os.environ['CUDA_HOME'], 'lib64/stubs')])
     kokkos_ext_env.Append(LIBS=env_etc.libm + default_libs + ["cudart", "cuda"])
   elif 'HIP' in os.getenv('KOKKOS_DEVICES'):
     kokkos_ext_env.Append(LIBPATH=[os.path.join(os.environ['ROCM_PATH'], 'lib')])
```

Im not sure if the diff starting on line 31 is necessary, but the diff starting on line 282 is definitely necessary, as for me libcuda.so is only found in in $CUDA_HOME/lib64/stubs

I built with 

```
 python bootstrap_kokkos.py build --builder=dials --nproc=24 --config-flags="--enable_cuda" --config-flags="--enable_kokkos" --config-flags="--enable_openmp_if_possible=True"  --use-conda=$CONDA_PREFIX --build-dir=kokkos_build2 --config-flags="--use_environment_flags" --config-flags="--enable_cxx11"
```

simtbx/diffBragg/hopper_utils.py
```diff
@@ -1673,6 +1673,56 @@ def __call__(self, x, *args, **kwargs):
         self.g = g
         return f
 
+    def callback(self, x, verbose=True):
```
dermen, line 1676: Was this method needed by you ? I cant remember why I deleted it, but seems its gone from master

Trzs: Thanks for pointing this out. It seemed to have slipped through. I removed it.

Trzs:
@dermen I added your changes. It seems that the lib/stubs directory is used when cuda is installed without the Nvidia drivers (e.g. a login node without a GPU). It should not be necessary if you have a GPU and the driver installed, but it's useful to cover this case.

dermen:
@Trzs , great. I ran some of the kokkos+Gpu unit tests and they passed except for the deallocation bug:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Kokkos allocation "m_Fhkl_scale_deriv" is being deallocated after Kokkos::finalize was called

Traceback functionality not available

Aborted (core dumped)
```

Do we want to handle the case when someone tries to compile for kokkos and CUDA 12+? Seems commit [f64e5a6](https://github.com/kokkos/kokkos/commit/f64e5a66db145ffb8c6dd61b649ab914ddbc058a) in the develop kokkos branch solves the issue, but I dont think this is in master yet.. 

I'd like to test processing the some full datasets at NERSC before we merge if possible, as this PR will disable my current CUDA workflow (I can always fork if necessary)

Also, I get these warning messages every time I run a test:

```

Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
                                  without setting CUDA_LAUNCH_BLOCKING=1.
                                  The code must call Cuda().fence() after each kernel
                                  or will likely crash when accessing data on the host.

Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
                                  without setting CUDA_MANAGED_FORCE_DEVICE_ALLOC=1 or
                                  setting CUDA_VISIBLE_DEVICES.
                                  This could on multi GPU systems lead to severe performance"
                                  penalties.
```

Do you also see these ?

Trzs:
I would not add something special for Cuda 12, it should be fixed by the kokkos team soon.

Concerning the warnings you get, you only need to define these env variables to take care of them:

export OMP_PLACES=threads
export OMP_PROC_BIND=spread
export CUDA_LAUNCH_BLOCKING=1

In your case, it seems you also need to define CUDA_VISIBLE_DEVICEs=0 or equivalent.

If you want to test it first, please let me know when you are finished. 
Unfortunately, I haven't found a way to get rid of these errors on shutdown and since they don't interfere with the program itself, I'm proceeding with the PR

bkpoon:
There is a test failure on Python 3.9, but not on Python 3.8
```
libtbx.python -m pytest -rsxX --basetemp=pytest/t001 "/__w/1/modules/xfel_regression/test/command_line/test_lunus_worker.py::test_lunus_worker" --regression [FAIL] 40.1s
  Time: 40.08
  Return code: 1
  OKs: 18
  Standard error:
    
    dials_data = <DataFetcher: /__w/1/build/dials_data>
    run_in_tmpdir = local('/__w/1/tests/pytest/t001/test_lunus_worker0')
    
        def test_lunus_worker(dials_data, run_in_tmpdir):
          from dxtbx.model.experiment_list import ExperimentList
          from dxtbx.imageset import ImageSetFactory
          pytest.importorskip("mpi4py")
        
          xfel_regression_dir = libtbx.env.find_in_repositories(relative_path="xfel_regression", test=os.path.isdir)
          lunus_dir = os.path.join(xfel_regression_dir, 'lunus')
        
          grid_path = dials_data("thaumatin_grid_scan")
        
          experiments = ExperimentList.from_file(os.path.join(lunus_dir, 'combined.expt'), check_format=False)
          assert len(experiments) == 3
        
          for experiment in experiments:
            img_name = os.path.basename(experiment.imageset.paths()[0])
            img_path = os.path.join(grid_path, img_name)
            experiment.imageset = ImageSetFactory.make_imageset([img_path])
        
          experiments.as_file('combined.expt')
          shutil.copy(os.path.join(lunus_dir, 'combined.refl'), '.')
        
          params_file = "lunus_merge.phil"
          shutil.copy(os.path.join(lunus_dir, params_file), '.')
    
          deck_file = "deck.sh"
          shutil.copy(os.path.join(lunus_dir, deck_file), '.')
        
          command = "mpirun -n 3 --map-by ppr:3:node:PE=2 cctbx.xfel.merge lunus_merge.phil"
        
          result = procrunner.run(command.split(), environment_override={'OMP_NUM_THREADS':2})
    >     assert not result.returncode and not result.stderr
    E     assert (not 143)
    E      +  where 143 = ReturnObject(args=('mpirun', '-n', '3', '--map-by', 'ppr:3:node:PE=2', 'cctbx.xfel.merge', 'lunus_merge.phil'), return...w/1/miniforge/envs/py39/lib/python3.9/site-packages/pandas/_libs/algos.cpython-39-x86_64-linux-gnu.so\n7f62bab21000-7").returncode
    
    /__w/1/modules/xfel_regression/test/command_line/test_lunus_worker.py:39: AssertionError
    ----------------------------- Captured stdout call -----------------------------
    The following parameters have been modified:
    
    dispatch {
      step_list = input balance statistics_unitcell lunus
    }
    input {
      path = .
      reflections_suffix = combined.refl
      experiments_suffix = combined.expt
    }
    lunus {
      deck_file = deck.sh
    }
    
    LPROCIMLT: (scale, error) = 1 0
    LPROCIMLT: (scale, error) = 0.980397 0.00263969
    *** Error in `/__w/1/build/../miniforge/envs/py39/bin/python': double free or corruption (!prev): 0x00005596b0312630 ***
```

dermen:
Trying to build now at NERSC, getting this cmake below, any ideas ? 

```
make: 'libkokkos.a' is up to date.

Copying Kokkos library
-------------------------------------------------------------------------------
Copied
  source:      /global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos/libkokkos.a
  destination: /global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/kokkos_gpubuild/lib/libkokkos.a

===============================================================================
Building kokkos with cmake
-------------------------------------------------------------------------------
-- Setting default Kokkos CXX standard to 14
-- The project name is: Kokkos
/bin/bash: /global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/kokkos_gpubuild/../../CrystalX/conda_base/lib/libtinfo.so.6: no version information available (required by /lib64/libreadline.so.7)
-- Compiler Version: 11.1.105
-- Using -std=c++14 for C++14 standard as feature
-- Built-in Execution Spaces:
--     Device Parallel: Kokkos::Cuda
--     Host Parallel: NoTypeDefined
--       Host Serial: SERIAL
-- 
-- Architectures:
--  VOLTA70
CMake Error at cmake/kokkos_functions.cmake:368 (MESSAGE):
  TPL Interface library LIBDL should not have an IMPORTED_LOCATION
Call Stack (most recent call first):
  cmake/kokkos_functions.cmake:761 (KOKKOS_CREATE_IMPORTED_TPL)
  cmake/Modules/FindTPLLIBDL.cmake:1 (KOKKOS_FIND_IMPORTED)
  cmake/kokkos_functions.cmake:279 (FIND_PACKAGE)
  cmake/kokkos_tpls.cmake:80 (KOKKOS_IMPORT_TPL)
  cmake/kokkos_tribits.cmake:247 (INCLUDE)
  CMakeLists.txt:189 (KOKKOS_SETUP_BUILD_ENVIRONMENT)


-- Using internal desul_atomics copy
-- Kokkos Devices: CUDA;SERIAL, Kokkos Backends: CUDA;SERIAL
-- Configuring incomplete, errors occurred!
See also "/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos_build/CMakeFiles/CMakeOutput.log".
-- Setting default Kokkos CXX standard to 14
-- The project name is: Kokkos
/bin/bash: /global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/kokkos_gpubuild/../../CrystalX/conda_base/lib/libtinfo.so.6: no version information available (required by /lib64/libreadline.so.7)
-- Compiler Version: 11.1.105
-- Using -std=c++14 for C++14 standard as feature
-- Built-in Execution Spaces:
--     Device Parallel: Kokkos::Cuda
--     Host Parallel: NoTypeDefined
--       Host Serial: SERIAL
-- 
-- Architectures:
--  VOLTA70
CMake Error at cmake/kokkos_functions.cmake:368 (MESSAGE):
  TPL Interface library LIBDL should not have an IMPORTED_LOCATION
Call Stack (most recent call first):
  cmake/kokkos_functions.cmake:761 (KOKKOS_CREATE_IMPORTED_TPL)
  cmake/Modules/FindTPLLIBDL.cmake:1 (KOKKOS_FIND_IMPORTED)
  cmake/kokkos_functions.cmake:279 (FIND_PACKAGE)
  cmake/kokkos_tpls.cmake:80 (KOKKOS_IMPORT_TPL)
  cmake/kokkos_tribits.cmake:247 (INCLUDE)
  CMakeLists.txt:189 (KOKKOS_SETUP_BUILD_ENVIRONMENT)


-- Using internal desul_atomics copy
-- Kokkos Devices: CUDA;SERIAL, Kokkos Backends: CUDA;SERIAL
-- Configuring incomplete, errors occurred!
See also "/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos_build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:631: cmake_check_build_system] Error 1
===============================================================================
Building kokkos_kernels
-------------------------------------------------------------------------------
-- The project name is: KokkosKernels
CMake Error at CMakeLists.txt:125 (FIND_PACKAGE):
  By not providing "FindKokkos.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Kokkos", but
  CMake did not find one.

  Could not find a package configuration file provided by "Kokkos" with any
  of the following names:

    KokkosConfig.cmake
    kokkos-config.cmake

  Add the installation prefix of "Kokkos" to CMAKE_PREFIX_PATH or set
  "Kokkos_DIR" to a directory containing one of the above files.  If "Kokkos"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos-kernels/build/CMakeFiles/CMakeOutput.log".
-- The project name is: KokkosKernels
CMake Error at CMakeLists.txt:125 (FIND_PACKAGE):
  By not providing "FindKokkos.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Kokkos", but
  CMake did not find one.

  Could not find a package configuration file provided by "Kokkos" with any
  of the following names:

    KokkosConfig.cmake
    kokkos-config.cmake

  Add the installation prefix of "Kokkos" to CMAKE_PREFIX_PATH or set
  "Kokkos_DIR" to a directory containing one of the above files.  If "Kokkos"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos-kernels/build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:631: cmake_check_build_system] Error 1
-- The project name is: KokkosKernels
CMake Error at CMakeLists.txt:125 (FIND_PACKAGE):
  By not providing "FindKokkos.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Kokkos", but
  CMake did not find one.

  Could not find a package configuration file provided by "Kokkos" with any
  of the following names:

    KokkosConfig.cmake
    kokkos-config.cmake

  Add the installation prefix of "Kokkos" to CMAKE_PREFIX_PATH or set
  "Kokkos_DIR" to a directory containing one of the above files.  If "Kokkos"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/modules/kokkos-kernels/build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:631: cmake_check_build_system] Error 1
Getting environment variables
-----------------------------------
```

dermen:
@Trzs  I keep seeing similar to @bkpoon  in production tests:

```
x/command_line/hopper.py", line 34, in <module>
    kokkos_run = gpu_instance(deviceId=0)
RuntimeError: Desul::Error: init_lock_arrays_cuda: post init kernel error(cudaErrorIllegalAddress): an illegal memory access was encountered
Kokkos::Cuda ERROR: Failed to call Kokkos::Cuda::finalize()
*** Error in `/global/cfs/cdirs/lcls/dermen/d9114_sims/CrystalNew/kokkos_gpubuild/../../CrystalX/conda_base/bin/python': double free or corruption (!prev): 0x000055555aa422a0 ***
```

however it is intermittent.. 

If I run a job, it might proceed (or hit this error).. 

Im testing 

```
DIFFBRAGG_USE_CUDA=1 hopper test.phil
```

(hopper is alias to simtbx/command_line/hopper.py)

Also, I had to add this commit to even test

```diff
diff --git a/simtbx/command_line/hopper.py b/simtbx/command_line/hopper.py
index e7796a441c..fb9cb19241 100644
--- a/simtbx/command_line/hopper.py
+++ b/simtbx/command_line/hopper.py
@@ -30,6 +30,8 @@ import numpy as np
 np.seterr(invalid='ignore')
 import os
 from libtbx.mpi4py import MPI
+from simtbx.kokkos import gpu_instance
+kokkos_run = gpu_instance(deviceId=0)
 
 COMM = MPI.COMM_WORLD
 # TODO, figure out why next 3 lines are sometimes necessary?!

```

dermen:
Below is a unit test which fails similar to hopper. It illustrates the hopper use-case, and passes under master branch:

```python
# usage : DIFFBRAGG_USE_CUDA=1 mpirun -n 2 libtbx.python tests/tst_gpu_free.py 1

import sys
from libtbx.mpi4py import MPI
COMM = MPI.COMM_WORLD
from dxtbx.model import DetectorFactory, BeamFactory
from simtbx.nanoBragg.tst_nanoBragg_multipanel import det_descr, beam_descr
from simtbx.diffBragg import diffBragg

NGPU_PER_NODE=int (sys.argv[1])

GPU_ID = COMM.rank % NGPU_PER_NODE
beam = BeamFactory.from_dict(beam_descr)
det = DetectorFactory.from_dict(det_descr)

# check if kokkos build
try:
    from simtbx.kokkos import gpu_instance
    gpu_instance(deviceId=GPU_ID)
except ImportError:
    pass

for i in range(COMM.size*3):
    if i % COMM.size != COMM.rank:
        continue
    D = diffBragg(det, beam)
    D.device_Id = GPU_ID   # perhaps unncessary in kokkos builds
    D.vectorize_umats()

    for _ in range(5):
        D.add_diffBragg_spots()

    D.free_all()
    D.free_Fhkl2()
    D.gpu_free()
    print("RANK %d, done with Iter %d" % (COMM.rank, i+1))

print("OK!")

```

mewall:
This might be related to the other errors where there's an attempt to deallocate memory managed by kokkos after kokkos is finalized.


Where can I find test_lunus_worker.py?

From: Billy K. Poon ***@***.***>
Sent: Saturday, February 25, 2023 12:08:57 PM
To: cctbx/cctbx_project
Cc: Wall, Michael E; Mention
Subject: [EXTERNAL] Re: [cctbx/cctbx_project] Diffbragg2kokkos (PR #846)


There is a test failure on Python 3.9, but not on Python 3.8

libtbx.python -m pytest -rsxX --basetemp=pytest/t001 "/__w/1/modules/xfel_regression/test/command_line/test_lunus_worker.py<https://urldefense.com/v3/__http://test_lunus_worker.py__;!!Bt8fGhp8LhKGRg!BXpptbr3LChbGPtPx9vb1AlDqiXOhAWCqjdDUuK9qb9gzoIHwol-cs86z2yVqYLsxRllvf5dBonqUAk2t1gVUDrn$>::test_lunus_worker" --regression [FAIL] 40.1s
  Time: 40.08
  Return code: 1
  OKs: 18
  Standard error:

    dials_data = <DataFetcher: /__w/1/build/dials_data>
    run_in_tmpdir = local('/__w/1/tests/pytest/t001/test_lunus_worker0')

        def test_lunus_worker(dials_data, run_in_tmpdir):
          from dxtbx.model.experiment_list import ExperimentList
          from dxtbx.imageset import ImageSetFactory
          pytest.importorskip("mpi4py")

          xfel_regression_dir = libtbx.env.find_in_repositories(relative_path="xfel_regression", test=os.path.isdir)
          lunus_dir = os.path.join(xfel_regression_dir, 'lunus')

          grid_path = dials_data("thaumatin_grid_scan")

          experiments = ExperimentList.from_file(os.path.join(lunus_dir, 'combined.expt'), check_format=False)
          assert len(experiments) == 3

          for experiment in experiments:
            img_name = os.path.basename(experiment.imageset.paths()[0])
            img_path = os.path.join(grid_path, img_name)
            experiment.imageset = ImageSetFactory.make_imageset([img_path])

          experiments.as_file('combined.expt')
          shutil.copy(os.path.join(lunus_dir, 'combined.refl'), '.')

          params_file = "lunus_merge.phil"
          shutil.copy(os.path.join(lunus_dir, params_file), '.')

          deck_file = "deck.sh<https://urldefense.com/v3/__http://deck.sh__;!!Bt8fGhp8LhKGRg!BXpptbr3LChbGPtPx9vb1AlDqiXOhAWCqjdDUuK9qb9gzoIHwol-cs86z2yVqYLsxRllvf5dBonqUAk2t0s3jLpY$>"
          shutil.copy(os.path.join(lunus_dir, deck_file), '.')

          command = "mpirun -n 3 --map-by ppr:3:node:PE=2 cctbx.xfel.merge lunus_merge.phil"

          result = procrunner.run(command.split(), environment_override={'OMP_NUM_THREADS':2})
    >     assert not result.returncode and not result.stderr
    E     assert (not 143)
    E      +  where 143 = ReturnObject(args=('mpirun', '-n', '3', '--map-by', 'ppr:3:node:PE=2', 'cctbx.xfel.merge', 'lunus_merge.phil'), return...w/1/miniforge/envs/py39/lib/python3.9/site-packages/pandas/_libs/algos.cpython-39-x86_64-linux-gnu.so\n7f62bab21000-7").returncode

    /__w/1/modules/xfel_regression/test/command_line/test_lunus_worker.py<https://urldefense.com/v3/__http://test_lunus_worker.py__;!!Bt8fGhp8LhKGRg!BXpptbr3LChbGPtPx9vb1AlDqiXOhAWCqjdDUuK9qb9gzoIHwol-cs86z2yVqYLsxRllvf5dBonqUAk2t1gVUDrn$>:39: AssertionError
    ----------------------------- Captured stdout call -----------------------------
    The following parameters have been modified:

    dispatch {
      step_list = input balance statistics_unitcell lunus
    }
    input {
      path = .
      reflections_suffix = combined.refl
      experiments_suffix = combined.expt
    }
    lunus {
      deck_file = deck.sh<https://urldefense.com/v3/__http://deck.sh__;!!Bt8fGhp8LhKGRg!BXpptbr3LChbGPtPx9vb1AlDqiXOhAWCqjdDUuK9qb9gzoIHwol-cs86z2yVqYLsxRllvf5dBonqUAk2t0s3jLpY$>
    }

    LPROCIMLT: (scale, error) = 1 0
    LPROCIMLT: (scale, error) = 0.980397 0.00263969
    *** Error in `/__w/1/build/../miniforge/envs/py39/bin/python': double free or corruption (!prev): 0x00005596b0312630 ***

dermen:
@mewall , I believe that `test_lunus_worker.py` is in the `xfel_regression` module. Is that on gitlab ? Nick would know. 

However, I think the problem we are seeing is the same, and it boils down to this line of code in simtbx/kokkos/__init__.py:

```
ext = bp.import_ext("simtbx_kokkos_ext")
```

Does lunus also import from `simtbx.kokkos` ?

dermen:
With the recent commit I was able to use Kokkos+GPU to process the entire cytochrome dataset (stage 1, 12,000 crystals) and get basically the same results as with the CUDA pipeline. It did take 27 minutes (Kokkos) versus 20 minutes (CUDA), however I suspect this timing difference is to do with some host-to-device speed ops I added to diffBraggCUDA.cu that aren't yet in master (mentioned above).  

But I feel this branch is nearly ready to merge. I only need to add ifdef statements to diffBragg_ext.cpp , see the TODO  in diffBragg_ext.cpp in the recent commit.  (@Trzs  do you know how to do this ? )

Note, with this recent commit, if you wrap diffBragg instances like so 

```
with DeviceWrapper(device_id) as wrapper:
    dB = diffBragg() 
    # whatever
    del dB
```

then the de-allocation error is avoided (test this by running the new script `DIFFBRAGG_USE_CUDA=1 libtbx.python tests/tst_hopper_usecase.py` ). 

Also note, adding the line `from simtbx import kokkos` to that same test will throw similar errors to what was reported above, so something is funky in my build's `simtbx_kokkos_ext` (and potentially the one Billy reported on above).

dermen:
As there are still some minor issues, I think it would be best if we maintain both a DIFFBRAGG_USE_CUDA flag and a DIFFBRAGG_USE_KOKKOS flag , just at least until the remaining kinks can be worked out. I think that is simply a couple lines of code in `diffBragg.cpp`. thoughts ?

mewall:
It is here


https://gitlab.com/cctbx/xfel_regression/-/blob/main/test/command_line/test_lunus_worker.py


and here is the merge worker


https://github.com/cctbx/cctbx_project/blob/master/xfel/merging/application/lunus/lunus.py


I just had a look, and I can't see a place where kokkos is initialized (the lunus.Process() constructor doesn't do it). I might be missing something, but it it's possible in this case that the worker is making use of kokkos code without first initializing kokkos.


I'm OK with allowing this test to fail for now and debugging this later. I'm going to meet with Billy next week and we can take a look at this issue then.

From: Derek Mendez ***@***.***>
Sent: Wednesday, March 1, 2023 7:36:15 PM
To: cctbx/cctbx_project
Cc: Wall, Michael E; Mention
Subject: [EXTERNAL] Re: [cctbx/cctbx_project] Diffbragg2kokkos (PR #846)


@mewall<https://urldefense.com/v3/__https://github.com/mewall__;!!Bt8fGhp8LhKGRg!DOspyRUuXZ1NF88C7-T3hBPG8avh43ssPcX9hCRwZK87tXzKcKj7yVCIabu_ruDeBpZem5_eTV1Aco4j-Hb_1WwU$> , I believe that test_lunus_worker.py<https://urldefense.com/v3/__http://test_lunus_worker.py__;!!Bt8fGhp8LhKGRg!DOspyRUuXZ1NF88C7-T3hBPG8avh43ssPcX9hCRwZK87tXzKcKj7yVCIabu_ruDeBpZem5_eTV1Aco4j-L9eMb5s$> is in the xfel_regression module. Is that on gitlab ? Nick would know.

However, I think the problem we are seeing is the same, and it boils down to this line of code in simtbx/kokkos/init.py:

ext = bp.import_ext("simtbx_kokkos_ext")


Does lunus also import from simtbx.kokkos ?

dermen (approved):

Trzs:
@dermen I'm still catching up:

- I don't know if this is still relevant, but the bug in `tst_gpu_free.py` is caused by initializing kokkos too late. I moved the initialize and it runs through:
```import sys
from libtbx.mpi4py import MPI
COMM = MPI.COMM_WORLD
NGPU_PER_NODE=int (sys.argv[1])

GPU_ID = COMM.rank % NGPU_PER_NODE
# check if kokkos build
try:
    from simtbx.kokkos import gpu_instance
    gpu_instance(deviceId=GPU_ID)
except ImportError:
    pass
from dxtbx.model import DetectorFactory, BeamFactory
from simtbx.nanoBragg.tst_nanoBragg_multipanel import det_descr, beam_descr
from simtbx.diffBragg import diffBragg

beam = BeamFactory.from_dict(beam_descr)
det = DetectorFactory.from_dict(det_descr)

for i in range(COMM.size*3):
    if i % COMM.size != COMM.rank:
        continue
    D = diffBragg(det, beam)
    D.device_Id = GPU_ID   # perhaps unncessary in kokkos builds
    D.vectorize_umats()

    for _ in range(5):
        D.add_diffBragg_spots()

    D.free_all()
    D.free_Fhkl2()
    D.gpu_free()
    print("RANK %d, done with Iter %d" % (COMM.rank, i+1))

print("OK!")```

- thanks for adding a solution for Kokkos:finalize!
- I'll look into the failing unit test

bkpoon:
Yeah, kokkos 4 is breaking a bunch of stuff. I did set the version to 3.7.01 here (https://github.com/cctbx/cctbx_project/commit/54534286c9b93cc06af40282d2836f989fc2d7ee), but I missed archive link.

bkpoon:
You may also need to set the version for `kokkos-kernels`.

Trzs:
Thanks for checking! I'll also fix the Kokkos-kernel version.

PR #847, opened 2023-02-21
experiment_filter: add ability to exclude cells matching a cluster

dwpaley:
This is motivated by the use case where a pseudomerohedral indexing ambiguity is resolved by unit cell clustering and manual reindexing. For example: for an orthorhombic unit cell 100 x 102 x 150 A, there is a pseudomerohedral indexing ambiguity around the rotation axis a+b. The misindexed lattices will cluster around 102 x 100 x 150 A. If the clusters are separable, then this is resolved by separately isolating the second cluster and reindexing it with modify_reindex_to_reference, then merging both clusters (which are now indexed consistently).

The skip_component parameter is needed because this approach fails if an experiment belongs to both components and thus appears twice in the final merge. Therefore when we isolate component 1, we exclude any experiment that also matches component 0.

### DIFF
--- a/xfel/merging/application/filter/experiment_filter.py
+++ b/xfel/merging/application/filter/experiment_filter.py
@@ -61,6 +61,11 @@ class experiment_filter(worker):
     features = np.array(features).reshape(1,-1)
     cov=self.cluster_data["populations"].fit_components[self.params.filter.unit_cell.cluster.covariance.component]
     m_distance = sqrt(cov.mahalanobis(features))
+    for other in self.params.filter.unit_cell.cluster.covariance.skip_component:
+      skip_cov = self.cluster_data["populations"].fit_components[other]
+      skip_distance = sqrt(skip_cov.mahalanobis(features))
+      if skip_distance < self.params.filter.unit_cell.cluster.covariance.skip_mahalanobis:
+        return False
     return m_distance < self.params.filter.unit_cell.cluster.covariance.mahalanobis
 
   def run(self, experiments, reflections):
--- a/xfel/merging/application/phil/phil.py
+++ b/xfel/merging/application/phil/phil.py
@@ -197,12 +197,19 @@ filter
           .type = path
         component = 0
           .type = int(value_min=0)
+        skip_component = None
+          .type = int(value_min=0)
+          .multiple = True
+          .help = If a lattice belongs to any of these components, exclude it from processing.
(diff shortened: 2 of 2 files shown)


xfel/merging/application/filter/experiment_filter.py
```diff
@@ -61,6 +61,11 @@ def check_cluster(self, experiment):
     features = np.array(features).reshape(1,-1)
     cov=self.cluster_data["populations"].fit_components[self.params.filter.unit_cell.cluster.covariance.component]
     m_distance = sqrt(cov.mahalanobis(features))
+    for other in self.params.filter.unit_cell.cluster.covariance.skip_component:
+      skip_cov = self.cluster_data["populations"].fit_components[other]
+      skip_distance = sqrt(skip_cov.mahalanobis(features))
+      if skip_distance < self.params.filter.unit_cell.cluster.covariance.skip_mahalanobis:
```
Baharis, line 67: @dwpaley Correct me if I'm wrong, but the output of this comparison operation in line 67 is a non-empty boolean list, which in Python always evaluates to True, so a `False` will be returned instead of a boolean iterable in all cases. Shouldn't this part rather declare an additional boolean array, to be considered alongside the boolean array in the return statement in line 69?

Baharis: I'm thinking about something along the lines of:

    accept_flag = m_distance < self.params.filter.unit_cell.cluster.covariance.mahalanobis
    dont_skip_flag = skip_distance > self.params.filter.unit_cell.cluster.covariance.skip_mahalanobis
    return accept_flag & dont_skip_flag

dwpaley: This whole method `check_cluster` gets called on 1 experiment at a time, meaning `skip_cov.mahalanobis(features)` is just a number (the normalized distance from the current cell to the cluster), not an array. The new "skip" mode uses the same syntax as the `m_distance` filtering on lines 63 and 69.

phyy-nx (approved):

irisdyoung (approved):

xfel/merging/application/filter/experiment_filter.py
Baharis: I must have used something similar that acted on an iterable, hence my confusion. Elegant solution 👍

PR #850, opened 2023-03-02
Dispersion

vganapati:
Added dispersion subdirectory to cctbx_project/cctbx to restrain f' and f" by the Kramers-Kronig relations (also known as the Hilbert transform). Modified .yml files as necessary for a successful build. Patches applied to diffBragg.cpp and merge.py, but these can be removed if necessary.


vganapati:
Thanks @nksauter I've removed the patches and will squash merge when ready.

nksauter (approved):
@bkpoon could you re-trigger the builds?

phyy-nx:
Hm, I just did and they all got auto-canceled.  Not sure why.  Also pinging @bkpoon :)

PR #852, opened 2023-03-07
wx.Timer.Start expects an integer value for milliseconds.

dagewa:
This fixes:

```
TypeError: Timer.Start(): argument 1 has unexpected type 'float'
```

observed with wx 4.2.0 when trying to run `cctbx.multiplicity_viewer`


dagewa:
https://docs.wxpython.org/wx.Timer.html#wx.Timer.Start

bkpoon:
Thanks!

PR #854, opened 2023-03-10
For wxpython 4.2.0 these functions expect certain parameters to be int.

dagewa:
These changes were required to get the `dials.reciprocal_lattice_viewer` working without error. There are probably other occurrences elsewhere.


bkpoon:
Thanks!

PR #857, opened 2023-03-15
diffBragg hardening

irisdyoung:
A couple fixes here to make diffBragg usable with LZ08 at scale. Mainly, we need to be able to skip over one or two bad files without crashing the whole job. Open to discussion if these are addressable individual failures though.

### DIFF
--- a/simtbx/command_line/hopper_ensemble.py
+++ b/simtbx/command_line/hopper_ensemble.py
@@ -99,7 +99,9 @@ if __name__ == "__main__":
                 os.makedirs(gather_dir)
 
     modelers = load_inputs(df, params, exper_key=args.exp, refls_key=args.refl, gather_dir=gather_dir)
-    # note, we only go beyond this point if perImport flag was not passed
+    # do not proceed beyond this point if we are only preimporting
+    if args.preImport:
+        sys.exit()
     modelers.cell_for_mtz = args.cell
     modelers.max_sigma = args.maxSigma
     modelers.outdir = args.outdir if args.outdir is not None else modelers.params.outdir
--- a/simtbx/command_line/hopper_ensemble.py
+++ b/simtbx/command_line/hopper_ensemble.py
@@ -22,6 +22,7 @@ import os
 import sys
 import logging
 import pandas
+import time
 
 from simtbx.diffBragg.hopper_ensemble_utils import load_inputs
 from libtbx.mpi4py import MPI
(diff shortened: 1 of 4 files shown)


simtbx/command_line/hopper_ensemble.py
```diff
@@ -75,7 +76,13 @@ def write_commandline(params):
     else:
         mpi_logger.setup_logging_from_params(params)
 
-    df = pandas.read_pickle(args.input)
+    for i in range(3):
```
dermen, line 79: what was the failure occurring here without the waiting ?

dermen: is there two jobs running simultaneously??

simtbx/diffBragg/hopper_ensemble_utils.py
```diff
@@ -641,16 +647,17 @@ def load_inputs(pandas_table, params, exper_key="exp_name", refls_key='predictio
...
+            if params.refiner.load_data_from_refl:
+                gathered = shot_modeler.GatherFromReflectionTable(expt, refls, sg_symbol=params.space_group)
+                MAIN_LOGGER.debug("tried loading from reflection table")
+            else:
+                gathered = shot_modeler.GatherFromExperiment(expt, refls, sg_symbol=params.space_group)
+                MAIN_LOGGER.debug("tried loading data from expt table")
             MAIN_LOGGER.debug("successfully loaded data")
+        except RuntimeError:
+            MAIN_LOGGER.critical("Failed to gather data from experiment %s; skipping.", exper_name)
+            continue
```
dermen, line 660: Got it, one way I got around this in the past was by pre-filtering the input pandas pickle, removing rows if they pointed to reflection tables with too few or too many reflections.. But this is ok to keep here

simtbx/diffBragg/utils.py
```diff
@@ -567,6 +567,8 @@ def get_roi_background_and_selection_flags(refls, imgs, shoebox_sz=10, reject_ed
     MAIN_LOGGER.debug("Number of skipped ROI with negative BGs: %d / %d" % (num_roi_negative_bg, len(rois)))
     MAIN_LOGGER.debug("Number of skipped ROI with NAN in BGs: %d / %d" % (num_roi_nan_bg, len(rois)))
     MAIN_LOGGER.info("Number of ROIS that will proceed to refinement: %d/%d" % (np.sum(selection_flags), len(rois)))
+    if np.sum(selection_flags) == 0:
+        raise RuntimeError("Can't proceed with zero ROIs")
```
dermen, line 571: Would it make sense to catch this error in hopper_utils.py / GatherFromExperiment:

```
try:
     roi_packet = utils.get_roi_background_and_selection_flag( ... 

except RuntimeError:  
    return False
```

something like that... the return value of this method is inspected in `command_line/hopper.py` ...

irisdyoung: Yep, that makes sense.

PR #861, opened 2023-03-21
Xfel drift table

Baharis:
Further improve robustness, maintainability, and speed of Drift tools by moving from pure python to numpy/pandas based implementation. Reorganize phil parameters to match actual workflow structure. Allow reading and saving data scraped by ,`DriftScraper` to a pickle in order to allow more dynamic and faster use, for example within xfel GUI. Increase the control over `DriftArtist`, making matplotlib errors less likely. Add type hints, examples, and docstrings.


Baharis:
@phyy-nx I added an option to dump and load scraped drift values to/from pickle(s). Resulting algorithm is still fairly slow, but can be performed in parts, e.g. individually for each run. Generated pickle files can be combined in any way desired and drawn together. In this way, if this part of the code was ever to be included in xfel GUI, it should be much easier.

PR #866, opened 2023-04-07
mmtbx: force rebuild of database pickle when loading fails

bkpoon:


chrissciwilliams:
Looks good.  I appreciate that you took care of cablam as well as rotarama.

PR #867, opened 2023-04-11
fixes to MetallicButton to address wxPython 4.2.0 changes

alyubimov:
- DrawBitmap requires x,y to be integers
- GetLabel rquires x,y to be integers


bkpoon:
Thanks!

Issue #868, opened 2023-04-14
undefined symbol error


Trzs:
A basic simtbx.kokkos script aborts with an undefined symbol error:

```
fwittwer@perlmutter$ cat test_script.py
from simtbx import get_exascale

def main():
    gpu_instance_type = get_exascale("gpu_instance", context='kokkos_gpu')
    gpu_instance = gpu_instance_type(deviceId = 0)

if __name__=="__main__":
    main()
    print("OK")

fwittwer@perlmutter$ libtbx.ipython test_script.py
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
File /pscratch/sd/f/fwittwer/workbench/alcc-recipes/cctbx/modules/cctbx_project/boost_adaptbx/boost/python.py:22, in import_ext(name, optional)
     21   sys.setdlopenflags(0x100|0x2)
---> 22 try: mod = __import__(name)
     23 except ImportError as e:

ImportError: /pscratch/sd/f/fwittwer/workbench/alcc-recipes/cctbx/build/lib/libsimtbx_kokkos.so: undefined symbol: _ZNK6simtbx9nanoBragg9nanoBragg9add_noiseEN6scitbx2af5versaIdNS3_9flex_gridINS3_5smallIlLm10EEEEEEE
[...]
```
However, unit tests run without a problem.

Trzs:
By simply adding an import from nanoBragg to the script, all runs correctly:
```
fwittwer@perlmutter$ cat test_script2.py
from simtbx import get_exascale
from simtbx.nanoBragg.tst_gauss_argchk import water

def main():
    gpu_instance_type = get_exascale("gpu_instance", context='kokkos_gpu')
    gpu_instance = gpu_instance_type(deviceId = 0)

if __name__=="__main__":
    main()
    print("OK")

fwittwer@perlmutter$ libtbx.ipython test_script2.py
OK
```

Trzs:
The missing `add_noise` function is currently defined in `nanoBragg_nks.cpp` and build into `simtbx_nanoBragg_ext.so`.
Possible solutions:
- include link to 'simtbx_nanoBragg_ext.so` into `libsimtbx_kokkos.so` -> non-conventional library name poses a challenge
- move `add_noise` to nanoBraggCUDA or some other library that is already linked

nksauter:
How about including "from simtbx import nanoBragg" in the file simtbx/__init__.py

This might be the most transparent way to accomplish the required library import.

PR #871, opened 2023-04-20
Fix undefined symbol error #868

Trzs:


nksauter:
Checks passed.  Assuming we still want/need this?  Squash merging now...

PR #890, opened 2023-06-15
Normalize diffuse rotations

mewall:
Normalize rotation matrices used to apply diffuse symmetry in diffBragg
o Required to make a number of them proper rotation matrices 
o Also add some comments and change matrix name to clarify what's
  going on when the principal axes of the diffuse model are rotated

### DIFF
--- a/simtbx/diffBragg/src/diffuse_util.h
+++ b/simtbx/diffBragg/src/diffuse_util.h
@@ -16,7 +16,7 @@ int gen_laue_mats(int laue_group_num, MAT3 *lmats) {
     return 0;
   }
 
-  double one_over_root2 = 1./sqrt(2.);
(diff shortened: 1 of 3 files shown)


simtbx/diffBragg/src/diffuse_util.h
```diff
@@ -15,6 +15,9 @@ int gen_laue_mats(int laue_group_num, MAT3 *lmats) {
   if ( laue_group_num < 1 or laue_group_num > 14) {
     return 0;
   }
+
+  double one_over_root2 = 1./sqrt(2.);
```
Trzs, line 19:
```suggestion
  const double one_over_root2 = 1./sqrt(2.);
```

Trzs (approved):

PR #891, opened 2023-06-16
Preserve `expt.identifier` in diffBragg hopper

Baharis:
This miniature PR will cause `simtbx.diffBragg.hopper` (or just `hopper`) to preserve the experimental identifier when saving a refined expt file from Modeler. I am staging the change here as a PR instead of committing directly mostly to avoid getting put on the blame list, as the current master does not build properly, and to remember to commit eventually.


Baharis:
@dermen I mentioned during the Tuesday meeting that I am going to suggest some small changes to diffBragg, but I did not expect it will be literally one line!

Issue #894, opened 2023-06-23
Matplotlib colormap deprecation


Baharis:
According to [matplotlib issue #20853](https://github.com/matplotlib/matplotlib/issues/20853) the interface for colormaps in matplotlib (`mpl`) changes. Method `mpl.cm.get_cmap('name')` is deprecated in version 3.7 and will be removed in version 3.9, probably in a year or so. Instead, `mpl.colormap.get_cmap('name')` or `mpl.colormap['name']` is encouraged. `plt.get_cmap()` is expected to remain available.

This is but a very minor issue for `cctbx.xfel` and other areas in active development. As they continue to use the most recent matplotlib version, the problem can be simply circumvented by refactoring all code to use the new syntax.

Nonetheless, this made me think about legacy software and codebase supported by cctbx. In their case, refactoring the code might break any functionality plotting with gradients, as the new syntax was introduced in matplotlib version 3.4.0, barely two years ago (Mar 26, 2021). If the current codebase should indeed support older applications such as the ones using python2, it needs to allow access to the current syntax existing back from matplotlib [version 2.2](https://matplotlib.org/2.2.4/api/cm_api.html?highlight=get_cmap#matplotlib.cm.get_cmap) from June 28, 2018.

irisdyoung:
Yep -- I think we have a fair number of try-excepts built into cctbx for cases like what you describe, and that's what I'd recommend here as well. It's a little clunky but does seem to do the job.

Baharis:
I agree, although quite frankly I'd just prefer the Python2 support to be dropped. @irisdyoung, are you aware of any place in the code, for example in libtbx, where such aliases are stored centrally? Scattering try-excepts in every place in cctbx where colormaps are used makes me feel physically ill.

irisdyoung:
I guess the closest to what you're hoping for would be how many files import from `six.moves` (as a result of the py2 -> py3 move and hypothetically maintaining backward compatibility). We don't have a centralized location of our own for this, to my knowledge. I think it's a good idea though, with the caveat that I wouldn't want to end up with a ton of unrelated code in one place. What about defining some plotting functionality in a central location, including colormaps? Some plotting utilities could be really handy anyway.

Baharis:
[...]

PR #897, opened 2023-06-26
Use `mpl.plt.get_cmap` instead of deprecated `mpl.cm.get_cmap`

Baharis:
In `cctbx.xfel`, the `matplotlib` colormaps are sometimes accessed using `matplotlib.cm.get_cmap`, which is deprecated starting from matplotlib version 3.7, see issue #894. This PR replaces these calls with a supported equivalent `plt.get_cmap`.

Furthermore, while browsing for colormap usage in cctbx, I have found multiple instances of colormaps being accessed directly, e.g. `matplotlib.cm.jet`. This way of accessing colormaps works in maplotlib 3.7 without raising a warning. I asked the developers whether this interface is expected to stay and currently left this code as-is.


phyy-nx (approved):

bkpoon:
The Windows failures can be ignored.

PR #902, opened 2023-06-29
Checking stage2

dermen:
Some minor updates needed to get the stage 2 results from the diffBragg paper working again. Results can be generated using instructions in the README here: 

https://github.com/dermen/cxid9114


dermen:
also, `roi.fit_tilt_using_weights` should probably always be False (it was ruining my stage2 refinement because the tilt planes were way off).

I think we should remove that parameter (and the functionality) altogether going forward, but this PR at least changes the default to be False..

nksauter:
@dermen in the file simtbx/nanoBragg/utils.py could you add a comment line specifying what data type is the parameter "roi"?  A novice coder might not know.

nksauter (approved):

PR #906, opened 2023-07-13
Update hopper_utils.py

Trzs:
Don't keep full background after ROIs are created


dermen (approved):
Looks good to merge this in!

PR #909, opened 2023-07-21
AttributeError bugfix in CCiso statistics.

nksauter:
Bug arises from cd5a215da6aa3fcdba4c86f1339133d9f056b573


Baharis (approved):
We actually noticed that working with Vidya (or Iris?) about 2 weeks ago! This should fix it. The only suggestion I have (to the previous commit) is `endswith` accepts a tuple of suffixes, so `assert model_file_path.endswith(('mtz', 'sf.cif'))` in line 480 is just as acceptable and shorter.

nksauter:
@bkpoon none of the Azure failures seem related to the pull request code, so I'm merging to master.

PR #912, opened 2023-07-27
Refactor `small_cell_index_detail`

dagewa:
This PR splits out the lattice determination part of `small_cell_index_detail` into a new function, `small_cell_index_lattice_detail`, so that this could be called by `dials.index`. This is necessary because `dials.index` loads experiments with `check_format=False`, so fails on the subsequent parts of the original function where the pixel data are accessed.

During the refactor, a certain amount of tidying-up was also done, such as reducing nested indents.

This PR has been tested to give the same results as `main` on this test case:

```
cd $(dials.data get -q 4fluoro_cxi)/lcls_2022_smSFX_workshop_data/indexing/
cctbx.xfel.small_cell_process params_1.phil ../ten_cbfs/*.cbf
```

Before a `small_cell` indexer is added to DIALS, @phyy-nx noted that it may be worth moving this code to `serialtbx`, to avoid adding a new DIALS --> cctbx.xfel dependency.

### DIFF
--- a/xfel/small_cell/small_cell.py
+++ b/xfel/small_cell/small_cell.py
@@ -352,9 +352,7 @@ def small_cell_index(path, horiz_phil):
   max_clique_len, experiments, refls = small_cell_index_detail(experiments, reflections, horiz_phil)
   return max_clique_len
 
-def small_cell_index_detail(experiments, reflections, horiz_phil, write_output = True):
-  """ Index an image with a few spots and a known, small unit cell,
-  with unknown basis vectors """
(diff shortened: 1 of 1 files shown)


phyy-nx (approved):
[Found](https://github.blog/2011-10-21-github-secrets/) a way to diff without whitespace by adding ?w=1 to the URL:
https://github.com/cctbx/cctbx_project/pull/912/files?w=1
Looks good, thanks!
@dwpaley we should consider if the integrate stuff here is needed anymore.  I think it might be left over from the original 2014-2015 implementation.

PR #915, opened 2023-08-10
adds support for reorienting diffuse model

dermen:
The diffBragg kernel is modified to accept an orientation matrix that will rotate the principal axes of the diffuse model. The default orientation is (a-b,a+b,c) which is the same as what was previously hard-coded into the kernel. The tst_diffbragg_diffuse_properties.py script shows how to set the orientation matrix to select either (a,b,c) or (a-b,a+b,c).


simtbx/diffBragg/tests/tst_diffBragg_diffuse_properties.py
```diff
@@ -16,6 +16,7 @@
 parser.add_argument("--stencil", type=int, default=0, help="sets the stencil property in diffBragg (default is 0)")
 parser.add_argument("--laue", action="store_true", help="sets the laue group number for the spacegroup")
 parser.add_argument("--grad", choices=['sigma','gamma'], default='gamma')
+parser.add_argument("--orientation", choices=[0,1], type=int, help="selects orientation for aniso. model", default=0)
```
Trzs, line 19: Could you add what option 0 and 1 mean? Something like:

"selects orientation for aniso. model (0: 45 deg to xy axes; 1: parallel to xyz axes)

Trzs (approved):

simtbx/diffBragg/src/diffBragg.cpp
```diff
@@ -382,6 +382,10 @@ diffBragg::diffBragg(const dxtbx::model::Detector& detector, const dxtbx::model:
                        0,.16,0,
                        0,0,.16;
 
+    db_cryst.rotate_principal_axes << 0.70710678,-0.70710678,0.,
```
mewall, line 385: Add a comment talking about justification for this default in terms of the calmodulin paper
Also change the root2 to make it computed as sqrt(2) in double precision

mewall (approved):

PR #920, opened 2023-08-23
Optimize setup performance and memory use by removing MPI.gather

Baharis:
This is the change to diffBragg discussed during today's and Monday's meeting. It heavily optimizes performance and memory use of diffBragg for experiments with a very large (>100 000 000) total number of reflections. This is achieved by moving previously separate calculations from across `ensemble_refine_launcher.py` and `stage_two_refiner.py` into a common class and utilizing `MPI.Allreduce` and `iterator`s in lieu of `MPI.gather` and `list`s. Changes include:

- `simtbx/diffBragg/ensemble_refine_launcher.py`:
  - Introduced class `HiAsu` responsible for managing Miller indices;
  - Replaced use of separate dicts, lists, counters with `HiAsu`;
  - Disabled multiplicity diagnostics `_get_unique_Hi` until a better implementation is ready;
  - Replaced `_mem_usage` with new `simtbx/diffBragg/utils.py` `memory_report`;
- `simtbx/diffBragg/hopper_ensemble_utils.py`
  - Moved `mem_usage` functionality to `memory_report` + `LOGGER.info` for more general logging which now correctly reports where it has been called;
- `simtbx/diffBragg/refiners/stage_two_refiner.py`:
  - Simplified + optimized by passing `HiAsu` from the launcher instead of recalculating statistics;
- `simtbx/diffBragg/utils.py`:
  - Introduced `memory_report` function which can be called anywhere to report memory use.
- `simtbx/command_line/stage_two.py`:
  - Add `max_sigz` phil par (defaults to 10.) to reject experiments diverged during stage1


Baharis:
I took the freedom to ask you all for reviews because:
- @dermen We mostly discussed the changes already, but I want you to be aware of all changes; I made a list of them and minimized the amount of changes so that the diff is much more readable than before;
- @Trzs GitHub suggested you authored some lines here, so if there is any functionality that needs to be considered e.g. for frontier, you are definitely most informed;
- @vganapati Similarily to Felix, you authored some lines recently, probably when debugging C2/m with Derek, so I mention you just in case you might have any valuable suggestions or worries.

This branch should be probably squash-merged into a single commit, and then can be easily merged to `frontier`.

Baharis:
Merged after agreement by @dermen and tacit agreement of @vganapati and @Trzs during the weekly CCTBX meeting.

PR #927, opened 2023-09-14
Building the latest source code for reduce, rather than a specific ve…

russell-taylor:
…rsion. Also switched to git pull for probe.


russell-taylor:
@bkpoon Vincent from the Richardson lab asked me to check whether we can use the current Reduce version in CCTBX. I had to modify the code so that it didn't export mmtbx_reduce_ext like the native CCTBX code does, but then it looks like it compiles and passes all of the tests. I renamed its library to mmtbx_reduceOrig_ext. This change also switches from SVN to Git to pull probe. Go ahead and merge if it looks okay to you.

PR #939, opened 2023-11-22
mpi_abort_on_exception: do nothing if communicator size is 1

dwpaley:
This prevents confusing error messages in some cases where the system has mpi4py but we are running single-process. One example is `cctbx.xfel.merge -h`, which currently prints:

```
Abort(1) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
```

at the end of its output because it hits a SystemExit inside the wrapped function. With this PR, in the given case we won't attempt to clean up the nonexistent MPI communicator.


dwpaley:
Thanks, will merge after the tests run.

PR #948, opened 2024-01-04
Debug kokkos cmake build error

mewall:
Try executing cmake in the kokkos source dir instead of in the build dir


Trzs (approved):

nksauter (approved):
Continuous integration is failing, but only in a MacOS build, thus not related to this project.

PR #960, opened 2024-01-17
properly set divergences using nanoBragg xray beams

dermen:


simtbx/command_line/stage_two.py
dermen, line 1: this sneaked in, this writes the stage 2 diff phil to disk , not relevant to PR but useful

simtbx/nanoBragg/nanoBragg_beam.py
```diff
@@ -4,17 +4,52 @@
...
+    self._divsteps = val
+
+  @property
+  def divergences(self):
+    divrange = self.divergence_mrad/1000.
+    if self.divsteps==0:
+      return [(0,0)]
+    else:
+      all_divs = np.arange(0, divrange+1e-7, divrange / self.divsteps) - divrange / 2
+      return [(hdiv, vdiv) for vdiv in all_divs for hdiv in all_divs]
```
Baharis, line 52: You might find the expression `np.linspace(-divrange/2, divrange/2, num=self.divsteps)` useful :) ([docs](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html#numpy.linspace)). I also don't know how many divsteps would be usually expected, but `[(hdiv, vdiv) for itertools.product(all_divs, all_divs)]` might be slightly faster and easier to read, though it's probably a matter of preference.

Baharis: As you can clearly see, my priorities are in a strange place, but I decided to comments in case anyone finds these useful. Also, as a pious member of the PEP8 church, there is a bunch of unnecessary semicolons in `rotate_axis`.

dermen: yea I will clean up rotate_axis, I think there is probably a one-liner in scipy.spatial somewhere

dermen: (or in scitbx somewhere.. )

Baharis: I accepted your challange and accidentally found one in [C++](https://github.com/cctbx/cctbx_project/blob/ca44a56a6632f52558d9bb049d2ca3e172b5f0a4/simtbx/nanoBragg/nanoBragg.cpp#L4122) and [CUDA](https://github.com/cctbx/cctbx_project/blob/02b6bb6c5bef18f0e54f9a64f51878bf17f07329/simtbx/nanoBragg/nanoBraggCUDA.cu#L957), but not in scitbx. There is `iotbx.detector.rotate_vector_around` which seems to have exactly expected signature that should work with numpy and is exposed to python ([link](https://github.com/cctbx/cctbx_project/blob/bc3b8eb7dfcbe7403cf40110e1ef0dea3a81aab6/iotbx/detectors/detectors_ext.cpp#L317)). This only grounds me in my opinion that there should be a common module with functions like that.

dermen: yea, I could actually expose the c++ one to python using boost (as opposed to copying it as I did, hence the semicolons).... there is also the route of creating a rotation matrix from axis and angle. I timed it and it took longer though

russell-taylor: I was looking for a fast way to rotate a point around an axis by an angle. I found some hidden down inside C++ code that was not accessible. I really had to make it faster because it was a bottleneck for me. So I exposed it as mmtbx.reduce.RotatePointDegreesAroundAxisDir() at https://github.com/cctbx/cctbx_project/blob/master/mmtbx/reduce/boost_python/reduce_bpl.cpp#L36 (you can find it used in Python code at https://github.com/cctbx/cctbx_project/blob/master/mmtbx/reduce/Movers.py#L1401). Hope it helps.

russell-taylor: You can use scitbx.matrix.rotate_point_around_axis if you don't need it to be super fast. The line after the one above that is commented out shows how to call that (which requires adding the offset to the origin).

Baharis: In case anyone is wondering, I looked at scipy and it has a truly elegant interface. `scipy.spatial.transform.Rotation.from_rotvec(np.pi/2 * np.array([0, 0, 1])).as_matrix()` returns a numpy array with rotation matrix. This method can be supplied an array of vectors to provide an array of rotation matrices. I refuse to believe something like this would be slower than python loop.
I felt like this command chain is mostly a tangent, but is there any chance this part of the code would visibly affect performance? If so, I can help with doing numpy magic on 3/4D arrays.

dermen: Thanks, it seems the mmtbx method offers a significant speedup, I made this change:

```
      vec_xyz = RotatePointDegreesAroundAxisDir((0,0,0), polar_vector, -beam_vector, vdiv*180/np.pi)
      unit_s0 = RotatePointDegreesAroundAxisDir((0,0,0), vert_vector, vec_xyz, hdiv*180/np.pi) 
```

If we converted the outer loop to a large matrix multiply and used the scipy.spatial methods, it might be faster still, but I think for now this is good

simtbx/nanoBragg/nanoBragg_beam.py
```diff
@@ -60,40 +94,83 @@ def polarization_fraction(self, val):
...
+    beams = []
+    for hdiv, vdiv in divs:
+      vec_xyz = rotate_axis(-beam_vector, polar_vector, vdiv)
+      unit_s0 = rotate_axis(vec_xyz, vert_vector, hdiv)
+      div_ang = np.arccos(np.dot(unit_s0, -beam_vector))
+      if hdiv == 0 and vdiv == 0:
+        assert div_ang == 0
+        assert np.allclose(unit_s0, nominal_beam.get_unit_s0())
+      if div_ang > 1.1*(self.divergence_mrad / 1000. / 2.):
+        continue
```
Baharis, line 120: One more comment since I burned myself looking for errors in DIALS recently. Can this condition ever be met? If no, wouldn't `raise ValueError` be more fitting?

simtbx/nanoBragg/nanoBragg_beam.py
russell-taylor: @bkpoon It may make sense to move RotatePointDegreesAroundAxisDir into scitbx and out of Reduce if it is going to be useful to other projects. I don't mind it being used elsewhere, but I know you like to keep things organized.

simtbx/nanoBragg/nanoBragg_beam.py
Baharis: Aah, I see now. But in this I have two more ideas:

- Isn't 10% a relatively big tolerance? Something like 1e-6 should be completely enough.
- I would strongly suggest baking this condition in self.divergences definition to avoid future errors. If the corner vectors are not physically valid, why ever have them accessible?

dermen:
I tested and with these changes, the exafel general no GT pipeline works as expected (improved CCanom after stage2). This is good to go for now

PR #969, opened 2024-02-20
MMTBX: remove comparison of int with NoneType

benjaminhwilliams:
Fix a couple of places in `mmtbx.scaling` where the attribute of a class, initially set to `None`, is implicitly expected to have since become an `int`.  There are seemingly valid paths through the code in which the attribute values remain `None`, so this leads to a comparison `TypeError`.  Removing these code paths seems pretty trivial.

@phyy-nx and I stubbed our toes on this last week when trying to run [`xia2.multiplex`](https://dials.github.io/documentation/programs/xia2_multiplex.html), of which MMTBX is a dependency.

I've no idea why tests are failing in the CI.  They don't, at first glance, seem to be related to these changes.  @bkpoon, could you please advise as to how you would like me to proceed?

### DIFF
--- a/mmtbx/scaling/data_statistics.py
+++ b/mmtbx/scaling/data_statistics.py
@@ -862,10 +862,10 @@ or omission of reflections by data-processing software.""")
         issues.append((0,
           "The resolution cutoff appears to be similar in all directions.",
           "Analysis of resolution limits"))
-    if (0 < self.low_resolution_completeness_overall < 0.75):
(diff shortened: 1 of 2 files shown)


benjaminhwilliams:
@phyy-nx, I notice that the only CI failures are XFEL CI.  I don't think I know how to run these locally.  Does this change set concern you?

bkpoon (approved):
Thanks, looks good to me!

benjaminhwilliams:
> Thanks, looks good to me!

Apologies @bkpoon, I missed your review.  Belatedly merging.

PR #977, opened 2024-03-22
Fix miller array `repr` when `label_string()` returns `None`

dagewa:
I came across this whilst running [this script](https://github.com/dials/dials/issues/2625#issuecomment-2015145506) on data in space group $F d d 2$. Simplest solution: if `label_string()` returns `None` just wrap it in a `str()` call.


phyy-nx (approved):
LGTM.  Thanks.

PR #980, opened 2024-03-26
Migrate `gltbx` to use `pyopengl`

### LINKED ISSUE #979: gltbx: Windows failure with Python => 3.8
On Windows, there is an issue with loading the `gltbx` extensions in Python 3.8 and greater. Python 3.7 is not affected.

This is probably due to the change in DLL resolution starting with Python 3.8 (https://github.com/python/cpython/issues/80266). The easiest solution is to migrate to `pyopengl` instead of debugging DLL resolution issues. One consequence is that OpenGL functions from `pyopengl` do not have keyworded arguments, so the keywords will have to be removed (but keep the arguments).

@dagewa Is it possible to add `pyopengl` as a dependency to CCP4?

bkpoon:
Fixes #979

### DIFF
--- a/gltbx/gl.py
+++ b/gltbx/gl.py
@@ -1,10 +1,11 @@
 from __future__ import absolute_import, division, print_function
-# import boost_adaptbx.boost.python as bp
-# ext = bp.import_ext("gltbx_gl_ext")
-# from gltbx_gl_ext import *
-import OpenGL  # implicit import
-from OpenGL.GLU import *
-from OpenGL.GL import *
+try:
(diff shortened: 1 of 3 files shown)


bkpoon:
This will make `cctbx.multiplicity_viewer` work on Windows with Python 3.11.

dagewa:
Confirmed this makes `dials.reciprocal_lattice_viewer` work again on Windows too. Thanks!

bkpoon:
Great! I have to do more testing before I can merge this, but I'll get it in the next release.

PR #982, opened 2024-04-04
Reduce2 updates

russell-taylor:
Makes Reduce2 use the same wrapping of Probe2 that Clashscore2 uses, which is much cleaner.

Factoring out argument construction in reduce_hydrogens.py so that it can be reused in reduce2.py.


russell-taylor:
@bkpoon I'm going to ignore clutter failures that are in parts of the code that I did not change.

PR #983, opened 2024-04-09
Memory policy

nksauter:


dermen:
@nksauter , just making sure I see where we are: in the latest commit, the test works, but there is duplicated code which we want to avoid ?

nksauter:
@dermen stand by please with regard to the working tests.  I'm going to
apply the test to additional cases today to double check.  The duplicated
code is a serious problem, as my original intent was to implement
polymorphism with the minimal lines of code, instead I had to duplicate an
entire kernel.  Further, I was unable to shrink the size of the
m_accumulate_floatimage array as was the original intent.

Baharis (approved):
After 3 weeks of testing via psii_spread's annulus worker, this PR was indeed found to significantly lower the memory footprint when simulating images on GPU. I don't have particular issues with the functionality of this code.

PR #987, opened 2024-04-24
Undowser2

russell-taylor:
The major change in this pull request is a fix to probe2 that avoids contacts between atoms that are in incompatible alternate conformations.

Also added an undowser2_validation.py script that uses reduce2 and probe2 for its tests.


russell-taylor:
@bkpoon It looks like most of the tests did not run on this pull request. It also looks like the automatic-merge check is hung. Closing this pull request and doing another one to see if that breaks things free.

PR #988, opened 2024-04-24
Undowser2

russell-taylor:
The major change is a modification to probe2 and the mmtbx.probe libraries so that they avoid collisions between atoms in incompatible conformations.

This problem was found while adding a new undowser2_validation.py script that uses the new probe and reduce to do its thing.


russell-taylor:
@bkpoon Nope... it is still stuck. How do I get it unstuck?  Thanks!

russell-taylor:
Wow, it looks like we're still requiring C++98.  Okay, will use old-school initializers.

PR #991, opened 2024-05-18
Mpi helper counter

Baharis:
I was working on a custom `cctbx.xfel` worker and found myself `gather`ing + `count`ing objects multiple times. I realized that `mpi_helper` could just as well use a simple `count` method that returns a [`Counter` object](https://docs.python.org/3/library/collections.html#collections.Counter) (a dict-like that automatically counts every occurrence of input list or dict). 

While at it, I realized that the mpi_helper methods could be simplified; in particular, the `flex_type` can be implicitly learned from `flex_array` – as long as the `root` rank has a not-None `flex_array`. Consequently, I allowed the `flex_type` to be inherited from `flex_array` if left undefined.

Initially, I wanted to simplify all instances where `cumulative_flex` and `aggregate_flex` use "redundant" `flex_type`, but since it works in case of completely empty input, I left it as it. After all, _[Explicit is better than implicit](https://peps.python.org/pep-0020/)_.

### DIFF
--- a/xfel/merging/application/mpi_helper.py
+++ b/xfel/merging/application/mpi_helper.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 from collections import Counter
-from six.moves import range
 from libtbx.mpi4py import MPI
 import numpy as np
 from dials.array_family import flex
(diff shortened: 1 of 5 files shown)


Baharis:
Note: I found an implementation of `MPI.reduce.SUM` for counters, but it was rather unsightly, and _Beautiful is better than ugly_.

phyy-nx (approved):

PR #999, opened 2024-07-19
Don't `self.params.input.path = None` after `list_input_pairs` on rank0

Baharis:
This PR removes line `self.params.input.path = None # Rank 0 has already parsed the input parameters` from `xfel/merging/application/input/file_loader.py`. With this change phil parameter `input.path` should be accessible for rank 0 for all downstream workers.

I looked at the code and this change should not affect any existing functionality; however, since theoretically it can indirectly affect all `cctbx.xfel.merge` workers, I am running it via PR to make sure it passes all tests first.


phyy-nx (approved):
The macos test failure started recently and is unrelated.  Currently I can't reproduce it, but it's on my list.  This PR itself seems reasonable to me.  Thanks.

PR #1001, opened 2024-07-30
Sanitize postrefinement results based on scaled intensities

Baharis:
When the `postrefinement` worker in `cctbx.xfel.merge` for any reason introduces an unreasonable scale factor, it might produce a "false negative" experiment that has not scaled correctly and features unphysical values of scaled intensity, but has been accepted by the postrefinement nonetheless. Downstream, this experiment is going to be merged with other "true negatives", resulting in an mtz file with unphysical unscaled intensities and hopeless cc1/2, as shown below for PSII:

```
--------------------------------------------------------------------------------------------------------------------------------------------------------------
                                               <asu    <obs                             Merged         Merged      Unmerged        Unmerged        Unmerged
Bin  Resolution Range   Completeness    %     multi>  multi>   n_meas  asu_m_meas        <I>         <I/sig(I)> mean(I/sig(I)) stddev(I/sig(I)) skew(I/sig(I))
--------------------------------------------------------------------------------------------------------------------------------------------------------------
  1 -1.0000 -  6.1074  [37127/37211]  99.77  1576.64 1580.21  58668337    37127         3675722         27.916        6.230          15.447          10.479
  2  6.1074 -  4.8475  [37163/37163]  100.00  741.28  741.28  27548209    37163         1585893         17.274        2.547           5.168           5.820
  3  4.8475 -  4.2347  [37245/37245]  100.00  633.91  633.91  23609935    37245         1842507         16.254        2.735           5.420           5.551
  4  4.2347 -  3.8474  [37203/37203]  100.00  482.78  482.78  17960727    37203         1227655         12.346        1.782           3.705           5.823
  5  3.8474 -  3.5717  [37192/37192]  100.00  384.55  384.55  14302217    37192         1018178          9.492        1.214           2.577           5.396
  6  3.5717 -  3.3611  [37182/37182]  100.00  323.78  323.78  12038700    37182         1956941          7.650        0.913           1.952           4.455
  7  3.3611 -  3.1927  [37246/37246]  100.00  229.19  229.19  8536296     37246        23028052          5.676        0.724           1.696           2.942
  8  3.1927 -  3.0537  [37209/37209]  100.00  139.46  139.46  5189297     37209       524052171          3.903        0.604           1.550           2.782
  9  3.0537 -  2.9361  [37094/37094]  100.00   98.76   98.76  3663416     37094       5751102194         3.264        0.595           1.456           2.101
[...]

PR #1003, opened 2024-08-01
"Truncate" XFEL expts/refls based on aggregated statistics

Baharis:
This PR ports the changes discussed in PR #1001 outside postrefinement into a dedicated `truncate` worker. In contrast to `postrefine` or `filter` workers, the `truncate` worker is strictly designed to act on the entire dataset and utilize global communication instead of refraining from using it. For the time being, it utilizes only the "intensity extrema threshold" mechanism described in PR #1001; however, it is strictly designed to be expanded by additional "truncation" criteria.

For more details on the mechanism behind the rejection mechanism, refer to #1001.

Bugfixes and code clean-up introduced to the `postrefine` worker are not moved here and have been made into a separate PR #1002 instead.


phyy-nx (approved):

Baharis:
Example output (from new XFEL regression test):
```
STEP 1: Read experiments and data
Experiments: (total,min,max): 33, 33, 33
Images:      (total,min,max): 33, 33, 33
Reflections: (total,min,max): 3561, 3561, 3561

STEP 2: Filter expts & refls based on their aggregated statistics
Experiments/reflections truncated globally due to:
- Intensity extremum outside IQR dist threshold:      7 expt,      1142 refl
- TOTAL:                                              7 expt,      1142 refl
Experiments: (total,min,max): 26, 26, 26
Images:      (total,min,max): 26, 26, 26
Reflections: (total,min,max): 2419, 2419, 2419
```

nksauter:
I'm concerned that the name "Truncate" conflicts with another well-known algorithm in protein crystallography implemented about 50 years ago and used ever since; namely the French-Wilson treatment that makes all merged intensities positive in the merging output.  I'd favor choosing another name to avoid the conflict.

Baharis:
I renamed and moved the `truncate` worker into `filter_global`. The only problem with this name is that `global` is a reserved name in python and, as such, the corresponding phil scope for this worker must not have a `.global.` segment in it. I solved it using the standard Python practice of adding an underscore, i.e. the threshold parameter is not `filter.global_.intensity_extrema_iqr_dist_threshold`, but it can be unintuitive; another approach would be to give this worker its own phil scope, e.g. `filter_global.intensity_extrema_iqr_dist_threshold`, similarly to how experiment worker and reflection worker use `filter` and `select`, accordingly, but this might be even more confusing.

Baharis:
Following Nick's remark on yesterday's SPREAD meeting, I plotted the distribution of max(intensity sum) versus min(intensity sum) for all 111k indexed PSII experiments to see how it looks. Since most minima are negative, I used an absolute value on the Y axis, with red points representing minima > 0 and blue points representing minima < 0. As you can see, the distribution of extrema is roughly log-normal (suggesting that, if we care enough, the threshold limit would also be better set in the log space). Surprisingly, despite ~60 expts being removed in this step, I could find only 3 that clearly surpass the threshold for maxima.
image

PR #1005, opened 2024-08-14
New cctbx.xfel.merge worker for computing and filtering on ΔCC½

phyy-nx:
This is a means of filtering out lattices that degrade the overall CC½. Uses the σ-τ method from [Assmann 2016](https://journals.iucr.org/j/issues/2016/03/00/zw5005/#FD4) to avoid splitting the data into odd/even datasets. Enable by adding the deltaccint worker after the group worker.

Includes parameter iqr_ratio = 10. If the ΔCC½ filter is enabled, first compute CC½ when removing every image one at a time, then compute the IQR of these CC½s.  Remove all lattices whose contribution degrades CC½ by more than IQR * iqr_ratio above the median. You can discover a good IQR by running the program once, examining the log file where possible values are listed, and running it again with the best value.

Implementation heavily uses MPI for fast computation of ΔCC½ on thousands of images.  Current run time with this version is ~5 minutes for 8.8k images.

Example output in the log for a PSII dataset:

```
STEP 11: ΔCC½ statistics (σ-τ method)
Beginning ΔCC½ analysis (σ-τ method from Assmann 2016)
Removing reflections with less than 3 measurements
N experiments after filtering: 8816

Five number summary
 91.3091% min
 91.3109% q1
 91.3111% median
 91.3114% q3
 91.4728% max

Showing ΔCC½ of the worst 30 lattices and IQR ratio needed to remove them
 ΔCC½ (%) IQR ratio Lattices removed
 91.3172       11.7 30
 91.3174       12.1 29
 91.3175       12.3 28
 91.3175       12.4 27
 91.3178       12.9 26
 91.3179       13.1 25
 91.3182       13.7 24
 91.3183       13.8 23
 91.3184       14.1 22
 91.3185       14.2 21
 91.3188       14.8 20
 91.3192       15.5 19
 91.3199       16.9 18
 91.3200       17.1 17
 91.3215       20.1 16
 91.3218       20.5 15
 91.3219       20.7 14
 91.3222       21.3 13
 91.3350       45.9 12
 91.3403       56.1 11
 91.3403       56.2 10
 91.3404       56.3 9
 91.3407       56.9 8
 91.3408       57.1 7
 91.3409       57.3 6
 91.3410       57.6 5
 91.3413       58.1 4
 91.3414       58.2 3
 91.3649      103.5 2
 91.4728      311.1 1

Removing 29 experiments out of 8816 using IQR ratio 12.0
```

### DIFF
--- a/xfel/merging/application/statistics/deltaccint.py
+++ b/xfel/merging/application/statistics/deltaccint.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 from six.moves import range
 from dxtbx import flumpy
 from dials.array_family import flex
+import itertools
 import numpy as np
 from scitbx.math import five_number_summary
 from xfel.merging.application.reflection_table_utils import reflection_table_utils
@@ -28,16 +29,7 @@ class deltaccint(worker):
       if len(set(refls['id'])) >= min_mult:
         filtered.extend(refls)
 
-    # Get list of all experiment identifiers
-    expt_ids = comm.gather(filtered.experiment_identifiers().values(), 0)
-    if self.mpi_helper.rank == 0:
-      all_expt_ids = flex.std_string()
-      for expt_ids_ in expt_ids:
-        all_expt_ids.extend(expt_ids_)
-      expt_ids = list(set(all_expt_ids))
-    else:
-      expt_ids = None
-    all_expt_ids = comm.bcast(expt_ids, 0)
+    all_expt_ids = list(sorted(set(itertools.chain.from_iterable(comm.allgather(filtered.experiment_identifiers().values())))))
     all_expts_map = dict(zip(all_expt_ids, range(len(all_expt_ids))))
 
     if self.mpi_helper.rank == 0:
--- a/xfel/merging/application/statistics/deltaccint.py
+++ b/xfel/merging/application/statistics/deltaccint.py
@@ -29,7 +29,7 @@ class deltaccint(worker):
       if len(set(refls['id'])) >= min_mult:
         filtered.extend(refls)
 
-    all_expt_ids = list(sorted(set(itertools.chain.from_iterable(comm.allgather(filtered.experiment_identifiers().values())))))
+    all_expt_ids = sorted(set(itertools.chain.from_iterable(comm.allgather(filtered.experiment_identifiers().values()))))
     all_expts_map = {v: k for k, v in enumerate(all_expt_ids)}
 
     if self.mpi_helper.rank == 0:
@@ -45,7 +45,7 @@ class deltaccint(worker):
     resolution_binner = self.params.statistics.resolution_binner
     hkl_resolution_bins = self.params.statistics.hkl_resolution_bins
 
-    hkl_set = [hkl for hkl in list(set(filtered['miller_index_asymmetric'])) if hkl in hkl_resolution_bins]
+    hkl_set = [hkl for hkl in set(filtered['miller_index_asymmetric']) if hkl in hkl_resolution_bins]
     n_hkl = len(hkl_set)
     hkl_map = {v: k for k, v in enumerate(hkl_set)}
 
@@ -100,16 +100,13 @@ class deltaccint(worker):
       all_i_n     [:,bin_idx] += 1
       all_var_sums[:,bin_idx] += variance[:,hkl_map[hkl]]
 
-    total_i_sums   = comm.reduce(all_i_sums,   MPI.SUM, 0)
-    total_i_n      = comm.reduce(all_i_n,      MPI.SUM, 0)
     total_var_sums = comm.reduce(all_var_sums, MPI.SUM, 0)
 
     # Broadcast the average intensities
-    if self.mpi_helper.rank == 0:
-      total_i_average_ = total_i_sums / total_i_n
-    else:
-      total_i_average_ = None
-    total_i_average = comm.bcast(total_i_average_, 0)
+    total_i_sums   = comm.allreduce(all_i_sums, op=MPI.SUM)
+    total_i_n      = comm.allreduce(all_i_n,    op=MPI.SUM)
+    total_var_sums = comm.reduce(all_var_sums,  op=MPI.SUM)
+    total_i_average = total_i_sums / total_i_n
 
     # Compute the variance of the average intensities
     # First, the numerator, the difference between each hkl and the average for that hkl's bin (ommiting each experiment once)
--- a/xfel/merging/application/statistics/deltaccint.py
+++ b/xfel/merging/application/statistics/deltaccint.py
@@ -163,7 +163,7 @@ class deltaccint(worker):
       self.logger.main_log("")
 
       if self.params.statistics.deltaccint.iqr_ratio:
-        cut = med + iqr * self.params.statistics.deltaccint.iqr_ratio
+        cut = q3 + iqr * self.params.statistics.deltaccint.iqr_ratio
         sel = data > cut
         worst_expts_ = flex.std_string(all_expt_ids).select(sel)
 
--- a/xfel/merging/application/statistics/deltaccint.py
+++ b/xfel/merging/application/statistics/deltaccint.py
@@ -30,7 +30,7 @@ class deltaccint(worker):
         filtered.extend(refls)
 
     all_expt_ids = list(sorted(set(itertools.chain.from_iterable(comm.allgather(filtered.experiment_identifiers().values())))))
-    all_expts_map = dict(zip(all_expt_ids, range(len(all_expt_ids))))
+    all_expts_map = {v: k for k, v in enumerate(all_expt_ids)}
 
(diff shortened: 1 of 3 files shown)


phyy-nx:
[Test added to xfel_regression](https://gitlab.com/cctbx/xfel_regression/-/commit/21965cba3fa4181d83724ef050d0ba381a254f13)

Baharis:
I have a note regarding the terminology used here (and one other place in cctbx) to measure distance in IQR. IQR (Inter-Quartile Range) is defined as the distance between the values of the 1st and 3rd quartile. "IQR ratio" to me reads like a ratio applied to the IQR "segment" to determine the acceptance range. If that would be the case, "IQR ratio" of 1 should (exactly) encompass all reflections across the inter-quartile space, i.e. between 1st and 3rd quartile:
```
iqr_ratio=
             <----1--->
q0__________q1___q2___q3__________q4
```
This is not the case; IQR ratio of 1 translates to "Distance of 1 from median, expressed in IQR units". Consequently, the space between q1 and q3 is (rougly) described by `iqr_ratio=0.5`, whereas `iqr_ratio=1` marks a 2x larger area:
```
iqr_ratio=
        <---------1-------->
             <---.5--->
q0__________q1___q2___q3__________q4
```
I checked and this wording is consistent with its one other use in `cctbx.xfel`. However, I am wondering whether migrating to a more descriptive name (such as the `iqr_dist_treshold` I used in PR #1003) would not be more readable. It is arguably longer, and the parameter is ultimately just a number with a relatively irrelevant physical meaning, but maybe it would be easier to understand in the long-term whenever someone is interested in the exact definition or the physical meaning.

xfel/merging/application/statistics/deltaccint.py
```diff
@@ -0,0 +1,190 @@
...
+
+    # Get list of all experiment identifiers
+    expt_ids = comm.gather(filtered.experiment_identifiers().values(), 0)
+    if self.mpi_helper.rank == 0:
+      all_expt_ids = flex.std_string()
+      for expt_ids_ in expt_ids:
+        all_expt_ids.extend(expt_ids_)
+      expt_ids = list(set(all_expt_ids))
+    else:
+      expt_ids = None
```
Baharis, line 39: All of this can be simplified as:
```
expt_ids = list(set(self.mpi_helper.aggregate_flex(filtered.experiment_identifiers().values(), 0)))
```
or
```
expt_ids = list(self.mpi_helper.count(filtered.experiment_identifiers().values(), 0).keys())
```

Baharis: In fact, I was thinking about writing a small PR so that the mpi_helper uses `allgather` instead of `gather` if `root=None`. This would also remove a need for the second communication on line 40, making things more readable and faster.

xfel/merging/application/statistics/deltaccint.py
```diff
@@ -0,0 +1,190 @@
...
+    all_expt_ids = comm.bcast(expt_ids, 0)
+    all_expts_map = dict(zip(all_expt_ids, range(len(all_expt_ids))))
```
Baharis, line 41: This is mostly a matter of style, but I like to write these reverse-maps as dict comprehension: `all_expts_map = {v: k for k, v in enumerate(all_expt_ids)}`.

xfel/merging/application/statistics/deltaccint.py
```diff
@@ -0,0 +1,190 @@
...
+      self.logger.main_log("")
+
+    # We need to compute
+    # 1) variance of the average intensities -> compute averages of all intensities and then compute variance per bin
+    # 2) average variance of the intensities -> compute variances of all intensities and then compute average per bin
+
+    resolution_binner = self.params.statistics.resolution_binner
+    hkl_resolution_bins = self.params.statistics.hkl_resolution_bins
+
+    hkl_set = [hkl for hkl in list(set(filtered['miller_index_asymmetric'])) if hkl in hkl_resolution_bins]
```
Baharis, line 56: Fun fact: as wrong as it feels, in Python you can iterate over sets so you don't need the `list(...)`.

xfel/merging/application/statistics/deltaccint.py
```diff
@@ -0,0 +1,190 @@
...
+    total_i_sums   = comm.reduce(all_i_sums,   MPI.SUM, 0)
+    total_i_n      = comm.reduce(all_i_n,      MPI.SUM, 0)
+    total_var_sums = comm.reduce(all_var_sums, MPI.SUM, 0)
+
+    # Broadcast the average intensities
+    if self.mpi_helper.rank == 0:
+      total_i_average_ = total_i_sums / total_i_n
+    else:
+      total_i_average_ = None
+    total_i_average = comm.bcast(total_i_average_, 0)
```
Baharis, line 120: All of this can be simplified using and sped up (probably unnoticable?) using `comm.allreduce`:
```
total_i_sums   = comm.allreduce(all_i_sums, op=MPI.SUM)
total_i_n      = comm.allreduce(all_i_n,    op=MPI.SUM)
total_var_sums = comm.reduce(all_var_sums,  op=MPI.SUM)
total_i_average = total_i_sums / total_i_n
```

Baharis:
One last comment: is ~5 minutes for 8.8k images manageable? Because I'd bet you can vectorize the loops over refls, but I don't know whether (1) these are the bottleneck, (2) it's worth it, because it certainly wouldn't be easy 😅 .

Baharis (approved):

xfel/merging/application/statistics/deltaccint.py
```diff
@@ -0,0 +1,190 @@
...
+      self.logger.main_log("")
+
+      n_worst = min(len(data), 30)
+      worst = sorted_data[-n_worst:]
+      iqr = q3-q1
+
+      self.logger.main_log("Showing ΔCC½ of the worst %d lattices and IQR ratio needed to remove them"%n_worst)
+      self.logger.main_log(" ΔCC½ (%) IQR ratio Lattices removed")
+      for i in range(len(worst)):
+        self.logger.main_log("% 8.4f % 10.1f %d"%(worst[i], (worst[i]-med)/iqr, n_worst-i))
```
Baharis, line 162: How about logging the "worst" in reverse order so that the biggest offender appears at the very beginning? I think it would be more intuitive:
```
      for i, value in enumerate(reversed(worst)):
        self.logger.main_log("% 8.4f % 10.1f %d"%(value, (value-med)/iqr, i+1))
```

Baharis: This will produce the following table (props to uncle GPT for reversing 😛 ):
```
Showing ΔCC½ of the worst 30 lattices and IQR ratio needed to remove them
 ΔCC½ (%) IQR ratio Lattices removed
 91.4728      311.1 1
 91.3649      103.5 2
 91.3414       58.2 3
 91.3413       58.1 4
 91.3410       57.6 5
 91.3409       57.3 6
 91.3408       57.1 7
 91.3407       56.9 8
 91.3404       56.3 9
 91.3403       56.2 10
 91.3403       56.1 11
 91.3350       45.9 12
 91.3222       21.3 13
 91.3219       20.7 14
 91.3218       20.5 15
 91.3215       20.1 16
 91.3200       17.1 17
 91.3199       16.9 18
 91.3192       15.5 19
 91.3188       14.8 20
 91.3185       14.2 21
 91.3184       14.1 22
 91.3183       13.8 23
 91.3182       13.7 24
 91.3179       13.1 25
 91.3178       12.9 26
 91.3175       12.4 27
 91.3175       12.3 28
 91.3174       12.1 29
 91.3172       11.7 30
```

Baharis: Also, maybe I misunderstood but isn't this table misleading? We don't get  ΔCC½=91.3172 by removing 30 worst lattices, we get it by removing the 30th lattice only – in which case the last column should say something along the lines of "-th worst lattice removed" instead of "Lattices removed". Otherwise this seems to suggest that removing more and more lattices lowers ΔCC½, which I hope is not the case.

xfel/merging/application/statistics/deltaccint.py
phyy-nx: allgather works!

xfel/merging/application/statistics/deltaccint.py
phyy-nx: Heh, I like this one as it is, but I think it's a matter of taste.

phyy-nx:
> One last comment: is ~5 minutes for 8.8k images manageable? Because I'd bet you can vectorize the loops over refls, but I don't know whether (1) these are the bottleneck, (2) it's worth it, because it certainly wouldn't be easy 😅 .

I started looking at the critical loops in c++ and it's doable, but no need to hold this for that.  The use case here is the delta cc int method would only be used in situations where the user needs to find some bad images, which happens but isn't as common.  Hence not adding this to the default worker list.

PR #1009, opened 2024-09-03
Ribbons2

russell-taylor:
Marks nucleic acids as helices and draws them even wider than alpha helices by default.

Has an option to color all chains solid rather than rainbow or the first by secondary structure.


russell-taylor:
The failed checks are in files I did not touch, so I'm a bit hesitant to modify them to fix the clutter for fear of breaking someone else's code.

PR #1020, opened 2024-09-27
Replace (unmaintained) orderedset with ordered_set

rolandmas:
Hi, it seems that the orderedset module is abandoned by its author, and it started failing; in particular with Python 3.13 as discovered in Debian, but see also https://github.com/simonpercivall/orderedset/issues/36

There's an API-compatible replacement called ordered_set, see https://github.com/rspeer/ordered-set

This patch replaces the one with the other. It's already been used as part of the Debian packaging of cctbx with no influence on the build or the testsuite. I suggest applying it to upstream cctbx.


bkpoon:
The `libtbx` changes are fine. Those are either deprecated or for building other things. I added reviewers for the `xfel` stuff.

olegsobolev:
Maybe [libtbx/containers.py](https://github.com/cctbx/cctbx_project/blob/78a106ff3f974536f5e663c0516de470d1823983/libtbx/containers.py) can cover the functionality, and there won't be a need for just another dependency? Or should we consider removing cctbx implementation for the sake of consistency?

PR #1030, opened 2024-12-02
Xfel merge filter by resolution n obs

dwmoreau:
Implemented experiment filters by resolution and number of observations within cctbx.xfel.merge


xfel/merging/application/filter/experiment_filter.py
```diff
@@ -69,18 +60,89 @@ def check_cluster(self, experiment):
     return m_distance < self.params.filter.unit_cell.cluster.covariance.mahalanobis
 
   def run(self, experiments, reflections):
-    if 'unit_cell' not in self.params.filter.algorithm: # so far only "unit_cell" algorithm is supported
+    filter_by_unit_cell = 'unit_cell' in self.params.filter.algorithm[0]
+    filter_by_n_obs = 'n_obs' in self.params.filter.algorithm[0]
+    filter_by_resolution = 'resolution' in self.params.filter.algorithm[0]
+    # only "unit_cell" "n_obs" and "resolution" algorithms are supported
+    if (not filter_by_unit_cell) and (not filter_by_n_obs) and (not filter_by_resolution):
```
phyy-nx, line 67: These validation steps should likely be in the validate function.  validate runs before the program starts execution, which will save time if the parameters are invalid.

xfel/merging/application/phil/phil.py
```diff
@@ -129,14 +129,18 @@
   .help = The filter section defines criteria to accept or reject whole experiments
   .help = or to modify the entire experiment by a reindexing operator
   .help = refer to the select section for filtering of individual reflections
+  .help = only unit_cell, resolution, and n_obs are implemented
   {
   algorithm = n_obs reindex resolution unit_cell report
-    .type = choice
+    .type = strings
     .multiple = True
```
phyy-nx, line 136: If .type = choice and .multiple = true, then params.filter.algorithm should be a list of already validated strings.  Is there a reason that doesn't work?

xfel/merging/application/filter/experiment_filter.py
```diff
@@ -69,18 +60,89 @@ def check_cluster(self, experiment):
...
+      self.logger.main_log("Total experiments rejected because of space group %d"%total_removed_for_space_group)
+      self.logger.main_log("Total experiments rejected because of n_obs %d"%total_removed_for_n_obs)
+      self.logger.main_log("Total experiments rejected because of resolution %d"%total_removed_for_resolution)
+      self.logger.main_log("Total reflections rejected because of rejected experiments %d"%total_reflections_removed)
+
+    self.logger.log_step_time("FILTER_EXPERIMENTS", True)
+
+    # Do we have any data left?
+    from xfel.merging.application.utils.data_counter import data_counter
+    data_counter(self.params).count(new_experiments, new_reflections)
```
phyy-nx, line 137: data_counter result not checked?

xfel/merging/application/phil/phil.py
```diff
@@ -129,14 +129,18 @@
   .help = The filter section defines criteria to accept or reject whole experiments
   .help = or to modify the entire experiment by a reindexing operator
   .help = refer to the select section for filtering of individual reflections
+  .help = only unit_cell, resolution, and n_obs are implemented
```
phyy-nx, line 132: Let's go ahead and drop reindex and report if they aren't implemented.

phyy-nx (changes requested):
Looks like a good feature.  Some review comments added.

xfel/merging/application/phil/phil.py
dwmoreau: reindex and report have been removed from the phil file and experiment_filter.py

xfel/merging/application/filter/experiment_filter.py
dwmoreau: The problem was that with multiple filters, each filter can flag the same experiment. I use a list(set()) operation to get the unique experiment id's and use the length of this list to get the number of unique experiments that are removed.

xfel/merging/application/filter/experiment_filter.py
dwmoreau: I moved these into the validate method. If the validation fails because:

- No algorithm in the algorithms list: the original experiments and reflections are returned.
- Error in input parameters: assertion fail.

dwmoreau:
Test failures unrelated to feature.

Issue #1033, opened 2024-12-15
Unit cell parameter is zero or negative


frgoe003:
Hello, I'm trying to run the reduce2 CLI with the flags `approach=add`, `add_flip_movers=True`, but for some systems I keep getting: `ValueError: Unit cell parameter is zero or negative.` Here are two examples of systems I am trying to process with reduce2: `8j3c`, `4rsy`

Here is the full traceback: 
```
File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/mmtbx/command_line/reduce2.py", line 6, in <module> <br>
    run_program(reduce2.Program) <br>
  File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/iotbx/cli_parser.py", line 946, in run_program <br>
    task.run() <br>
  File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/mmtbx/programs/reduce2.py", line 1191, in run <br>
    self.model = shift_and_box_model(model=self.model, shift_model=False) <br>
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ <br>
  File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/cctbx/maptbx/box.py", line 1338, in shift_and_box_model <br>
    crystal_symmetry = crystal.symmetry((a, b, c, 90, 90, 90), 1) <br>
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ <br>
  File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/cctbx/crystal/__init__.py", line 70, in __init__ <br>
    unit_cell = uctbx.unit_cell(unit_cell) <br>
                ^^^^^^^^^^^^^^^^^^^^^^^^^^ <br>
  File "/home/fgoerlich/miniconda3/envs/docking/lib/python3.11/site-packages/cctbx/uctbx/__init__.py", line 46, in __init__ <br>
    ext.unit_cell.__init__(self, <br>
ValueError: Unit cell parameter is zero or negative.

```

russell-taylor:
It looks like this is caused by shift_and_box_model() being called with shift_model=False and one of the atom coordinates for all atoms is negative.  Here is a test PDB file that demonstrates the problem:

`
[...]

PR #1039, opened 2025-01-13
Computing the length of the crystal box by subtracting its two ends a…

russell-taylor:
…nd adding cushion on each end rather than by assuming that the origin is zero or cushion above zero.  This makes the box side lengths always positive even when all atoms are on the negative side of one of the axes and the shift_model parameter to shift_and_box_model is False


russell-taylor:
@bkpoon Can you or someone run the full test suite on this fix for https://github.com/cctbx/cctbx_project/issues/1033? I'm pretty sure it is right but it could screw things up if there is a corner case.

PR #1044, opened 2025-02-11
Added fable docker

dand-oss:
Dockerfile to build and run fable
described in README

### DIFF
--- a/fable/docker/Dockerfile
+++ b/fable/docker/Dockerfile
@@ -1,0 +1,39 @@
+# syntax=docker/dockerfile:1.4
+# use latest debian
+FROM debian:sid-slim
+
+ENV RUNLEVEL 1
+ENV DEBIAN_FRONTEND noninteractive
+ENV LC_ALL C.UTF-8
+
+RUN <<EOF
+#!/usr/bin/env bash
+set -exo pipefail
+
+# update OS
+apt-get -y update
+apt-get -y upgrade
+
+# install cctbx
+declare -ar pkgs=(
+    # seems to be needed
+    #10 70.95 Setting up dbus-system-bus-common (1.16.0-1) ...
+    #10 70.96 Can't locate Encode.pm in @INC (you may need to install the Encode module)
+    'libencode-perl'
+    # this is huge, qt, hdf...
+    'python3-cctbx'
+    # fable wants to compile c++
+    'g++'
+)
+apt-get -y --no-install-recommends install "${pkgs[@]}"
+
+# run fable to check works
+fable.cout --example
+
+# docker cleanup
+apt-get -y autoremove --purge
+rm -rf /var/lib/apt/lists/*
+EOF
+
+# /usr/bin
+ENTRYPOINT ["fable.cout"]
--- a/fable/docker/README.md
+++ b/fable/docker/README.md
@@ -1,0 +1,44 @@
+
+docker to run fable FORTRAN converter
+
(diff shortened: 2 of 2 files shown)


bkpoon:
Alternatively, you can use the Debian cctbx package for bookworm (https://packages.debian.org/bookworm/python3-cctbx). That should have `fable` and looks like it is using Python 3.11. The Boost sources in our source is for supporting older linux distributions. You should be able to use the Boost libraries in Debian for building. You do need to delete the `modules/boost` directory before building, though.

dand-oss:
> Alternatively, you can use the Debian cctbx 

Hi, simplified Dockerfile as you suggested.  Now pushed update as

# syntax=docker/dockerfile:1.4
FROM debian:sid-slim

ENV RUNLEVEL 1
ENV DEBIAN_FRONTEND noninteractive
ENV LC_ALL C.UTF-8

RUN <<EOF
#!/usr/bin/env bash
set -exo pipefail

# update OS
apt-get -y update
apt-get -y upgrade

# install cctbx
declare -ar pkgs=(
    # this is huge, qt, hdf...
    'python3-cctbx'
    'g++'
)
apt-get -y --no-install-recommends install "${pkgs[@]}"

# run fable to check works
fable.cout --example

# docker cleanup
apt-get -y autoremove --purge
rm -rf /var/lib/apt/lists/*
EOF

# /usr/bin
ENTRYPOINT ["fable.cout"]

bkpoon:
I tried following the instructions in the README, but the image does not build. You can see the output in this workflow

https://github.com/cctbx/cctbx_project/actions/workflows/test_pr.yml

Also, what is the purpose of having the `--build-arg=BUILD_PARENT_DOCKPATH=python:3.9-slim` argument since the `Dockerfile` does not have a `BUILD_PARENT_DOCKPATH` ARG defined? 

Please make sure that the commands in the `README.md` file work and provide explanations for any user provided input (e.g. use something like `<user fortran file>` instead of `fort1.for`).

dand-oss:
Hi Billy thanks for testing

I updated the README according to your instructions

I also ran each command to ensure they work (on my machine!)

bkpoon:
I'm still getting a failure to build the Docker image. The error message is copied below.
```
#10 70.95 Setting up dbus-system-bus-common (1.16.0-1) ...
#10 70.96 Can't locate Encode.pm in @INC (you may need to install the Encode module) (@INC entries checked: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.40.1 /usr/local/share/perl/5.40.1 /usr/lib/x86_64-linux-gnu/perl5/5.40 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.40 /usr/share/perl/5.40 /usr/local/lib/site_perl) at /usr/sbin/adduser line 27.
```
The error can be fixed by installing the `libencode-perl` package with
```
apt-get -y install libencode-perl
```
after the OS update/upgrade. I was able to test the other commands in `README.md` once I added that to the Dockerfile, so this can be merged after the Dockerfile update.

dand-oss:
Good work... Wonder why mine is different...

dand-oss:
ok, added the lib, ran the build instruction, it worked for me.

You might try the build again.

The parent docker "sid" changes daily, so a "build error" could creep in.

However, if one wants the latest cctbx, package, this will pick it up as soon as the package appears.

PR #1049, opened 2025-03-09
add some missing functions to intrinsics.hpp

dand-oss:
these are in intrinsics.py but not intrinsics.hpp


bkpoon:
I'm traveling so I won't be able to look at this until next week.

bkpoon:
Thanks for filling in the gaps! Did you use `fable` for something and came across a problem with these missing functions?

PR #1060, opened 2025-04-14
Refactor pydiscamb interface

viljarjf:
Hello!

This PR aims to make use of the existing structure factor + gradient framework in cctbx when using TAAM with pydiscamb.

I refactored the algorithm selection slightly, and added TAAM as an option. Currently, the actual implementation is in pydiscamb: https://github.com/viljarjf/pyDiSCaMB/blob/main/pydiscamb/cctbx_interface.py
I would like some input on this solution, if maybe the `gradients_taam` and `from_scatterers_taam` classes should be in cctbx and the results classes in pydiscamb?
The runtime measurement still needs to be added to the taam classes, regardless.

Extra parameters can be passed to DiSCaMB by means of a phil scope. I put this in pydiscamb, to more easily support / update the options if and when they change in DiSCaMB. The `extra`-scope is meant to allow for other algorithms in the future, with a `discamb` subscope defined in pydiscamb. Here, I also have the option to add more parameter sets for other available calculators in DiSCaMB.

Happy for any input on this! I have not been able to test this much, as I'm having difficulties properly installing cctbx.

### DIFF
--- a/cctbx/miller/__init__.py
+++ b/cctbx/miller/__init__.py
@@ -1585,7 +1585,8 @@ class set(crystal.symmetry):
                                         u_base=None,
                                         b_base=None,
                                         wing_cutoff=None,
-                                        exp_table_one_over_step_size=None):
+                                        exp_table_one_over_step_size=None,
+                                        extra_params=None):
     """
     Calculate structure factors for an :py:class:`cctbx.xray.structure` object
     corresponding to the current set of Miller indices.  Can use either FFT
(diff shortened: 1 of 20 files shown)


viljarjf:
Looks like windows has some permission error for the atom assignment file. I only test locally on linux so this slipped by. I'll look into it

viljarjf:
Windows tests should work now, as of https://github.com/viljarjf/pyDiSCaMB/pull/36. I set up windows and macOS testing for pydiscamb as well

bkpoon:
Thanks! We see the other commits passing now.

We're still discussing the implementation, but we are pretty much on board with it.

PR #1070, opened 2025-05-01
Stop attempting "legacy editable" installs

dwpaley:
Legacy-editable mode was removed in pypa/setuptools#4953, this env
variable no longer does anything. We accommodate the new behavior in
cctbx/dxtbx#807 and dials/dials#2906.

### DIFF
--- a/.azure-pipelines/xfel/unix-conda-build.yml
+++ b/.azure-pipelines/xfel/unix-conda-build.yml
@@ -105,8 +105,6 @@ steps:
     conda activate $(PYTHON_VERSION)
     cd $(Pipeline.Workspace)
     source ./build/setpaths.sh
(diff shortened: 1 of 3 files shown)


dwpaley:
We can't get rid of legacy editable mode because some builders are pinned to previous versions of dials and dxtbx. Those old versions still expect (in `libtbx_refresh.py`) to have been installed in legacy editable mode, therefore they add e.g. `dials/src` to the pythonpath instead of `build/dials/lib/python3.x/site-packages`. For those builders (at least cctbx and phenix) we will just have to keep setuptools <79 as long as the dials and dxtbx versions remain pinned. For this PR I'll just remove Aaron's workaround in the tests, and scrap the other part for the foreseeable.

PR #1073, opened 2025-05-06
Fix the parsing of command-line lock-down parsing in reduce2 to handl…

russell-taylor:
…e the Unflipped case properly.


russell-taylor:
@bkpoon The python3.9 check seems to be stuck on an old copy. It has done this for the past two pull requests. The others all work.

russell-taylor:
**Original MolProbity keeps the original conformation and does not flip.** When you run Probe on it, the counts of the scores are consistent with it summing the H bond scores (and perhaps clashes, of which there are none) to determine whether to flip or not.

PR #1099, opened 2025-09-16
remove simtbx dependency from xfel

dermen:


xfel/ui/components/ebeam_plotter.py
```diff
@@ -4,7 +4,8 @@
 from xfel.cxi.cspad_ana import cspad_tbx
 from matplotlib import pyplot as plt
 import numpy as np
-from simtbx.nanoBragg.utils import ENERGY_CONV
+from scipy import constants
+ENERGY_CONV = 10000000000.0 * constants.c * constants.h / constants.electron_volt
```
phyy-nx, line 8: Could this be `factor_kev_angstrom` or `factor_ev_angstrom`?  E.g. `from cctbx import factor_kev_angstrom`

dermen:
```
In [10]: utils.ENERGY_CONV, factor_kev_angstrom*1000
Out[10]: (12398.419843320027, 12398.419843320025)
```

dermen: Id say either way

phyy-nx (approved):

xfel/merging/application/phil/phil.py
```diff
@@ -735,11 +735,14 @@
 }
 """
 
-diffbragg_phil = """
-diffBragg {
-  include scope simtbx.diffBragg.phil.phil_scope
-}
-"""
+try:
```
nksauter, line 738: Unless I'm missing something this change should have no effect.  All this code does is define the string diffbragg_phil; it will never generate an importerror because the import is done at a later time when the includes are evaluated.

phyy-nx (changes requested):

Issue #1127, opened 2026-02-09
TypeError: '>=' not supported between instances of 'NoneType' and 'int':


sprock:
Building on gentoo linux using python3.8:

`(python3.8) forsterite cctbx python bootstrap.py --use-conda --python=38`

the build fails like this:

```
===== Running in build: libtbx.scons
scons: Reading SConscript files ...
TypeError: '>=' not supported between instances of 'NoneType' and 'int':
  File "/home/rmason/Software/CCTBX/cctbx/build/SConstruct", line 6:
    SConscript("libtbx/SConscript")
  File "/home/rmason/Software/CCTBX/cctbx/conda_base/lib/python3.8/site-packages/SCons/Script/SConscript.py", line 660:
    return method(*args, **kw)
  File "/home/rmason/Software/CCTBX/cctbx/conda_base/lib/python3.8/site-packages/SCons/Script/SConscript.py", line 597:
    return _SConscript(self.fs, *files, **subst_kw)
  File "/home/rmason/Software/CCTBX/cctbx/conda_base/lib/python3.8/site-packages/SCons/Script/SConscript.py", line 285:
    exec(compile(scriptdata, scriptname, 'exec'), call_stack[-1].globals)
  File "/home/rmason/Software/CCTBX/cctbx/modules/cctbx_project/libtbx/SConscript", line 658:
    if (env_etc.gcc_version >= 40400 and determine_cpp0x_flag()):
```

The problem may be that libtbx.env_config.get_gcc_version(command_name=cc) at modules/cctbx_project/libtbx/SConscript line 640 returns nothing.

PR #1129, opened 2026-02-15
Guard gcc_version comparisons

dwpaley:
get_gcc_version() returns None when the compiler is not gcc or version detection fails. Therefore check for None before doing numerical comparisons. This was already done in many places but this PR catches a few more.

Fixes cctbx/cctbx_project#1127


dwpaley:
@bkpoon ok to merge this if tests pass?

PR #1131, opened 2026-02-18
Register Boost.Python converters for NumPy 2.0 scalar types

### LINKED ISSUE #1084: += operator fails for flex arrays when adding numpy-typed values
Example:
```
$ cat repro.py
import numpy as np
arr1 = np.array([1,2,3], dtype=np.float32)
arr2 = flex.double(10)
arr2[0] += arr1[0]

$ libtbx.python repro.py
  File "/dev/shm/dwpaley/20250610/modules/repro.py", line 5, in <module>
    arr2[0] += arr1[0]
    ~~~~^^^
TypeError: 'int' object is not iterable

```

dwpaley:
Workaround for https://github.com/boostorg/python/issues/511

NumPy 2.0 scalar types (np.float32, np.int32, np.int64, etc.) no longer subclass Python's built-in float/int, so Boost.Python's rvalue converters fail to convert them. Register custom from-Python converters that use the PyNumber_Float/PyNumber_Long protocols instead.

Fixes https://github.com/cctbx/cctbx_project/issues/1084

### DIFF
--- a/scitbx/array_family/boost_python/numpy_bridge.cpp
+++ b/scitbx/array_family/boost_python/numpy_bridge.cpp
@@ -1,6 +1,7 @@
 #include <stdint.h>
 #include <boost/python/numpy.hpp>
 #include <boost/python.hpp>
+#include <boost/python/converter/registry.hpp>
 #include <scitbx/array_family/versa.h>
 #include <scitbx/array_family/accessors/flex_grid.h>
 #include <boost_adaptbx/type_id_eq.h>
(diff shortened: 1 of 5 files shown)


bkpoon (commented):
## Integer overflow detection

The current `PyLong_AsLong` + `static_cast<CppType>` pattern silently truncates values that don't fit the target type (e.g. `np.int64(2147483648)` assigned to `flex.int` silently stores a wrong value). `PyLong_AsLong` does not set an error on overflow — it truncates to C `long` range.

**Fix:** Use `PyLong_AsLongLong` to get the full value, then explicitly range-check against `std::numeric_limits<CppType>` before casting. This raises `OverflowError` for out-of-range values instead of silently corrupting data.

Three suggested changes:
1. Add `#include <limits>` for `std::numeric_limits`
2. Replace the integer converter body with overflow-safe version
3. Add overflow boundary tests

scitbx/array_family/boost_python/numpy_bridge.cpp
```diff
@@ -1,6 +1,7 @@
 #include <stdint.h>
 #include <boost/python/numpy.hpp>
 #include <boost/python.hpp>
+#include <boost/python/converter/registry.hpp>
```
bkpoon, line 4: Add `#include <limits>` for `std::numeric_limits` used in the overflow check:

```suggestion
#include <boost/python/converter/registry.hpp>
#include <limits>
```

scitbx/array_family/boost_python/numpy_bridge.cpp
```diff
@@ -230,4 +231,105 @@ namespace scitbx { namespace af { namespace boost_python {
...
+          data)->storage.bytes;
+      PyObject* as_long = PyNumber_Long(obj);
+      if (!as_long) boost::python::throw_error_already_set();
+      CppType value = static_cast<CppType>(PyLong_AsLong(as_long));
+      if (value == static_cast<CppType>(-1) && PyErr_Occurred()) {
+        Py_DECREF(as_long);
+        boost::python::throw_error_already_set();
+      }
+      Py_DECREF(as_long);
+      new (storage) CppType(value);
```
bkpoon, line 301: `PyLong_AsLong` silently truncates values outside C `long` range, and even within `long` range, the `static_cast<CppType>` silently narrows (e.g. `long` value `2147483648` cast to `int` wraps to `-2147483648`). Neither sets a Python error, so the `-1 && PyErr_Occurred()` check only catches genuinely non-numeric inputs.

Using `PyLong_AsLongLong` + explicit range checking catches all overflow cases:

```suggestion
      long long wide_value = PyLong_AsLongLong(as_long);
      if (wide_value == -1 && PyErr_Occurred()) {
        Py_DECREF(as_long);
        boost::python::throw_error_already_set();
      }
      Py_DECREF(as_long);
      if (wide_value < static_cast<long long>(std::numeric_limits<CppType>::min()) ||
          wide_value > static_cast<long long>(std::numeric_limits<CppType>::max())) {
        PyErr_SetString(PyExc_OverflowError,
          "NumPy scalar value out of range for target C++ integer type");
        boost::python::throw_error_already_set();
      }
      new (storage) CppType(static_cast<CppType>(wide_value));
```

scitbx/array_family/boost_python/tst_numpy_scalar_conversions.py
```diff
@@ -0,0 +1,109 @@
...
+    return
+  from scitbx.array_family import flex
+
+  exercise_original_reproducer(np, flex)
+  exercise_element_setitem(np, flex)
+  exercise_element_iadd(np, flex)
+  exercise_array_iadd(np, flex)
+  exercise_construction_from_numpy_scalars(np, flex)
+  exercise_value_preservation(np, flex)
+  print("OK")
```
bkpoon, line 20: Call the new overflow test:

```suggestion
  exercise_value_preservation(np, flex)
  exercise_integer_overflow(np, flex)
  print("OK")
```

scitbx/array_family/boost_python/tst_numpy_scalar_conversions.py
```diff
@@ -0,0 +1,109 @@
...
+  a[0] = np.float64(1.234567890123456)
+  assert a[0] == 1.234567890123456
+
+  # Large integers
+  b = flex.int(1)
+  b[0] = np.int32(2147483647)  # INT32_MAX
+  assert b[0] == 2147483647
+
+if __name__ == "__main__":
+  run(args=sys.argv[1:])
```
bkpoon, line 109: Add overflow boundary tests to verify out-of-range numpy scalars raise errors rather than silently truncating:

```suggestion
def exercise_integer_overflow(np, flex):
  """Test that out-of-range numpy scalars raise errors, not silently truncate."""
  c = flex.int(1)
  # INT32_MAX should work
  c[0] = np.int64(2147483647)
  assert c[0] == 2147483647
  # INT32_MIN should work
  c[0] = np.int64(-2147483648)
  assert c[0] == -2147483648
  # INT32_MAX + 1 should raise an error
  raised = False
  try:
    c[0] = np.int64(2147483648)
  except Exception:
    raised = True
  assert raised, "Expected error for INT32_MAX + 1, got silent truncation"
  # Large uint64 should raise an error
  raised = False
  try:
    c[0] = np.uint64(4294967295)
  except Exception:
    raised = True
  assert raised, "Expected error for large uint64, got silent truncation"

if __name__ == "__main__":
  run(args=sys.argv[1:])
```

bkpoon:
@dwpaley Can this be merged?

There are some edge cases with `np.int64` and integer truncation, but that can resolved later.

PR #1146, opened 2026-04-10
Niggli reduction c++

dwmoreau:
This PR replaces the existing Niggli reduction with a c++ implementation. It was copied from the Gemmi implementation and edited to match differences in how cctbx counted iterations. The testing was edited to cross-check the new c++ implementation with the existing python implementation. 

During serial data reduction, when multiple lattices are present on an image, indexing is performed by repetitive indexing attempts with subsampled observations. Profiling showed the existing Niggli Reduction contributed a significant amount of execution time. The c++ reduces the fraction of time in the Niggli reduction from 27.4%  to 4% in a test case.

### DIFF
--- a/cctbx/regression/tst_krivy_gruber.py
+++ b/cctbx/regression/tst_krivy_gruber.py
@@ -117,7 +117,8 @@ def do_reduce(inp):
   if (relative_epsilon is None):
     assert check_is_niggli_cell(red_cell).itva_is_niggli_cell()
   # Cross-check: verify C++ niggli_reduction gives identical results to Python.
-  cpp_red = inp.niggli_reduction(relative_epsilon=relative_epsilon or 1.e-5)
+  cpp_red = inp.niggli_reduction(
+    relative_epsilon=1.e-5 if relative_epsilon is None else relative_epsilon)
   assert cpp_red.as_unit_cell().is_similar_to(red_cell), \
     "C++ niggli cell differs from Python: %s vs %s" % (
       cpp_red.as_unit_cell(), red_cell)
--- a/cctbx/regression/tst_krivy_gruber.py
+++ b/cctbx/regression/tst_krivy_gruber.py
@@ -116,6 +116,15 @@ def do_reduce(inp):
   assert inp.change_basis(red.r_inv().elems).is_similar_to(red_cell)
   if (relative_epsilon is None):
     assert check_is_niggli_cell(red_cell).itva_is_niggli_cell()
+  # Cross-check: verify C++ niggli_reduction gives identical results to Python.
+  cpp_red = inp.niggli_reduction(relative_epsilon=relative_epsilon or 1.e-5)
+  assert cpp_red.as_unit_cell().is_similar_to(red_cell), \
+    "C++ niggli cell differs from Python: %s vs %s" % (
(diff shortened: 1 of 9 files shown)


bkpoon (commented):
## Review Summary

### What it does

Replaces the pure-Python Krivý–Gruber (1976) Niggli reduction with a header-only C++ implementation (`niggli_reduction.h`) exposed via Boost.Python. The Python `unit_cell.niggli_reduction()` now calls C++ and wraps the result in `cpp_niggli_reduction_result`. Claimed speedup: 27.4% → 4% of serial indexing time. Ported from gemmi's `GruberVector`.

### Algorithm correctness

**Verified correct.** All actions (A1–A8), conditions, G6 arithmetic, column operations, sign normalization (including the zero-component edge case in `do_a3_a4`), and iteration counting match the Python reference implementation exactly. API compatibility is preserved — no callers will break.

### Items to address before merge

| Priority | Issue | Location |
|----------|-------|----------|
| **High** | **Licensing.** gemmi is MPL-2.0, cctbx is BSD-3. If code was copied verbatim, the file may need an MPL-2.0 header. Author should confirm clean-room status. | `niggli_reduction.h` |
| **High** | **Over-broad `except RuntimeError`.** Any C++ `RuntimeError` gets relabeled as `iteration_limit_exceeded`, masking unrelated errors. Narrow the catch (check message, or register a dedicated Python exception type). | `__init__.py:204-205` |
| **Medium** | **Epsilon inconsistency.** C++ computes `epsilon` from the input cell volume; the Python wrapper computes it from the reduced cell. Expose `epsilon_` from C++ and reuse it, or compute Python-side from the input cell. | `reduction_base.py:136` |
| **Medium** | **Missing test for C++ exception path.** No test exercises `unit_cell.niggli_reduction(iteration_limit=1)` — only the old Python path is tested. | `tst_krivy_gruber.py` |

### Nice-to-haves

| Issue | Location |
|-------|----------|
| Misleading comment says "all-positive branch" but the edge case only fires in the all-negative branch | `niggli_reduction.h:218` |
| `super().__init__()` bypass should have an inline comment (not just the class docstring) explaining the deliberate skip | `reduction_base.py:130` |
| Magic defaults `1.e-5` and `1000` appear in three places (C++ default args, Boost binding, Python wrapper) — consider consolidating | `niggli_reduction.h`, `niggli_reduction.cpp`, `__init__.py` |
| `raise ... from e` for proper exception chaining | `__init__.py:205` |
| Targeted unit test for the `do_a3_a4` zero-component edge case would help future readers | `tst_krivy_gruber.py` |

### Strengths

- Robust cross-check testing — exact `r_inv` integer equality across the full Gruber-type test battery
- G6 parameters returned directly from C++, avoiding metrical-matrix round-trip drift
- `r_inv` stored as `mat3<int>`, keeping determinants and equality checks exact
- Iteration semantics carefully matched to the Python reference
- Good inline documentation of the algorithm

### Verdict

Solid port with meaningful speedup and a strong test strategy. Resolve the licensing confirmation, narrow the exception catch, fix the epsilon source, and add the exception-path test before merge. The rest are optional improvements.

bkpoon:
For testing, do we want to remove the `--Quick` flag from `cctbx/run_tests.py`? The test runs pretty quickly anyway.

PR #1147, opened 2026-04-10
libtbx: fix wrong reading mode when unpickling temporary file

sebth:
When trying to submit a Slurm job in Phenix 2.0, the job submission crashes with
```
  File "<stdin>", line 2, in <module>
  File "/software/presto/e/sebth/software/Phenix/2.0-5936-foss-2023a/phenix-2.0-5936/build/../conda_base/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```
due to libtbx attempting to read a pickle in text mode instead of binary mode. Applying the commit in this PR to libtbx in our local Phenix 2.0 installation fixes the job submission.

### DIFF
--- a/libtbx/queuing_system_utils/processing/transfer.py
+++ b/libtbx/queuing_system_utils/processing/transfer.py
@@ -7,7 +7,7 @@ class TemporaryFile(object):
   Sends data by writing out a temporary file
   """
 
-  SCRIPT= "( target, args, kwargs ) = pickle.load( open( \"%s\" ) )"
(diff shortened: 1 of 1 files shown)


bkpoon:
Thanks! We're in the process of setting up a SLURM system so we try to test this more thoroughly.

We are in the release candidate process for Phenix 2.1, so the next release candidate (this weekend or early next week) will have this fix.

PR #1149, opened 2026-04-15
xcif parser now wired to the iotbx.cif.reader 

olegsobolev:
xcif — a fast C++ CIF parser — is now wired into `iotbx.cif.reader` with a single global switch in `iotbx/cif/__init__.py` (the module-level `DEFAULT_ENGINE` constant) plus a per-call `engine=` kwarg on the reader. `DEFAULT_ENGINE` stays `"ucif"` in this PR; callers who want to opt in today can pass `engine="xcif"` explicitly, and a one-line follow-up will flip the default after bake-in.

Whole set of tests including Phenix ones are passing with the switch flipped.

Benchmark on a 34 MB mmCIF (`4v9d.cif`): ~6× speedup end-to-end through `iotbx.cif.reader`, ~21× for the raw parser alone.

### DIFF
--- a/.github/workflows/quick.yml
+++ b/.github/workflows/quick.yml
@@ -127,6 +127,13 @@ jobs:
           ls ./modules
           df -h
 
+      - name: Skip build cache, if requested
+        if: contains(github.event.head_commit.message, '[skip cache]')
+        run: |
+          set -xe
+          rm -fr ./build
+          ls
+
       - name: Run bootstrap.py
         run: |
           set -xe
@@ -273,6 +280,13 @@ jobs:
         shell: cmd
         run: |
           del /S /Q .\build\lib\_pycbf.pyd 2> nul
+
+      - name: Skip build cache, if requested
+        if: contains(github.event.head_commit.message, '[skip cache]')
+        shell: cmd
+        run: |
+          rmdir /s /q .\build
+          dir
 
       - name: Run bootstrap.py
         shell: cmd
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,7 @@ mmtbx/suitename/assist/
 mmtbx/suitename/*.show.txt
 
 *sublime*
+
+# Local Claude AI guidance files (not for shared use)
+CLAUDE.md
+*/CLAUDE.md
--- a/iotbx/cif/__init__.py
+++ b/iotbx/cif/__init__.py
@@ -26,6 +26,7 @@ from libtbx.utils import flat_list
 from libtbx.utils import detect_binary_file
 from libtbx import smart_open
 
+import os
 import sys
 
 distances_as_cif_loop = geometry.distances_as_cif_loop
@@ -42,21 +43,22 @@ DEFAULT_ENGINE = "ucif"
 _VALID_ENGINES = ("ucif", "xcif")
 
 
-def _drive_builder_from_xcif(builder, input_string, strict):
-  """Parse input_string with xcif and drive the given cif_model_builder-style
-  builder via its callback methods (add_data_block / add_data_item / add_loop /
-  start_save_frame / end_save_frame).
+_XCIF_COMPRESSED_SUFFIXES = (".gz", ".Z", ".bz2")
 
-  Returns a (possibly empty) list of error message strings. xcif throws on
-  first parse error rather than accumulating, so the list has at most one
-  entry. When it does, the builder is left with whatever state it had
-  before the throw — which is nothing, since xcif_ext.parse runs to
-  completion before the walker touches the builder.
 
-  strict=False relaxes two things: pair items appearing before the first
-  data_ block are attached to an implicit 'global_' block; and explicit
-  'global_' block headers (STAR/DDL2 convention, used by the cctbx
-  monomer library) are accepted. strict=True rejects both.
+def _xcif_can_use_parse_file(file_path):
+  """True when file_path is safe to hand to xcif_ext.parse_file (which
+  memory-maps). False for compressed files; smart_open handles those,
+  parse_file does not."""
+  if file_path is None:
+    return False
+  return not file_path.endswith(_XCIF_COMPRESSED_SUFFIXES)
+
+
+def _walk_xcif_doc(builder, doc):
+  """Drive `builder` (a cif_model_builder-style object with
+  add_data_block / add_data_item / add_loop / start_save_frame /
+  end_save_frame) from an already-parsed xcif Document.
 
   The xcif C++ parser stores block and save-frame names with the data_/save_
   prefix stripped; the builder's strip-before-first-underscore logic
@@ -64,15 +66,6 @@ def _drive_builder_from_xcif(builder, input_string, strict):
   tokens, so we prepend the prefix back.
   """
   import xcif_ext
-  try:
-    doc = xcif_ext.parse(input_string, strict=strict)
-  except (ValueError, RuntimeError) as e:
-    # xcif raises ValueError for std::invalid_argument (explicit translator)
-    # and RuntimeError for CifError (inherits std::runtime_error, default
-    # boost.python translation). Return the message so the caller can
-    # either surface it as CifParserError or stash it for error_count().
-    return [str(e)]
-
   # Use the enum values from xcif_ext directly — no ABI drift.
   # int(...) because bp::enum_ wraps values in a Python type;
   # the tuple comparisons below want plain ints.
@@ -119,6 +112,45 @@ def _drive_builder_from_xcif(builder, input_string, strict):
         builder.end_save_frame()
       else:
         _emit_pair_or_loop(kind, idx, pair_tags, pair_values, loops)
+
+
+def _drive_builder_from_xcif(builder, input_string, strict):
+  """Parse input_string with xcif and drive the given builder.
+
+  Returns a (possibly empty) list of error message strings. xcif throws on
+  first parse error rather than accumulating, so the list has at most one
+  entry. When it does, the builder is left with whatever state it had
+  before the throw — which is nothing, since xcif_ext.parse runs to
+  completion before the walker touches the builder.
+
+  strict=False relaxes two things: pair items appearing before the first
+  data_ block are attached to an implicit 'global_' block; and explicit
+  'global_' block headers (STAR/DDL2 convention, used by the cctbx
+  monomer library) are accepted. strict=True rejects both.
+  """
+  import xcif_ext
+  try:
+    doc = xcif_ext.parse(input_string, strict=strict)
+  except (ValueError, RuntimeError) as e:
+    # xcif raises ValueError for std::invalid_argument (explicit translator)
+    # and RuntimeError for CifError (inherits std::runtime_error, default
+    # boost.python translation). Return the message so the caller can
+    # either surface it as CifParserError or stash it for error_count().
+    return [str(e)]
+  _walk_xcif_doc(builder, doc)
+  return []
+
+
+def _drive_builder_from_xcif_file(builder, file_path, strict):
+  """Like _drive_builder_from_xcif but dispatches to xcif_ext.parse_file
+  (memory-mapped, zero-copy). Avoids allocating a Python string copy of
+  the file contents when the caller supplied a plain uncompressed path."""
+  import xcif_ext
+  try:
+    doc = xcif_ext.parse_file(file_path, strict=strict)
+  except (ValueError, RuntimeError) as e:
+    return [str(e)]
+  _walk_xcif_doc(builder, doc)
   return []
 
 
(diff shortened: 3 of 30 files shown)


bkpoon (commented):
## Code Review

**Verdict: Ready to merge as opt-in — a few items should be resolved before the follow-up PR that flips `DEFAULT_ENGINE` to `"xcif"`.**

Because `DEFAULT_ENGINE = "ucif"` is preserved, merging this PR today ships zero behavior change to the ~120 existing `iotbx.cif.reader(...)` call sites. The items below are mostly scoped at the default-flip, not this merge.

### Strengths

- Default preserved (`iotbx/cif/__init__.py:41`); opt-in via `engine="xcif"` is the right shipping path for a dual-parser change.
- Walker docstring on `_drive_builder_from_xcif` (`iotbx/cif/__init__.py:48-81`) is load-bearing — explains strict/non-strict semantics, the `data_`/`save_` prefix round-trip to match the builder's `[find('_')+1:]` logic, and the ordering limitation.
- Real parity test at `xcif/regression/tst_iotbx_cif_reader_parity.py` drives `engine="xcif"` against ucif on 7 inline fixtures plus file fixtures, comparing block names, pair items, loop tag sets and column content, and save-frame structure.
- Non-trivial correctness fixes bundled in:
  - CIF 1.1-compliant embedded-delimiter handling (`xcif/tokenizer.cpp:118-142`) — fixes the `N-METHYL-PYRIDOXAL-5'-PHOSPHATE` monomer-library case.
  - Unterminated-quote / semicolon now throws (previously returned partial values silently).
  - `global_` recognition for mon_lib files; DOS `0x1A` EOF compat for SHELX output.
  - Locale fix via `strtod_l` / `_strtod_l` behind a C-locale handle (`xcif/numeric.cpp:12-27`), thread-safe static init.
  - Integer parsing via `strtol` with explicit overflow check (`xcif/numeric.cpp:63-80`); old hand-rolled loop silently overflowed at `INT_MAX`. Paired with `INT_MAX`/`INT_MIN` tests in `tst_numeric.cpp`.
  - `string_view` bounds now assertable in debug (`xcif/include/xcif/string_view.h`).
- GIL released during `parse`/`parse_file` in `xcif/xcif_ext.cpp:83-100`; Python walker runs after heavy C++ work with GIL reacquired.

### Issues

#### Critical
None.

#### Important (should be resolved before flipping the default)

1. **Builder contract fragility with save frames.** `iotbx/cif/__init__.py:102-103` synthesises the save-frame heading as `"save_" + sf.name`. This round-trips only because `iotbx/cif/builders.py:68-72` strips it back with `save_frame_heading[save_frame_heading.find('_')+1:]`. Third-party `builder` objects passed via `reader(builder=...)` that string-compare the full `save_<name>` token will silently diverge between engines. At minimum, document the builder contract on the `builder=` parameter.

2. **Pair/loop source order not preserved by xcif.** The walker explicitly emits all pairs before all loops within a block (`iotbx/cif/__init__.py:87-94`) because `xcif::Block` stores `pairs_` and `loops_` in separate vectors. ucif preserves source order. The parity test's `_compare_block_like` (`xcif/regression/tst_iotbx_cif_reader_parity.py:159-178`) compares via `dict(blk_u._items)`, so order is discarded and CI will not catch this. `block.show()` output will change once the default flips, which makes file diffs noisy and can break snapshot tests downstream.
   - Fix (minimum): tighten the parity test to assert list-order equivalence. Then either store original order in xcif and honour it in the walker, or flag this as a known behavioral difference in release notes.

3. **`raise_if_errors=False` semantics diverge.** On parse failure, `_drive_builder_from_xcif` returns before the walker touches the builder at all (`iotbx/cif/__init__.py:67-74`) — xcif yields an empty `self.model()`. ucif accumulates errors while partially populating the model. Callers that only check `error_count() == 0` are unaffected, but callers trying to salvage a partial model from a damaged file behave radically differently. Should at least be documented in the `engine=` docstring.

4. **No docstring on `reader.engine`.** The `reader` class has no class docstring at all; `DEFAULT_ENGINE` has a one-line comment but doesn't document runtime-override semantics. Nothing in `sphinx/iotbx/iotbx.cif.rst` is updated either. For a switch that affects ~120 call sites, this is inadequate.

5. **Zero-copy `parse_file` path is not used.** When `file_path=` is given, reader reads the file into a Python string (`iotbx/cif/__init__.py:127-130`) then calls `xcif_ext.parse`. xcif has a `parse_file` entry point that memory-maps and parses directly. This is plausibly why the headline speedup is 6× through `iotbx.cif.reader` but 21× for the raw parser on the same 34 MB file. Suggest dispatching to `parse_file` when `file_path` is given and engine is `xcif`; binary-detection only reads the first few KB so it remains cheap.

6. **Engine validation happens after I/O.** `iotbx/cif/__init__.py:134-140` validates `engine` only after the file has been read and binary-detected. A typo'd engine forces needless I/O before `ValueError`. Easy to move the check to the top of `__init__`.

7. **`DEFAULT_ENGINE` thread-safety undocumented.** Reader reads `DEFAULT_ENGINE` at construction time (line 137), so a concurrent test mutating it during another test's parse races. Unlikely given Phenix's spawn/fork model but worth noting in the module docstring. Alternatively, freeze it to `"ucif"` and expose only the per-call kwarg until the flip — making the flip literally a one-line commit as described.

#### Minor

8. `_VALID_ENGINES = ("ucif", "xcif")`: fine for 2 items; a `dict` mapping engine → walker factory scales better when a third option lands.
9. Walker catches `(ValueError, RuntimeError)` at `iotbx/cif/__init__.py:68`. If a `std::exception` subclass ever escapes without a custom translator, it would bypass `_xcif_errors` and the `raise_if_errors=False` contract. Defensively broadening the except with a class filter is cheap insurance.
10. xcif error messages prefixed `"<source>:line:col: "` vs ucif's plain messages. Test updates already accommodate both wordings (`iotbx/cif/tests/tst_model_builder.py:34-52`, `iotbx/pdb/tst_pdb.py:718-728`) — worth calling out in release notes for anyone grepping logs for error patterns.
11. `.gitignore` adds `CLAUDE.md` / `*/CLAUDE.md`. Tooling-specific patterns like this are better in `~/.gitignore_global` or `.git/info/exclude` rather than the shared project `.gitignore`. Trivial nit.
12. Two-dot diff shows `mmtbx/ligands/rdkit_utils.py` and `mmtbx/validation/validate_ligands.py` as deletions; this is a rebase artifact (branch is behind master by `3077f4bc11` and `f37c1b9f65`). Not a real issue, but rebasing onto master would clean the diff view.
13. `tags = list(xloop.tags)` at walker — minor allocation, not worth fixing.
14. `xcif/regression/example.cif` was modified to add `_chem_comp` / `_entity` / `_atom_type` stubs with matching `tst_example_tokenize.py` golden-count updates. Verify no other cctbx modules reference this fixture.

### Recommendations

- **Keep the merge two-step.** Ship this PR as opt-in, then flip the default in a separate one-line PR with a bake-in window. Do not squash — makes any regression cleanly attributable to the flip and trivially revertible.
- **Add a CI matrix job that runs the cctbx test suite under `DEFAULT_ENGINE="xcif"`.** The "whole test suite passes" claim is currently author-local and will decay with each new PR that lands. An env-var-honoring path in `iotbx/cif/__init__.py` is enough.
- **Strengthen the parity test to fail on source-order mismatches** (see item 2).
- **Document semantic differences in the `engine=` docstring**: partial-model-on-error, error-wording prefixes, pair/loop source order, file-read path performance characteristics.
- **Consider extracting the walker** to `iotbx/cif/_xcif_walker.py`. It's ~40 lines of non-trivial Python glue in a module that already carries a lot.

### Assessment

Merging this PR today is low-risk because the default path is untouched. Items 1–4 and the CI-matrix recommendation are the blocking work for the subsequent default-flip PR; 5 is a sizeable perf win left on the table; 6–7 are small-effort hardening.

---

*Review assisted by Claude Code.*

olegsobolev:
Thanks for the thorough review — I've addressed all items 1–7 plus a few of the minor ones. Summary, with commit references:

- **Item 2 (source order)** — Fixed in bf2e49ee. Block now records an ordered (kind, index) log; the iotbx.cif walker iterates it. Parity test tightened to compare `block._set` as a list, so a regression is caught in CI.
- **Item 5 (zero-copy parse_file)** — Fixed in 2a8df9da. `engine="xcif"` with a plain uncompressed `file_path` now dispatches to `xcif_ext.parse_file` (mmap). Compressed files (.gz/.Z/.bz2) and `file_object=` callers keep the read-into-string path. Binary detection runs via `detect_binary_file.from_initial_block` on the first 1000 bytes.
- **Item 6 (engine validation)** — Fixed in b1528af0. Validation moved to the top of `reader.__init__`; a typo'd engine raises `ValueError` before any file I/O.
- **Items 1, 4, 7, 10 (docstrings & semantics)** — Done in fbe030ed. Class docstring on `reader` covers the ucif/xcif divergences (error-message prefix, partial-model-on-error, source order, fast-path). The `builder=` doc pins the `data_`/`save_` prefix contract. Module comment clarifies `DEFAULT_ENGINE` is read at construction time.
- **Item 12 (rebase artifact)** — Rebased onto current master; the mmtbx deletion noise is gone.

On **item 3** (partial-model-on-error) — I went with "document loudly" rather than unify. The unification route needs a parallel non-throwing C++ API (`parse_lenient`/`parse_file_lenient` plus `Document::parse_errors_`) and would be ~100 lines of new public surface. No current cctbx caller uses `raise_if_errors=False` to salvage partial state — failures are either raised or error-count-checked and surfaced, not mined. The divergence is now called out explicitly in the engine= docstring so anyone hitting it knows why. Happy to add the lenient variants in a follow-up if you disagree.

**Deferred** (happy to take them in follow-ups or here if you prefer):
- Items 8, 9, 11, 14 — all minor.
- **CI matrix job** running the suite under `DEFAULT_ENGINE="xcif"`. Best done as its own PR since it touches `.github/` infra rather than the library.

Ready for another pass.
Edit: Going to merge as it is now.

bkpoon (commented):
## Follow-up Review (delta)

Verified the four new commits against the Important items from my previous review. Scope: only what changed since `f5099ea`.

### Item-by-item verification

- **Item 1 — builder contract re: `data_`/`save_` prefix** (`fbe030ed`): ✅ Docstring under `builder=` now explicitly states that headings are passed with the prefix and that the default strips-before-first-underscore.
- **Item 2 — pair/loop source order** (`bf2e49ee`): ✅ `xcif::Block::source_order_` records insertion order; walker iterates it. The parity test fixture now interleaves pair-loop-pair-loop-pair and `_compare_block_like` asserts `list(blk_u._set) == list(blk_x._set)` before value comparison — the pre-fix bug would fail the new test.
- **Item 3 — `raise_if_errors=False` semantics** (doc-only decision): ✅ with one caveat worth noting. The docstring explicitly directs salvage-semantics callers to `engine="ucif"`, which makes the decision defensible. One in-tree caller actually does mine post-parse state: `mmtbx/chemical_components/cif_parser.py:259` iterates `cm.items()` after `cif.reader(filename, strict=False, raise_if_errors=False).model()`. Unaffected as long as `DEFAULT_ENGINE="ucif"`; worth remembering when the default flips so that this caller is either (a) migrated to a lenient-xcif variant, or (b) pinned to `engine="ucif"`. The only two other `raise_if_errors=False` callers in the repo are a unit test and `cctbx/web/iotbx_cif_validate.py:28`, so the exposure is narrow.
- **Item 4 — `reader` docstring** (`fbe030ed`): ✅ Class now carries a full docstring covering all kwargs + engine divergences.
- **Item 5 — zero-copy `parse_file`** (`2a8df9da`): ✅ Fast path gated on `engine=="xcif"` AND `file_object is None` AND uncompressed plain path, with binary detection preserved. Walker factored into shared `_walk_xcif_doc` so both `parse` and `parse_file` paths funnel through identical source-order / error-translation logic. Tests monkey-patch `xcif_ext.parse` / `xcif_ext.parse_file` to verify dispatch.
- **Item 6 — validate `engine=` before I/O** (`b1528af0`): ✅ Check moved to right after the file-source assertion. New regression (`tst_iotbx_cif_reader_engine.py::exercise_engine_validated_before_io`) uses a guaranteed-nonexistent path to prove the check short-circuits before `smart_open`.
- **Item 7 — `DEFAULT_ENGINE` thread-safety**: ⚠️ Read-at-construction is now documented in the module comment; an explicit "mutating concurrently is a data race" caveat is not. Low risk in practice (GIL + Phenix's fork-based multiprocessing); acceptable as is.

### Minor observations (non-blocking)

- Fast-path error messages for missing files now come from xcif rather than `smart_open`'s `Sorry("Cannot find...")`. Callers grepping logs for the `Sorry` wording would see a different string under `engine="xcif"`. Not covered by the docstring's error-prefix note.
- Per-block list allocations in the walker (`pair_tags`, `pair_values`, `loops`, `save_frames`) — O(items-per-block), dwarfed by the `parse_file` win.

### Verdict

**Ready to merge.** All Important items are addressed in substance; item 7 is partially addressed but low-risk for an opt-in engine. The new tests would catch regressions of items 2, 5, and 6. No new bugs introduced by the four follow-up commits.

The `mmtbx/chemical_components/cif_parser.py` caller noted above and a CI matrix job running the suite under `DEFAULT_ENGINE="xcif"` are the natural follow-up items alongside flipping the default.

---

*Delta review assisted by Claude Code.*

bkpoon:
LGTM. One other comment is that when the switch is made to `xcif`, does the `run` function in `mmtbx/chemical_components.py` need any updates?

olegsobolev:
Nice catch. mmtbx/chemical_components/cif_parser.py will be updated to with raise_if_errors=True along with general cleanup after merging this PR.

PR #1152, opened 2026-04-21
Niggli reduction c++ port

dwmoreau:
This PR replaces the existing Niggli reduction with a c++ implementation. This PR ports the cctbx Niggli from python to c++ and was done with Claude Code. A new comphrensive test cross-check the public methods of the new c++ implementation with the existing python implementation.

During serial data reduction, when multiple lattices are present on an image, indexing is performed by repetitive indexing attempts with subsampled observations. Profiling showed the existing Niggli Reduction contributed a significant amount of execution time. This PR reduces the fraction of time in the Niggli reduction from 27.4% to 4% in one test case and from 34% to 3.4% in another.

### DIFF
--- a/cctbx/uctbx/__init__.py
+++ b/cctbx/uctbx/__init__.py
@@ -191,8 +191,11 @@ class _():
     return gruber_parameterization(self, relative_epsilon).is_niggli_cell()
 
   def niggli_reduction(self, relative_epsilon=None, iteration_limit=None):
-    from cctbx.uctbx import krivy_gruber_1976
-    return krivy_gruber_1976.reduction(self, relative_epsilon, iteration_limit)
+    if relative_epsilon is None:
+      relative_epsilon = 1.e-5
+    if iteration_limit is None:
+      iteration_limit = 1000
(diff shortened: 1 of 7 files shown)


cctbx/uctbx/__init__.py
```diff
@@ -297,6 +300,16 @@ def debye_waller_factors(self,
...
+# r_inv() is exposed directly from C++ and returns scitbx.matrix.sqr.
+# change_of_basis_op() uses a deferred sgtbx import because cctbx_sgtbx_ext
+# is not guaranteed to be loaded by a plain "from cctbx import uctbx".
+@bp.inject_into(ext.niggli_reduction)
+class _():
+  def change_of_basis_op(self):
+    from cctbx import sgtbx
+    return sgtbx.change_of_basis_op(
+      sgtbx.rt_mx(sgtbx.rot_mx(self.r_inv().elems, 1))).inverse()
+
```
phyy-nx, line 303: Some grepping indicates all these functions are in c++.  Do we need boost injection here?

dwmoreau: The next commit moves this method into c++.

PR #1153, opened 2026-04-22
Add bootstrap filter method to cctbx.xfel.merge

phyy-nx:
See Grünbein 2021, this adds subsampling with replacement of the lattices to enable creating replicates that can be individually used with structure refinement.  The average atomic positions provide an estimate of coordinate error.


phyy-nx:
Matching test is here: https://gitlab.com/cctbx/xfel_regression/-/merge_requests/23

PR #1159, opened 2026-05-11
start of ION select (and other new selections)

nwmoriarty:


olegsobolev:
looks good to me if tests pass. Also suggest adding a test case to cctbx_project/iotbx/pdb/tst_atom_selection.py

nwmoriarty:
Did have a test but didn't choice the correct commit. All should be well.

Cheers

Nigel

---
Nigel W. Moriarty
Building 92-4-431M1, Molecular Biophysics and Integrated Bioimaging
Berkeley, CA 94720-8235
Email : ***@***.***
Web  : CCI.LBL.gov
ORCID : orcid.org/0000-0001-8857-9464

PR #1162, opened 2026-05-14
Added the 'UHF' keyword for non-singlet calculations.

krlun:
MOPAC defaults to RHF calculations, i.e. singlet state. If the multiplicitly is changed from singlet, this also requires the UHF keyword to be present.


nwmoriarty:
Works well

PR #1167, opened 2026-06-02
Fix probe2 crash on atoms near crystallographic symmetry mates

vbchen:
## Summary
Fix a probe2 crash on atoms near crystallographic symmetry mates.

When an atom sits near a crystallographic symmetry mate (e.g. GLN NE2 ~2.06 Å from its symmetry copy in 3q9v), the restraints engine expresses some covalent bonds as **ASU bond proxies** rather than simple bond proxies. `Helpers.getBondedNeighborLists()` only processed the simple proxies, so hydrogens such as HE21 were left with no bonded neighbors and probe2 crashed with `Found Hydrogen with no neighbors`.

## Change
- `mmtbx/probe/Helpers.py`: `getBondedNeighborLists()` gains an optional `asuBondProxies=None` argument. When provided, ASU proxies with `j_sym == 0` (same asymmetric unit) are added to the bonded-neighbor lists. Behavior is unchanged when no ASU proxies are passed (purely additive).
- Callers in `mmtbx/programs/probe2.py`, `mmtbx/programs/reduce2.py`, `mmtbx/reduce/Optimizers.py`, and the `Helpers` self-tests now capture the second return value of `get_all_bond_proxies()` (the ASU proxies) and pass it through, instead of discarding it with `[0]`.

## Test
Adds `mmtbx/validation/regression/tst_probe2_asu_bond_proxies.py`, which exercises the ASU-proxy path. Passes locally.

### DIFF
--- a/mmtbx/probe/Helpers.py
+++ b/mmtbx/probe/Helpers.py
@@ -161,7 +161,7 @@ def createDotScorer(extraAtomInfo, probePhil):
 ##################################################################################
 # Other helper functions.
 
-def getBondedNeighborLists(atoms, bondProxies):
+def getBondedNeighborLists(atoms, bondProxies, asuBondProxies=None):
   """
     Helper function to produce a dictionary of lists that contain all bonded
     neighbors for each atom in a set of atoms.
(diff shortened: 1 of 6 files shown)


russell-taylor (approved):
Thank you for adding this improvement!

PR #1176, opened 2026-06-10
Fix reduce2 to emit PDB v3 atom names for RNA/DNA hydrogens

vbchen:
## Summary
Fix `reduce2` so the hydrogens it places on nucleic acids carry **PDB v3 atom names** (prime notation) instead of the old monomer-library convention.

The monomer library uses old-convention names (`H4*`, `H5*1`, `HO2*`, …) but PDB v3 requires prime notation (`H4'`, `H5'`, `HO2'`, …). This adds a translation step in `place_hydrogens` that converts monomer-library names to v3 reference names for nucleic-acid residues, using the existing mapping from `iotbx.pdb` plus a fallback `*`→`'` substitution for modified nucleotides.

## Test
- New `mmtbx/validation/regression/tst_reduce2_atom_names.py`: builds a model with all 8 standard NA residues (RNA A/G/C/U, DNA DA/DG/DC/DT), runs reduce2 H placement, and verifies every newly-added atom name uses the modern PDB v3 convention.
- Updated `tst_add_hydrogen_2.py` / `tst_add_hydrogen_5.py` to expect the new-style names.

All three pass locally.

## Notes
Single cohesive commit. Scope is limited to nucleic-acid hydrogen naming in the reduce2 path.


mmtbx/hydrogens/reduce_hydrogen.py
```diff
@@ -18,6 +18,28 @@
...
+# (e.g. H4*, C5*, O1P) but we want to emit new-convention names
+# (e.g. H4', C5', OP1) when placing hydrogens.
+_mon_lib_to_pdb_ref = {}
+for _ref, _ml in iotbx.pdb.rna_dna_atom_names_reference_to_mon_lib_translation_dict.items():
+  _mon_lib_to_pdb_ref[_ml] = _ref.strip()
+# Special cases not in the forward dict (see comment at line 177 of iotbx/pdb/__init__.py)
+_mon_lib_to_pdb_ref["H2*"] = "H2'"    # RNA
+_mon_lib_to_pdb_ref["H2*1"] = "H2'"   # DNA
+
+def _translate_na_atom_name(name):
```
russell-taylor, line 32: Would it be better to fix these at the source, rather than adjusting them after the fact? If that is not possible, how can we add a "tickler" reminder to remove these once the source has been fixed? Perhaps leaving an @todo comment?

russell-taylor (approved):

pafonine:
This is a question for Nigel and Dorothee (Dorothee is away on vacation 
right now). My impression was that we use GeoStd for most things:

https://journals.iucr.org/d/issues/2026/03/00/von5004/index.html

And yes, fixing at the source (as opposed to band-aiding later 
elsewhere) sounds like the best approach.

Pavel

nwmoriarty:
I'm in the middle of correcting a few things NA related. I'm working with Ezra and Helen Berman to improve the ideal and ESD in the restraints so PDB validation is more informative. I'm making the call to update all Phenix code to only use V3 names. Should take a fortnight or so.

PR #1178, opened 2026-06-15
Speed up clashscore2: make redundant 2nd probe2 run optional (~2x)

vbchen:
## Summary
Speed up `clashscore2` ~2× by making the redundant second `probe2` run optional, plus two small efficiency cleanups. Output on the default path is unchanged.

1. **Gate the second probe2 run behind a new `save_probe_output` phil (default `False`).** `probe_clashscore_manager` previously *always* ran probe2 a second time to produce full non-condensed output with VDW contacts ("for Coot"), stored it in `self.probe_json`, and never consumed it. That run is now skipped unless `save_probe_output=True`, which roughly halves runtime. The flag is threaded `programs/clashscore2.py` → `clashscore2.__init__` → `probe_clashscore_manager`.
2. **Classify water by residue group, single pass.** The non-water atom count now iterates `residue_groups() → atom_groups() → atoms()`, skipping whole water groups once instead of calling `common_residue_names_get_class` per atom; the `natoms_b_cutoff` tally is folded into the same pass (was a second full atom loop).
3. **`used = set()` for clash dedup** (was a list), and removed a dead `id_str_no_atom_name()` assignment that was immediately overwritten.

## Correctness
Verified `mmtbx.clashscore2` on a hydrogenated model (1nxb) before/after: **identical clashscore (293.67) and a byte-identical bad-clashes list** with the default `save_probe_output=False`. Wall time dropped from ~7.8s to ~5.5s.

Style gates pass locally (`libtbx.find_clutter`, `py_compile`). One now-unused `import re` was removed — the old crude clutter check had been treating the substring in a `we're` comment as a use; rewriting that comment exposed the genuinely dead import.

## Notes
Touches `mmtbx/programs/clashscore2.py` and `mmtbx/validation/clashscore2.py`.

### DIFF
--- a/mmtbx/validation/clashscore2.py
+++ b/mmtbx/validation/clashscore2.py
@@ -14,7 +14,6 @@ import iotbx.pdb
 import iotbx.cli_parser
 import mmtbx
 import os
-import re
 import sys
 import six
 import json
@@ -63,6 +62,7 @@ class clashscore2(validation):
       b_factor_cutoff=None,
       verbose=False,
       do_flips=False,
+      save_probe_output=False,
       out=sys.stdout):
     validation.__init__(self)
     self.b_factor_cutoff = b_factor_cutoff
@@ -146,7 +146,8 @@ class clashscore2(validation):
         b_factor_cutoff=b_factor_cutoff,
         use_segids=use_segids,
         verbose=verbose,
-        model_id=model.id)
+        model_id=model.id,
+        save_probe_output=save_probe_output)
       self.probe_clashscore_manager.run_probe_clashscore(data_manager, subset_model_manager)
 
       self.clash_dict[model.id] = self.probe_clashscore_manager.clashscore
@@ -329,7 +330,8 @@ class probe_clashscore_manager(object):
                b_factor_cutoff=None,
                use_segids=False,
                verbose=False,
-               model_id=""):
+               model_id="",
+               save_probe_output=False):
     """
     Calculate probe (MolProbity) clashscore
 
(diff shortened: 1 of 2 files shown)


mmtbx/validation/clashscore2.py
```diff
@@ -553,28 +569,21 @@ def run_probe_clashscore(self, data_manager, hydrogenated_model):
...
 
       for clash_obj in sorted(temp):
-        test_key = clash_obj.id_str_no_atom_name()
         test_key = clash_obj.id_str()
         if test_key not in used:
-          used.append(test_key)
+          used.add(test_key)
           self.bad_clashes.append(clash_obj)
 
-      if self.b_factor_cutoff is not None:
```
russell-taylor, line 565: Why was this code removed?

russell-taylor (changes requested):

mmtbx/validation/clashscore2.py
vbchen: It wasn't dropped, the per-atom loop was folded into the n_atoms counting loop just above it, which now iterates by residue group and accumulates natoms_b_cutoff in the same pass (if has_b_cutoff and a.b < self.b_factor_cutoff: self.natoms_b_cutoff += 1), so we no longer traverse every atom a second time. When b_factor_cutoff is set, the count and the resulting clashscore_b_cutoff are identical to before.
  
One behavior change that turned up when I was double-checking the pull request: the clashscore_b_cutoff calc is now gated on has_b_cutoff = (not fast) and (b_factor_cutoff is not None). Previously, when no b_factor_cutoff was given, natoms_b_cutoff stayed 0 and the code fell through to clashscore_b_cutoff = 0.0. Now it's left as None, which seems more correct — reporting 0.0 when no B-factor cutoff was applied is misleading. This isn't visible in show_summary (which already guards on b_factor_cutoff is not None) or as_JSON, but it does change get_clashscore_b_cutoff() from 0.0 to None in that default case. We can switch it back to the old behavior if we think that is better.

russell-taylor (approved):

PR #1180, opened 2026-06-17
Refactor mmtbx/kinemage/validation.py: fix bugs, dead code, consolidate entry points, probe2 dots

vbchen:
## Summary
Foundational refactor of `mmtbx/kinemage/validation.py`: fix several bugs, remove dead code, consolidate the two kinemage entry points onto a shared builder, and migrate probe-dot generation from shelling out to the legacy `reduce`/`probe` executables to the in-process `reduce2`/`probe2` Python API.

Key changes:
- Consolidate `make_multikin` and `export_molprobity_result_as_kinemage` onto a shared `_build_kinemage`.
- Migrate `make_probe_dots` to the probe2 Python API (no more external `reduce`/`probe` subprocess).
- Fix correctness bugs: an altloc `==`-vs-`=` assignment, an `is not` string comparison that should be `!=`, duplicate geometry-deviation overlays (emitted per chain segment instead of per chain id), and a duplicate `{length dev}` line in the footer.
- Remove dead code (`add_fan`, `add_spring`, `get_residue_bonds`, `get_angle_outliers`, `get_bond_outliers`) and extract readable helpers out of `get_kin_lots`.

## Probe-dot correctness
The probe2-API migration originally had two bugs that made `make_probe_dots` throw and silently return empty (a bare `except` swallowed it): the `Optimizer` was handed top-level params instead of `params.probe`, and it called a nonexistent `DataManager.set_default_output_dir()`. Both are fixed here (the latter via `CCTBXParser`, matching `clashscore2.py`), so `run()` again emits probe/H-bond dots.

Verified that the refactored `run()` produces the **same 69 greentint H-bond dots** as the previous (legacy-engine) code on the standard His31–Asp70 H-bond test case.

## Tests
- New `mmtbx/kinemage/tst_validation.py` covering the extracted helpers, altloc handling, `build_kinemage`, `make_multikin`, and that the deleted functions are gone. Passes locally.
- The companion `phenix_regression/validation/tst_kinemage.py` (made engine-agnostic in a paired MR) passes against this branch.

## Notes
Large but cohesive single commit (the refactor is internally interdependent). Touches only `mmtbx/kinemage/validation.py` and its new test.

### DIFF
--- a/mmtbx/kinemage/validation.py
+++ b/mmtbx/kinemage/validation.py
@@ -10,7 +10,7 @@ from mmtbx.validation import rna_validate
 from mmtbx.validation import omegalyze
 from mmtbx.kinemage import kin_vec
 from iotbx.pdb import common_residue_names_get_class
-from libtbx import easy_run, Auto
+from libtbx import Auto
 from scitbx import matrix
 from cctbx import geometry_restraints
 from mmtbx.monomer_library import pdb_interpretation
@@ -19,7 +19,6 @@ from iotbx.pdb.rna_dna_detection import residue_analysis
 import iotbx.phil
 from libtbx.utils import Sorry, Usage
 from libtbx.str_utils import show_string
-import libtbx.load_env
 
 def get_master_phil():
   return iotbx.phil.parse(
@@ -57,33 +56,6 @@ def build_name_hash(pdb_hierarchy):
     i_seq_name_hash[atom.i_seq]=atom.pdb_label_columns()
   return i_seq_name_hash
 
-def get_angle_outliers(angle_proxies, chain, sites_cart, hierarchy):
-  i_seq_name_hash = build_name_hash(pdb_hierarchy=hierarchy)
-  kin_text = "@subgroup {geom devs} dominant\n"
-  for ap in angle_proxies:
-    restraint = geometry_restraints.angle(sites_cart=sites_cart,
-                                          proxy=ap)
-    res = i_seq_name_hash[ap.i_seqs[1]][5:]
-    altloc = i_seq_name_hash[ap.i_seqs[1]][4:5].lower()
-    cur_chain = i_seq_name_hash[ap.i_seqs[1]][8:10]
-    if chain.id.strip() is not cur_chain.strip():
-      continue
-    atom1 = i_seq_name_hash[ap.i_seqs[0]][0:4].strip()
-    atom2 = i_seq_name_hash[ap.i_seqs[1]][0:4].strip()
-    atom3 = i_seq_name_hash[ap.i_seqs[2]][0:4].strip()
-    if atom1[0] == "H" or atom2[0] == "H" or atom3[0] == "H":
-      continue
-    sigma = ((1/restraint.weight)**(.5))
-    num_sigmas = - (restraint.delta / sigma) #negative to match MolProbity direction
-    if abs(num_sigmas) >= 4.0:
-      angle_key = altloc+res[0:3].lower()+res[3:]+' '+atom1.lower()+ \
-                  '-'+atom2.lower()+'-'+atom3.lower()
-      kin = add_fan(sites=restraint.sites,
-                    delta=restraint.delta,
-                    num_sigmas=num_sigmas,
-                    angle_key=angle_key)
-      kin_text += kin
-  return kin_text
 
 def angle_outlier_as_kinemage(self):
   """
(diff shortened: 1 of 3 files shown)


mmtbx/kinemage/validation.py
```diff
@@ -299,135 +245,98 @@ def bond_outlier_as_kinemage(self):
...
+        model_manager = reduce_add_h_obj.get_model()
+
+        # Build probe parameters for optimizer using probe2 defaults
+        # Optimizer expects the probe sub-params, not the top-level params
+        import iotbx.phil
+        probe_phil = iotbx.phil.parse(probe2.master_phil_str, process_includes=True)
+        probe_params = probe_phil.extract()
+        opt = Optimizers.Optimizer(probe_params.probe, False, model_manager,
+          modelIndex=None, fillAtomDump=False)
+      except Exception:
```
russell-taylor, line 292: With an upcoming change, you'll need to remove any requested hydrogens from the model here because the Optimizer no longer does this. See https://github.com/cctbx/cctbx_project/blob/65a531c4b822698aa9492ab16b01e1de36e07747/mmtbx/programs/reduce2.py#L1299

I'll add a tickler to that issue to remind us to change this in master in case this pull request has been merged before that change. @olegsobolev Keep this in mind as well.

mmtbx/kinemage/validation.py
```diff
@@ -299,135 +245,98 @@ def bond_outlier_as_kinemage(self):
...
+        # If hydrogen addition fails, continue with existing atoms
+        pass
+
+    # Rebuild model manager after hydrogen changes
```
russell-taylor: Consider doing the sorting of atoms and other rebuilding like is done in _ReinterpretModel when make_restraints=False: https://github.com/cctbx/cctbx_project/blob/783fdd586e53c4149149dbdbd9d228c84e7e8de5/mmtbx/programs/reduce2.py#L1077

This is what Reduce2 had to do for it to work in all cases.

russell-taylor (changes requested):
I didn't look through the changes with a fine-toothed comb, presuming that the testing you did will reveal problems with the changes in common cases.

mmtbx/kinemage/validation.py
russell-taylor: You might consider importing and calling this function?

mmtbx/kinemage/validation.py
vbchen: Thanks, good to know. Since that Optimizer change isn't in master yet, I've left this as-is for now; when it lands we'll add the explicit hydrogen-removal step here. Happy to coordinate with @olegsobolev so the two stay in sync (and this gets updated in master if this PR merges first).

mmtbx/kinemage/validation.py
vbchen: Good call — done. Replaced the naive model rebuild with the same reinterpretation reduce2 does in _ReinterpretModel(make_restraints=False): sort_atoms_in_place() + reset_serial(), then process(make_restraints=False) with get_reduce_pdb_interpretation_params(...) and flip_symmetric_amino_acids=False. Verified the kinemage tests still produce identical probe/H-bond dots.

I looked at calling it directly, but _ReinterpretModel is a method on the reduce2 Program (it operates on self.model/self.params), so it isn't importable as a standalone. I mirrored its body instead, reusing the shared reduce_hydrogen.get_reduce_pdb_interpretation_params() helper so the interpretation params live in one place. If you'd prefer, I'm happy to extract the reinterpretation logic into a shared free function that both reduce2 and this can call — let me know.

mmtbx/kinemage/validation.py
vbchen: Update: now that #1185 has landed, I've reworked this — make_probe_dots calls the new mmtbx.hydrogens.place_and_optimize_hydrogens, which calls getHydrogensToDelete() and removes them after optimization. So the hydrogen removal you flagged is handled, and it stays in sync with reduce2 since it's the shared path.

mmtbx/kinemage/validation.py
vbchen: The reinterpretation is now done inside place_and_optimize_hydrogens (the same sort_atoms_in_place / reset_serial / process(make_restraints=False) with the reduce params), so we get it by calling the shared helper rather than duplicating it.

make_probe_dots now imports and calls place_and_optimize_hydrogens instead of duplicating the logic. One small addition: I gave it a raise_on_missing flag (default True, so no change for existing callers); the kinemage view calls it with raise_on_missing=False so a few residues without restraints don't fail the whole structure.  I wanted this mode for MolProbity Next so it can still generate dots for the rest of the structure even with a few missing restraints. Happy to adjust if you'd prefer that gated differently.

russell-taylor (approved):

PR #1184, opened 2026-06-18
Match Niggli iteration-limit exception message

dwmoreau:
The Niggli reduction algorithm was ported from python to c++ in PR
1152. When the iteration-limit was exceeded, the python code exception string was "Krivy-Gruber iteration limit exceeded" and the c++ port is "Niggli reduction iteration limit exceeded". This PR changes the c++ exception to match.


phyy-nx:
Rebased on master

PR #1185, opened 2026-06-19
Issue 1173 rebase

russell-taylor:
Moved hydrogen-removal out of Optimizers.Optimizer() and into the caller, which is reduce2 and other programs.

@vbchen you will need to add code to your use of this that matches what is done in mmtbx/hydrogens/__init__.py before you reinterpret the model in your branch once this is merged.

It also returns a list of flips that were made, whether they were due to locking down a configuration or optimizing a Mover; this is to help with issue 1173.


russell-taylor:
Okay @vbchen and @olegsobolev it has passed all tests on several platforms, so it looks like I didn't break anything fundamental. Going ahead with the merge. You'll want to remove the hydrogens by hand.

PR #1187, opened 2026-06-24
Find best cell performance

dwmoreau:
A case was encountered in dials.stills_process where indexing with subsampling was not able to be applied due to the excessive run-time. Profiling found that 68% of execution time was stuck in the __init__ method of cctbx/crystal/find_best_cell.py. This PR reduces it to 18%.

The find_best_cell method takes an arbitrary crystal.symmetry object as an input and exposes the best conventional cell as a new crystal.symmetry object, the change of basis operation to reach that setting, and every equivalent setting that was enumerated.

Within the method, it creates separate four different pathways: 1) triclinic, 2) monoclinic, 3) orthorhombic, and 4) the other high-symmetry crystal systems. In the triclinic case, it utilizes the Niggli reduction, in the high-symmetry case, it assumes the change of basis operator is identity (no-op). In the orthorhombic case, it sorts the axes in increasing magnitudes, while preserving the spacegroup. For the monoclinic case, it enumerates the unit cells under the change of basis operator, and chooses the reduced cell with the "best" conventional setting, that preserves the spacegroup. In the specific profiled example, the known unit cell was monoclinic, and the enumeration in the monoclinic case, caused an excessive runtime. 

The main branch code works by first by obtaining the spacegroup number to identify the correct branch to utilize. Triclinic and high-symmetry cases and returns early. Then it branches off into either the monoclinic or orthorhombic cases.

This work is a performance improvement that works in several ways. Performance is improved by caching repetitive operations that are the same for a specific spacegroup (via functools.lru_cache). Then it is further improved by replacing specific operations with faster calls:
	- change_of_basis_op(rt_mx) -> change_of_basis_op(rt_mx.as_xyz())
		rt_mx is the Rotation–Translation matrix object in sgtbx
	- space_group_info().type() -> hash(space_group)
	- crystal.symmetry.change_basis(cb_op) -> space_group.average_unit_cell(unit_cell.change_basis(cb_op))
		cb_op is a change of basis operator.
The method all_cells() returns the symmetry object of each enumerated setting and symmetry() returns the symmetry object of the best setting. These symmetry objects are created during the __init__method. However, in the stills_process pathway, neither of these methods are utilized, only the change of basis operator is desired. The new code creates the symmetry objects for these methods lazily, only retaining the change of basis operators.

A new test is established that tests find_best_cell.py, which did not exist before. Test data is generated from the main branch and the unit cells from the niggli reduction test. The test was rerun on this branch and the returned values from the exposed methods, cb_op(), symmetry(), and all_cells(), are bit-identical.

The Niggli reduction pathway for triclinic is only taken for P1 (spacegroup #1). P -1 (spacegroup #2) takes the identity route. I suspect this could be a latent bug in the existing code. This needs to be confirmed. If it is a bug, it will be addressed in a separate PR.


dwmoreau:
The test I created for this branch that verified binary equivalence between the old & new codes succeeds locally, but is failing on the Azure pipeline.

bkpoon:
It looks like the new test only passes on linux (from the "quick" GitHub Actions check). Also, rebase your branch to get the latest changes from `master`. That should speed up the `CI branch` checks.

bkpoon:
Something changed and the test fails on linux as well now.

Is the pickle file in regression something that's going to change regularly with code changes? It's pretty small, but it's a binary file.

dwmoreau:
I'm open to changes in how the testing is done. What is the preferred way to make tests? The xfel_regression tests that I'm more familiar with use expt & refl files with established results, the tests must reproduce them.

There is currently no test for this method, unless it is included in a larger test. I did an extensive test comparing the old and new code for exact equivalence between the old and new code. I pared back that test and added it here. I also ran dxtbx, dials, & xia2 testing, which all pass.

bkpoon:
For binary files that may or may need to be updated, we use the `*_regression` repositories because they are on GitLab and they are more generous with `git-lfs`. That way, every time the binary file is changed, the repository does not keep every copy of the file in the repository. And we won't hit the `git-lfs` quotas that are on GitHub.

How do you generate the expected result? Is it something that can be done on the fly?

The testing infrastructure already pulls in `phenix_regression` by default. I can probably add `xfel_regression` to the other pipelines and GitHub actions. The `xfel_regression` directory is already cached and used in the `XFEL CI` pipelines.

PR #1188, opened 2026-06-24
Xfel gui sacct fix

dwmoreau:
Reduce Slurm load from the xfel GUI job monitor

Summary

The cctbx.xfel GUI's JobMonitor thread polls the status of submitted processing jobs. Its previous implementation ran a separate sacct subprocess for every submission id of every active job, every 5 seconds, querying Slurm's accounting database (slurmdbd). On the shared NERSC cctbx account this produced ~6 sacct queries/second sustained, slowing Slurm cluster-wide.

This PR reworks how the GUI tracks job status to put minimal, bounded load on Slurm.

Changes

- Poll the Slurm controller instead of the accounting database. Live status now comes from a single squeue call per cycle (in-memory slurmctld state — cheap), with a single batched sacct used only to resolve the final outcome of jobs that have already left the queue. In steady state this places essentially zero load on slurmdbd.
- One query per cycle, not one per job. All active jobs are queried together rather than spawning a subprocess per submission id. Finished jobs (including TIMEOUT) are not re-queried.
- Sensible status for multi-part (ensemble) jobs. Jobs split across several sub-jobs now report a derived status (RUNNING if any part runs, FAILED if any part fails, DONE only when all finish) instead of showing "unknown"; finished sub-jobs are cached and not re-queried.
- Bulk DB writes and reduced UI churn. Status changes for a cycle are written to the application database in a single UPDATE, and the Jobs tab is redrawn only when something actually changed.
- Poll interval relaxed from 5s to 10s.

Net effect: from dozens of accounting-DB queries every few seconds to ~one controller query per 10s, independent of job count.

Files

- xfel/ui/components/submission_tracker.py — squeue-based batched query with sacct terminal-state fallback; ensemble aggregation; per-chunk terminal cache; shared
state-mapping helper.
- xfel/ui/db/xfel_db.py — update_job_statuses() bulk status-update helper.
- xfel/ui/components/xfel_gui_init.py — JobMonitor reconcile loop (batched query, bulk write, delta-only refresh).

### DIFF
--- a/xfel/ui/components/submission_tracker.py
+++ b/xfel/ui/components/submission_tracker.py
@@ -27,6 +27,31 @@ def _map_slurm_state(raw):
     print('Unknown job status', status)
   return SLURM_STATUS_MAP.get(status, 'UNKWN')
 
+# Per-chunk statuses that are terminal and stable, so they never need to be
+# re-queried. 'UNKWN' is deliberately excluded (it is transient -- a job not yet
+# registered with the scheduler reads as UNKWN until it appears).
+CACHEABLE_TERMINAL = frozenset(["DONE", "EXIT", "DELETED", "ERR", "SUBMIT_FAIL", "TIMEOUT"])
+
+def _aggregate_ensemble(statuses):
+  """Collapse the statuses of a job's chunks (comma-separated submission ids,
+  e.g. ensemble refinement) into one overall status. Active states win so the
+  job keeps being tracked; once nothing is active, a single failed chunk marks
+  the whole job EXIT, and only an all-clear set is DONE."""
+  s = [st for st in statuses if st is not None]
+  if not s:
+    return "UNKWN"
+  if "RUN" in s:
+    return "RUN"
+  if "PEND" in s or "SUBMITTED" in s:
+    return "PEND"
+  if "SUSP" in s:
+    return "SUSP"
+  if any(st in ("EXIT", "ERR", "TIMEOUT") for st in s):
+    return "EXIT"
+  if "UNKWN" in s:
+    return "UNKWN"
+  return "DONE"
+
 class JobStopper(object):
   def __init__(self, queueing_system):
     self.queueing_system = queueing_system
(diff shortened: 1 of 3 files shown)


xfel/ui/components/submission_tracker.py
```diff
@@ -191,11 +276,23 @@ def track(self, submission_id, log_path):
     if submission_id is None:
       return "UNKWN"
     all_statuses = [self._track(sid, log_path) for sid in submission_id.split(',')]
+    return self._aggregate(all_statuses)
+
+  @staticmethod
+  def _aggregate(all_statuses):
```
phyy-nx: Why is this static method created?  It doesn't appear to be called anywhere.  The new comment is fine though.

xfel/ui/components/xfel_gui_init.py
```diff
@@ -363,19 +365,34 @@ def run(self):
...
         jobs = db.get_all_jobs(active = self.only_active_jobs)
 
-        for job in jobs:
-          if job.status in ['DONE', 'EXIT', 'SUBMIT_FAIL', 'DELETED']:
+        # Collect the jobs that still need tracking, then query the queueing system
+        # in a single batched call (one squeue per cycle) rather than once per job.
+        # TIMEOUT is terminal, so skip it too; UNKWN is intentionally NOT skipped
+        # because it is transient for slurm (e.g. ensemble jobs with mixed states).
+        active_jobs = [job for job in jobs
+                       if job.status not in ['DONE', 'EXIT', 'SUBMIT_FAIL', 'DELETED', 'TIMEOUT']]
```
phyy-nx: `CACHEABLE_TERMINAL` also lists `ERR`, so `ERR` should be treated as terminal here, right?  Should we just import `CACHEABLE_TERMINAL` from submission_tracker?

phyy-nx (changes requested):
Changeset looks great, just a few questions

xfel/ui/components/submission_tracker.py
dwmoreau: It actually was reachable — track() calls it, and the base-class track_many still routes through track() for the non-Slurm queueing systems (LSF/PBS/SGE/Local/HTCondor). But since it's a single call site that's never overridden or reused, the @staticmethod wasn't earning its keep, so I've inlined the unanimity logic back into track() and kept the docstring as an inline comment.

Note the Slurm path is deliberately separate: SlurmSubmissionTracker.track_many uses the module-level _aggregate_ensemble (precedence collapse: RUN > PEND > SUSP >  EXIT/ERR/TIMEOUT > UNKWN > DONE) rather than the strict all-must-agree rule, so I left that untouched.

xfel/ui/components/xfel_gui_init.py
dwmoreau: The filter now imports and uses CACHEABLE_TERMINAL directly (job.status not in CACHEABLE_TERMINAL) instead of a hand-maintained list, which adds the missing ERR and gives us a single source of truth so the two can't drift again.

One subtlety worth flagging: CACHEABLE_TERMINAL intentionally excludes UNKWN (it's transient — a job not yet registered with the scheduler reads as UNKWN), so this keeps UNKWN jobs being tracked, which is the behavior we want. That's also why I didn't reuse finished_job_statuses from job.py, which does include UNKWN — different meaning ("finished for bookkeeping" vs. "terminal, stop polling").

phyy-nx (approved):

PR #1194, opened 2026-06-30
iotbx: allow pickling/deepcopy of pdb input with a populated __dict__

bkpoon:
- declare __getstate_manages_dict__ and round-trip the __dict__ in __getstate__/__setstate__ so a populated cache (e.g. _scale_matrix) no longer breaks pickle/deepcopy; preserve the dict rather than drop it (scale_matrix()'s cache can be a deliberate None not recoverable from the source lines)
- add tst_input_pickling.py regression test (pdb + cif, incl. suppressed-None)


olegsobolev:
LGTM

PR #1195, opened 2026-07-01
cctbx.xfel.merge export batch MTZ for careless

nksauter:
Define a new "export" worker with cctbx.xfel.merge to consolidate the expt+refl files from all ranks, and write a single MTZ to disk in "batch mtz" format.  The file will have appropriate columns and metadata for careless.

### DIFF
--- a/xfel/merging/application/export/README.md
+++ b/xfel/merging/application/export/README.md
@@ -1,0 +1,9 @@
+# Export batch MTZ for parallel merging by Careless
+
+## Problem statement
+
+
+## Installation notes
+
+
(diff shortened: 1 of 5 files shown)


nksauter:
Feedback received from @phyy-nx as follows
1) Phil string help should make it clear that this specific instance exports data for careless but that other export options might be provided in the future.
2) worker_step information breaks encapsulation; would cause the worker to fail if it is run in an environment other than the merge.py program.  Provide this information another way.
I will revise accordingly before merging the pull request.

PR #1204, opened 2026-07-14
model: preserve stop_for_unknowns across manager.select()

vbchen:
## Summary

`mmtbx.model.manager.select()` rebuilds the selected model via the `manager(...)`
constructor, forwarding `crystal_symmetry`, `restraint_objects`, and
`monomer_parameters` — but **not** `stop_for_unknowns`. So a model created with
`stop_for_unknowns=False` silently reverts to the constructor default (`True`) after
any `.select()`. Downstream interpretation of the selected model then raises `Sorry`
on unknown residues that the caller intended to tolerate (e.g. the reduce2/clashscore2
pipeline with custom ligands).

This forwards `self._stop_for_unknowns` into the `select()` rebuild.

```python
       monomer_parameters         = self._monomer_parameters,
+      stop_for_unknowns          = self._stop_for_unknowns,
       expand_with_mtrix          = False,
```

### Scope

One line in `model.py` plus a regression in `tst_model_2.py`. The behavioral change is
narrow: models left at the default `stop_for_unknowns=True` are unaffected (the rebuilt
model was already `True`); only models that explicitly set `False` change — they now
correctly keep `False` across `select()`.

Split out from a larger custom-ligand fix so it can be reviewed on its own; the other
half (a prime/star atom-name synonym fallback in `pdb_interpretation`) is held back
separately as it overlaps ongoing nucleic-acid atom-name work.

## Testing

- New `exercise_select_preserves_stop_for_unknowns` in
  `mmtbx/regression/model/tst_model_2.py`: asserts `select()` preserves both the default
  `True` and an explicit `False`. It **fails on master** (selected model reverts to
  `True`) and passes with this change.
- `tst_model_2.py` full run passes, as do `tst_model_ncs`, `tst_model_mtrix`,
  `tst_model_biomt_mtrix`, and `tst_model_remove_alternative_conformations`.

### DIFF
--- a/mmtbx/model/model.py
+++ b/mmtbx/model/model.py
@@ -3955,6 +3955,7 @@ class manager(object):
       crystal_symmetry           = self._crystal_symmetry,
       restraint_objects          = self._restraint_objects,
       monomer_parameters         = self._monomer_parameters,
+      stop_for_unknowns          = self._stop_for_unknowns,
       expand_with_mtrix          = False,
       pdb_hierarchy              = new_pdb_hierarchy,
       log                        = self.log)
--- a/mmtbx/regression/model/tst_model_2.py
+++ b/mmtbx/regression/model/tst_model_2.py
@@ -628,8 +628,29 @@ END
   assert not m4.altlocs_present_only_hd()
 
 
+def exercise_select_preserves_stop_for_unknowns():
+  # model.manager.select() must carry stop_for_unknowns onto the selected
+  # model. Otherwise a model built with stop_for_unknowns=False silently
+  # reverts to the constructor default (True) after any selection, which breaks
+  # pipelines (e.g. reduce2/clashscore2 with custom ligands) that deliberately
+  # tolerate unknown residues.
+  # Default (True) is preserved across select().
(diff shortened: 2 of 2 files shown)


mmtbx/model/model.py
```diff
@@ -3955,6 +3955,7 @@ def select(self, selection, exclude_flags=False):
       crystal_symmetry           = self._crystal_symmetry,
       restraint_objects          = self._restraint_objects,
       monomer_parameters         = self._monomer_parameters,
+      stop_for_unknowns          = self._stop_for_unknowns,
```
russell-taylor, line 3958: Nice. Are there other arguments that should also be treated the same as the ones that were already there and this new one? Finding and adding all of them could save surprises later on.

vbchen: hmm.  since this touches model.py, I wanted to be conservative in my changes, but that's certainly an interesting question.  I guess we need Billy or Oleg to weigh in on that though.

olegsobolev (approved):
Thanks, LGTM, will merge.
@russell-taylor , other parameters are fine, will slightly widen the test as a separate commit.

PR #1206, opened 2026-07-16
Repair dps_extended.get_new_detector

### LINKED ISSUE #3213: dials.search_beam_position fails on some single panel detectors
I noticed that `dials.search_beam_position` almost always fails on small molecule MicroED datasets.

The attached archive contains `imported.expt` and `strong.refl` from  [an XRDa entry 558](https://xrda.pdbj.org/entry/558/path/frames), `frames/20260708_147422_04_-50_06_MicroED_1168_0_movie.tif`.

79 % of spots can be trivially indexed with `dials.index imported.expt strong.refl detector.fix=distance`. So the initial geometry is more or less fine.

Nonetheless `dials.search_beam_position imported.expt strong.refl mm_search_scope=0.5` fails with `Sorry: No valid scores`. (Note that the pixel size is only 16 um, so the default search scope 4 mm is too large)

[search_beam_pos_failure.tar.gz](https://github.com/user-attachments/files/30037034/search_beam_pos_failure.tar.gz)

biochem-fan:
This is to fix https://github.com/dials/dials/issues/3213. Please look at the discussion there.

@phyy-nx 
> Yes dropping the else path makes sense to me.

In this PR I kept the old (`else`) code path because the unification would change the behaviour (whether to store the shift in the hierarchy root or panels). Please feel free to delete the old branch if you think we don't need to preserve the old behaviour.

> For the hierarchy root there should be no difference between set_frame and set_local_frame, right?

Right. Both should do the same for the root. I nonetheless changed it to `set_frame`, because `get_origin` etc is semantically global.

### DIFF
--- a/rstbx/indexing_api/lattice.py
+++ b/rstbx/indexing_api/lattice.py
@@ -176,14 +176,14 @@ class _():
 
     if len(new_detector) > 1 and len(new_detector.hierarchy()) > 1:
       h = new_detector.hierarchy()
-      h.set_local_frame(fast_axis=h.get_fast_axis(),
-                        slow_axis=h.get_slow_axis(),
-                        origin=matrix.col(h.get_origin()) + origin_offset)
+      h.set_frame(fast_axis=h.get_fast_axis(),
+                  slow_axis=h.get_slow_axis(),
+                  origin=matrix.col(h.get_origin()) + origin_offset)
     else:
       for panel in new_detector:
-        panel.set_local_frame(fast_axis=panel.get_fast_axis(),
-                              slow_axis=panel.get_slow_axis(),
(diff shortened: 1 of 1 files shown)


phyy-nx:
> Please feel free to delete the old branch if you think we don't need to preserve the old behaviour.

I think we don't need the else branch.  The code as written either sets the origin shift on the hierarchy root or on individual panels, and it's trying to make that decision based on whether it thinks its a hierarchical detector or not.  But realistically, all detectors in dxtbx are hierarchical, even single panel detectors, so might as well only set the origin shift on the root.  At least, that's my opinion.

biochem-fan:
@phyy-nx Thanks for your comment. I removed the legacy code path.

phyy-nx (approved):

PR #1208, opened 2026-07-21
Fix reduce2 His flip breaking user-defined bonds (#1199)

### LINKED ISSUE #1199: reduce2 NQH flips can break user-defined (custom) bonds — His ring especially
Follow-up from #1173 (raised there by @pafonine).

### Problem

When reduce2's `Optimizer` flips an Asn/Gln/His residue, the flip decision, placement, and ion "lock-down" are driven only by local geometry and element type (a spatial neighbor search plus `element_is_positive_ion()`). They never consult the model's actual **bond restraints**, so a **user-defined/custom bond** on a flipping atom can be silently broken: the ring or amide is flipped into an orientation that makes the bond geometrically impossible, while the bond restraint remains in the model.

### What actually happens today (verified in the code)

- **Amide (ASN/GLN):** already protected, but by accident. `MoverAmideFlip.__init__` requires the amide N to have exactly 3 bonded neighbors and the O exactly 1, so a custom bond there raises `ValueError`, which `_PlaceMovers` catches and logs as "Could not add MoverAmideFlip." The flip is suppressed, but it's reported as an error rather than an intentional skip.
- **Histidine:** this is the real gap. `MoverHisFlip.__init__` validates the ring nitrogens ND1/NE2 by counting hydrogens (`!= 1`) and carbons (`!= 2`), not total neighbors. A custom bond from ND1 or NE2 to a metal or other partner leaves those counts unchanged, so the Mover is built and the ring is flipped — breaking the bond. (The ring carbons are protected by strict neighbor-count checks; only the two ring N's slip through.)
- The ion **lock-down** paths (amide `foundIon`, His `bondedConfig`) are purely geometric/element-based and ignore the restraint bond list entirely.

### Good news

The information needed to fix this is already available inside `_PlaceMovers`: `bondedNeighborLists` is built from `model.get_restraints_manager().geometry.get_all_bond_proxies()`, which already contains user-defined bonds. No new plumbing is required.

### Suggested fix

In `mmtbx/reduce/Optimizers.py`, in the amide and His flip blocks of `_PlaceMovers`, before placing a flip Mover or applying an ion lock-down, check the bonded neighbors of the atoms the flip would move for any bond outside the residue's standard connectivity; if an unexpected bond is present, skip the flip and leave the residue in its original orientation (and log it as a deliberate skip). The Mover's `_atoms` list enumerates exactly the moving atoms, so the check can compare each moved atom's neighbors against that set plus the expected backbone attachment. This should close the His hole and make the amide behavior an intentional, logged skip rather than a caught exception.

/cc @russell-taylor @pafonine

olegsobolev:
Fixes #1199.

## Problem

reduce2's `Optimizer` decides Asn/Gln/His flips (and the ion "lock-down" that suppresses them) from local geometry and element type only — a spatial neighbour search plus `element_is_positive_ion()`. It never consults the model's bond restraints, so a user-defined (custom) bond on a flipping atom is invisible to the flip logic and can be silently broken.

The Histidine ring is the exposed case: `MoverHisFlip` validates the ring nitrogens ND1/NE2 by counting hydrogen (`== 1`) and carbon (`== 2`) neighbours, not their total number of bonds, so an extra bond — to a metal, or to another partner — slips through, the flip Mover is built, and the ring can be flipped into an orientation that makes the bond geometrically impossible. (The ring carbons use strict neighbour-count checks and are already protected; the Asn/Gln amide is protected because `MoverAmideFlip` raises on an unexpected neighbour count.)

## Fix

`_PlaceMovers` now skips the His flip when a ring nitrogen carries a bond outside the standard ring connectivity, logging `Did not add MoverHisFlip … (bond outside standard ring connectivity)`.

The check runs **after** the ion lock-down (`bondedConfig`), so a genuinely metal-coordinated histidine found by the spatial ion search is still locked in place, not skipped. Verified on 1xso: the full set of His flip/lock decisions is byte-identical before and after this change.

Also rewords the amide message `Could not add` → `Did not add MoverAmideFlip` (intentional skip, not an error).

## Tests

`mmtbx/regression/tst_reduce_custom_bond.py` (now registered in `run_tests.py`):
- **control** — the histidine is flippable when nothing unusual is bonded;
- **bug** — a user-defined ND1–SG bond suppresses the flip;
- **no-regression** — a genuinely Zn-coordinated His (His 69 + Zn from 1xso) stays locked; this fails if the bonded-neighbour check is placed before the lock-down instead of after it.

Verified locally: the new tests, `tst_reduce.py`, and `tst_reduce_switch.py` all pass; the full-1xso reduce2 flip decisions are unchanged; `find_clutter` clean.

### DIFF
--- a/mmtbx/programs/reduce2.py
+++ b/mmtbx/programs/reduce2.py
@@ -34,7 +34,7 @@ import copy
 from iotbx.data_manager import DataManager
 import csv
 
-version = "3.0.0"
+version = "3.1.0"
 
 master_phil_str = '''
 approach = *add remove optimize
--- a/mmtbx/reduce/Optimizers.py
+++ b/mmtbx/reduce/Optimizers.py
@@ -1036,7 +1036,7 @@ class Optimizer(object):
               self._infoString += _VerboseCheck(self._verbosity, 1,"Added MoverAmideFlip "+str(len(self._movers))+" to "+resNameAndID+"\n")
               self._moverInfo[self._movers[-1]] = "AmideFlip at "+resNameAndID+" "+aName
           except Exception as e:
-            self._infoString += _VerboseCheck(self._verbosity, 0,"Could not add MoverAmideFlip to "+resNameAndID+": "+str(e)+"\n")
+            self._infoString += _VerboseCheck(self._verbosity, 0,"Did not add MoverAmideFlip to "+resNameAndID+": "+str(e)+"\n")
 
       # See if we should insert a MoverHisFlip here.
       # @todo Is there a more general way than looking for specific names?
@@ -1136,6 +1136,14 @@ class Optimizer(object):
             self._infoString += _modifyIfNeeded(fixUp.atoms[4], coarsePositions[4], fixUp.atoms[5])
 
             self._infoString += _VerboseCheck(self._verbosity, 1,"Set MoverHisFlip on "+resNameAndID+" to state "+str(bondedConfig)+"\n")
+          elif addFlipMovers and _HisRingNitrogenHasNonstandardBond(a, bondedNeighborLists):
+            # Issue #1199: a ring Nitrogen carries a bond outside the standard ring
+            # connectivity that the ion lock-down above (bondedConfig) did not handle
+            # -- a user-defined/custom bond, or an ion the spatial search missed.
+            # Flipping the ring could break that bond, so leave the residue in its
+            # original orientation rather than adding a flip Mover.
+            self._infoString += _VerboseCheck(self._verbosity, 1,"Did not add MoverHisFlip to "+resNameAndID+
+              " (bond outside standard ring connectivity)\n")
           elif addFlipMovers: # Add a Histidine flip Mover if we're adding flip Movers
             # Check to see if the state of this Mover has been specified. If so, place it in
             # the requested state and insert the Mover as a non-flipping Histidine.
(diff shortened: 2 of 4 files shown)


russell-taylor (changes requested):
This is a very similar approach to the one I took in https://github.com/cctbx/cctbx_project/tree/issue-1199 and I think that they both suffer from the same problem: because I asked Dorothee to make hydrogenate put all potential hdyrogens onto the histidine, when we discover that there is a conflict we must remove the associated hydrogen. Otherwise, it leaves that hydrogen hanging on. To fix this, it must be appended to the deleteAtoms list as is done in _ModifyIfNeeded. I see three potential approaches to fix this:

- Modify the hydrogen-adding code to not add a hydrogen to either or both of the Nitrogens that already has a bond. If this is done, then the code in _PlaceMovers that looks for ionic bonds should be modified to skip this check if all expected hydrogens are not present.
- Modify your pull request to FIRST check for _HisRingNitrogenHasNonstandardBonds and append the related hydrogen(s) to the deleteAtoms list. Then do the existing `if bondedConfig is not None:` and `elif addFlipMovers` when the _HisRingNitrogenHasNonstandardBonds fails.
- Modify my branch to deal with this all sensibly.

I think that the middle approach, modifying your pull request, is best. I can do this if you prefer or I'm glad to review a modified request.

olegsobolev:
Thanks Russell — agreed, the dangling hydrogen has to be removed. I've updated the PR so the His branch appends the bonded nitrogen's hydrogen to `deleteAtoms` and marks it as an acceptor, mirroring `_modifyIfNeeded`. The regression test now also asserts HD1 lands in `getHydrogensToDelete()`.

On the all-hydrogens decision: I don't think it was wrong — I think it's exactly right, and it's the reason the removal has to live in the Optimizer rather than in the hydrogen-adding code (your first option). A caller can hand reduce a model that already has hydrogens and bypass the add step entirely — that's what `approach=optimize` does, it never calls `_AddHydrogens`, it just interprets restraints and optimizes — and callers can place hydrogens first and add a custom bond afterwards (the regression test does exactly that). So the Optimizer can't assume hydrogen placement was bond-aware, or that it ran at all. "Put all potential hydrogens on, then let the Optimizer prune the ones that conflict" is the robust design: it's what the ion lock-down already does, and the custom-bond case should match it.

I pulled your "Did not add MoverHisFlip" wording and version bump from issue-1199 into the PR. The check runs after the existing ion lock-down, so the 1xso flip/lock decisions come out byte-identical to master.

Let me know what you think — happy to iterate.

russell-taylor (approved):
Well done!

PR #1210, opened 2026-07-27
Add a Hessian runtype to xtb_manager

krlun:
run_hessian runs xtb --hess at a given geometry and returns the path to the Cartesian Hessian file, removing a stale one first and requiring normal termination. get_cmd gains a hessian runtype, get_xyz_lines an optional geometry override, and run_cmd a hessian pass-through.


mmtbx/geometry_restraints/xtb_manager.py
```diff
@@ -367,6 +390,24 @@ def _check_log(self, log_filename, log=None):
...
+      os.remove(hessian_filename)
+    self.run_cmd(log=log, hessian=True)
+    if not os.path.exists(hessian_filename):
+      raise Sorry(f'xtb --hess did not produce {hessian_filename}; '
+                  f'check {self.get_log_filename()}')
+    return os.path.abspath(hessian_filename)
+
   # ---- reading outputs -----------------------------------------------------
 
   def read_xyz_output(self):
```
nwmoriarty, line 423: Maybe we should give these functions a more descriptive names.

PR #1221, opened 2026-09-01
clashscore2: hydrogenate and process ensembles per model, not per file

vbchen:
`clashscore2` processed and hydrogenated multi-model files as one combined
hierarchy. Ensemble models are spatially superimposed, so making restraints
across the whole file is quadratic in the model count: a 21-model NMR entry was
measured peaking over 13 GB. Placement and flip optimization are per-model
operations anyway (the Optimizer already iterates MODEL records independently),
so this hydrogenates each model separately and recombines, then runs the single
interpretation-only re-process, which is linear, on the combined model. It also
processes restraints once per ensemble and caches the probe2 phil parse.

On 1D3Z (10 models, 12,310 atoms), peak memory with hydrogens regenerated goes
from 1,980 MB to 278 MB, and the kept-hydrogens run from 29.5 s to 10.8 s.

It is also a correctness fix. The hydrogen-deletion set the Optimizer reports
after a whole-ensemble run describes a single model, and applying it across the
combined hierarchy removes the wrong atoms: on 1D3Z the whole-file path returns
6,543 of 12,310 atoms and 523 of 6,290 hydrogens. All ten per-model clashscores
change as a result. Afterwards, regenerating hydrogens reproduces the
deposited-hydrogen clashscores exactly in all ten models, where before it
disagreed in all ten.

Single-model probe2 output is byte-identical, and tst_add_hydrogen_1 through
_11 pass.

`mmtbx/reduce/Optimizers.py` still resets its deletion set per model, so direct
multi-model reduce2 runs keep the old behaviour and want a separate fix.

### DIFF
--- a/mmtbx/hydrogens/__init__.py
+++ b/mmtbx/hydrogens/__init__.py
@@ -592,11 +592,9 @@ def default_probe_phil():
 
 def _place_and_optimize_core(model, do_flips, nuclear, keep_existing_H,
       probe_phil, raise_on_missing, log):
-  """Placement + orientation/flip optimization for a model, without the final
-  interpretation-only re-process.  Returns the hydrogenated model.  Raises
-  Sorry right after placement (before optimizing) when restraints were not
-  found for some residues and raise_on_missing is set, matching the original
-  order of operations."""
+  """Placement + orientation/flip optimization, without the final
+  interpretation-only re-process.  Raises Sorry after placement (before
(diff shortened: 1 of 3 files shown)


russell-taylor (approved):
The testing seems to show that it continues to work.

The code all looks good to me.

Seems like we need an issue created for Reduce2 to also handle things better. It would be great if this solves the class of failures where we run out of memory on model files.

PR #1222, opened 2026-09-01
Handle conda-forge c-compiler/cxx-compiler metapackages >=2.0.0

dwpaley:
As of conda-forge/compilers-feedstock#78 (Aug 2026), activating the conda env will no longer set the CC, CXX, LDFLAGS etc environment variables. Instead the conda env contains config files like bin/arm64-apple-darwin-clang++.cfg. To handle this new setup, if CC and CXX are unset, then we check for the compiler binaries in the conda environment. If no CC/CXX and no binaries in conda-bin, then fail loudly.

See dials/dials#3243

### DIFF
--- a/libtbx/SConscript
+++ b/libtbx/SConscript
@@ -805,14 +805,28 @@ int main() {
     cc = os.environ.get('CC', None)
     cxx = os.environ.get('CXX', None)
 
-    if cc is not None or cxx is not None:
-      print("Compiler settings:")
-      if cc is not None:
-        print("  CC: ", cc)
-      if cxx is not None:
-        print("  CXX:", cxx)
     if cc is None or cxx is None:
-      print("Warning: CC and/or CXX has not been set in the environment.")
+      # conda-forge compilers 2.0+ no longer require CC/CXX to be set; the
+      # compiler self-configures via a .cfg file next to the executable in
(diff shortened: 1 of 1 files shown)


bkpoon:
Do we want to recreate the flags provided by `conda-forge` instead? Adding these packages will work
```
cctools_osx-arm64
clang_osx-arm64
clangxx_osx-arm64
```
There are similar packages for the other platforms (`osx-64`, `linux-64`). The linux ones would be `gxx_linux-64` and `gcc_linux-64`.

dwpaley:
I guess it depends whether the .cfg approach is the general new direction for build environment setup, or if it's more of a weird experiment. I don't have a clear idea of which it is, but I could look into it unless you have any thoughts already?

If getting rid of those flags will cause other headaches that I'm unaware of, I would also be fine with changing the approach for that reason.

bkpoon:
The separation is the official way

https://conda-forge.org/docs/maintainer/infrastructure/#more-on-compiler-feedstocks

Also, the `--use-environment-flags` is for general use. If the flag is set, then any environment variables should already be set. We should not generally assume that the conda compilers are in the same conda environment as the dependencies. The `conda-build` process does not do this. The compilers are in a separate build environment. (`_build_env` vs `_h_env_placehold...`)

PR #1223, opened 2026-09-01
Enumerate symmetry images near selected atoms

krlun:
New module `cctbx/crystal/symmetry_images.py`, for finding the symmetry images lying within a radius of a chosen set of atoms.

### Interface

- `tables_within_radius(sites_cart, crystal_symmetry, radius)`: the pair-symmetry and site-symmetry tables covering every contact within the radius. Neither depends on which atoms are later used as seeds, so one build serves many selections.
- `images_near_seeds(symmetry_tables, seed_indices)`: `{j_seq: [rt_mx_ji]}` for the images those tables place near a seed.
- `images_within_radius(...)`: both in one call.

### What the result guarantees

- Each pair is listed from both ends, so an image is reported whether the seed is the first or the second atom of the pair.
- One operator per image: the canonical member of the coset the site's own symmetry generates. Seeding on more atoms therefore only adds images, and never changes which operator represents one.
- An image coincident with the site it came from is dropped.
- Operators carry `(1, 12)` denominators. `rt_mx` compares equal across differing denominators but does not hash equal, so normalising is what makes the results safe to put in sets and dicts.
- Site symmetry is derived at 1e-2 Å: a site written on a symmetry element at three decimals is recognised, while a site genuinely 0.2 Å off an axis is not snapped onto it.

### Testing

`cctbx/crystal/tst_symmetry_images.py`, 24 exercises, checked against a brute-force enumeration over all space group operators and a block of lattice translations. There are two references: operators, for sites in general positions, and positions, for special positions where several operators give one image.

Run bare, the test sweeps the `debug_utils` space groups up to order 96; given a symbol it runs that group alone. It is registered in `cctbx/run_tests.py` as `["$D/crystal/tst_symmetry_images.py", "I41/acd"]` and takes ~1.5 s there.

### For reviewers

- Additive only: nothing else in the tree calls the module yet.
- `images_near_seeds` costs about 0.05 s for 2000 atoms at 5 Å and 0.20 s at 8 Å; building the tables is comparable.
- Preparatory consolidation for future PRs of endo_exo and water protonation heuristics.

### DIFF
--- a/cctbx/crystal/symmetry_images.py
+++ b/cctbx/crystal/symmetry_images.py
@@ -1,45 +1,34 @@
 """Symmetry images lying within a radius of selected sites."""
 from __future__ import absolute_import, division, print_function
-from six.moves import range
 import collections
 
-# The two tables `images_near_seeds` reads.  Neither records which sites it
-# was built from; one paired with a table built from other sites can give
-# wrong images.
+# Both tables must be built from the same sites; neither records which, so a
+# mismatched pair gives wrong images.
 tables = collections.namedtuple(
   "tables", ["pair_sym_table", "site_symmetry_table"])
 
-# A site counts as lying on a symmetry element only when its symmetry mate is
-# within this distance.  Coordinates written to three decimals put a site on
-# an element within 0.002 A of its mate.  A site taken to lie on an element is
-# moved onto it, its own image then coincides with it and is dropped, and its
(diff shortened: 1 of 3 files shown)


krlun:
Refactoring super_cell.py (get_siiu in particular) instead. Closing this PR.

PR #1224, opened 2026-09-02
Add a regression test for barbed_wire_analysis

vbchen:
`mmtbx/validation/barbed_wire_analysis.py` had no test. There is no
`tst_barbed_wire*` anywhere in the tree, it is not registered in
`mmtbx/run_tests.py`, and the only files importing it are its own program and
command-line wrappers.

That matters more than usual here, because the analysis shells out to
`phenix.reduce` and `phenix.probe` (there is a `#TODO update to Reduce2 and
Probe2` on `do_probe`). Nothing would have reported those going missing: the
contacts would come back empty, every packed residue would be classified
unpacked instead, and the suite would stay green.

This runs the analysis on `pae_model.pdb`, an AlphaFold model already in
`mmtbx/regression/pdbs`, 309 residues with pLDDT in the B-factor column. It
takes about 3.5 seconds and the classification is stable across repeated runs.

### What it asserts, and why each one earns its place

- The residue count, and that every non-empty category is one of the five the
  analysis defines.
- The first two and last three residues carry no classification, which
  documents the boundary behaviour rather than leaving it implied.
- `Predictive` covers at least 250 of the 309, and at least 200 residues have
  recorded contacts. **These are the two that catch a broken or missing
  `phenix.probe`.** Verified by stubbing `do_probe` to return an empty dict:
  both fire.
- `text_code` is either empty or six slots.
- The working directory is left clean. `do_probe` writes its input as
  `probe_contacts_temp.pdb` under a fixed name in the current directory, so the
  test runs in a `mkdtemp` and compares the directory listing before and after.
  Verified by stubbing out the `os.remove` cleanup: it fires.

The category counts are asserted as floors rather than exact numbers,
deliberately. They depend on the output of two external binaries, and an exact
count would be a platform tripwire rather than a regression test. The floors are
far enough from the observed values (277 predictive, 250 asserted) to survive
normal variation while still failing decisively when contacts go missing
entirely.

### Not addressed here

The fixed-name scratch file is a real concurrency hazard for anything running
two analyses in one process, and the reduce1/probe1 shell-out is what the
existing TODO is asking to replace. Both want their own change; this only makes
sure something notices if they break.

### DIFF
--- a/mmtbx/run_tests.py
+++ b/mmtbx/run_tests.py
@@ -429,6 +429,7 @@ general_tests = [
   "$D/regression/tst_fmodel_2.py",
   "$D/regression/tst_rigid_bond_test.py",
   "$D/validation/regression/tst_cablam.py",
+  "$D/validation/regression/tst_barbed_wire_analysis.py",
   "$D/regression/real_space_refine_chain/tst_00.py",
   "$D/regression/real_space_refine_chain/tst_01.py",
   "$D/conformation_dependent_library/tst_omega_cdl.py",
--- a/mmtbx/validation/regression/tst_barbed_wire_analysis.py
+++ b/mmtbx/validation/regression/tst_barbed_wire_analysis.py
@@ -1,0 +1,87 @@
+from __future__ import absolute_import, division, print_function
+
+import os
+import shutil
+import tempfile
+
+import libtbx.load_env
+from libtbx.utils import format_cpu_times
+from iotbx.data_manager import DataManager
+from mmtbx.validation import barbed_wire_analysis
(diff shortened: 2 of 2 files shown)


vbchen:
Closing for now: opened before it had been reviewed on our side. The test itself is sound and passes; will reopen once we've settled whether the category floors are the right form for CI.

vbchen:
Superseded by #1226, which pairs this test with the sliding-window fix it turned up. The floors question that prompted the close is settled there: the per-model assertions are now structural rather than measured.

